Issue #125, opened 2020-06-08
Wrong favicon showing in Firefox


garethbowen:
The head of the docs site looks like...

```
<link rel="shortcut icon" href=/favicons/favicon.ico>
<link rel=apple-touch-icon href=/beta/favicons/apple-touch-icon-180x180.png sizes=180x180>
<link rel=icon type=image/png href=/beta/favicons/favicon-16x16.png sizes=16x16>
<link rel=icon type=image/png href=/beta/favicons/favicon-32x32.png sizes=32x32>
<link rel=icon type=image/png href=/beta/favicons/android-36x36.png sizes=36x36>
<link rel=icon type=image/png href=/beta/favicons/android-48x48.png sizes=48x48>
<link rel=icon type=image/png href=/beta/favicons/android-72x72.png sizes=72x72>
<link rel=icon type=image/png href=/beta/favicons/android-96x96.png sizes=96x96>
<link rel=icon type=image/png href=/beta/favicons/android-144x144.png sizes=144x144>
<link rel=icon type=image/png href=/beta/favicons/android-192x192.png sizes=192x192>
```

The first one 404s because the site is currently has the `/beta` path - this will be fixed when we go live.

The `favicon-16x16.png` and `favicon-32x32.png` look as you'd expect, and these are chosen by Chrome.

All the others render a list icon:

![image](https://user-images.githubusercontent.com/2064938/83983652-45337980-a984-11ea-8c02-f0a6d1ff227f.png)

This is what is shown in Firefox, and presumably Android and iOS too.

Remove or update the incorrect favicons.

garethbowen:
@MaxDiz The wrong icons still exist, eg: https://docs.communityhealthtoolkit.org/favicons/android-96x96.png

Also the wrong icon is still showing in my Firefox but this may be a caching issue.

The fix is to replace all the icons I listed above with the CHT icon. Alternatively you could remove the icon definitions in the header but I don't know how Hugo generates those.

SCdF:
On my desktop I get this:
![image](https://user-images.githubusercontent.com/583851/84076843-63be6280-a9ce-11ea-891c-586a9f781a10.png)
Which means we aren't doing transparency correctly.

I also get the icon linked above by Gareth in Firefox Mobile, probably because that is pulling a different resolution.

abbyad:
Thanks for debugging. I'll regenerate all the icons and republish.

garethbowen:
[...]

PR #130, opened 2020-06-08
Added missing favicons

abbyad:
Generated using file in issue #125 with https://cthedot.de/icongen


abbyad:
Ya, I find them jaggy too. We don't have any levers to pull using cthedot.de, so tried a couple others.  They seem better with https://realfavicongenerator.net, which also generated the same variety of sizes.

Here are some sample improvements:

image

image

Issue #135, opened 2020-06-09
Last modified date and commit is the same for all pages


abbyad:
The last modified date and commit is meant to be specific to each page, but showing as the same for all pages in the live deployed site.

![image](https://user-images.githubusercontent.com/789512/84194065-771f0b80-aa6a-11ea-81f4-6f690b3d21fb.png)

The date and commit are correct for local builds, so this is likely due to our GH Action not getting the full git history.

PR #136, opened 2020-06-09
Get git history

abbyad:
For #135. 

Also simplifies the build to get the submodule immediately as opposed to via another step.

Informed by https://github.com/google/docsy-example/issues/60#issuecomment-632903155


nomulex:
I like this brevity. LTM

Issue #137, opened 2020-06-09
Import API docs from cht-core


abbyad:
The [API docs](https://github.com/medic/cht-core/tree/100b5413ac9b09ab33f95b55a1a75df90f35f761/api) should be ported over to the doc site.

garethbowen:
### README.md

Much of this should be migrated to the docs site, but we need to have something at the root level. Maybe a brief introduction and then links to the docs?

### api/README.md

Move the API documentation to the docs site, and put the migrations and other internal details in code comments.

### release-notes/README.md

Leave this where it is - it's just an index of all the different versions. We could consider moving all the release notes to the docs site at some point.

### scripts/**/README.md

Leave these where they are.

### sentinel/README.md

This is mostly outdated and I think should just be deleted.

### shared-libs/search/README.md

This is internal API documentation so should be moved to the code as jsdoc comments.

### shared-libs/transitions/src/lib/pupil/README.md

This is the readme of a library that we inlined and then modified, so it's irrelevant. Leave it where it is. One day I hope to replace it with the version in npm...

### tests/scalability/README.md

Move this to the docs site. We should consider finding and moving all the testing resources ([example](https://docs.google.com/document/d/1ab_wjxsT9millrBAy3SjZxWBT8KRmBElBuwgz9XAGII/edit)) into the new docs site too as QA reference. Maybe this is a separate issue...

PR #142, opened 2020-06-11
Import API

abbyad:
For #137

### DIFF
--- a/content/en/apps/reference/api.md
+++ b/content/en/apps/reference/api.md
@@ -2,7 +2,6 @@
 title: "API to interact with CHT Applications"
 linkTitle: "API"
 weight : 1
-type: "swagger"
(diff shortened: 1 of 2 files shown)


abbyad:
Thanks, I forgot to commit the table of contents short code. I'll commit that and your recommended changes tomorrow.

abbyad:
>  I expect it's mostly copy and paste?

Correct, it was all copy and paste and fixed broken links.

> does the search work better if you add some keywords?

The keywords are not used for search. We previously used them to compute the related content pages. We now explicitly list them to be sure of which pages are shown.

abbyad:
Incorporated the recommended changes, and fixed the build -- ready for review again.

I'll squash and merge when it passes review.

PR #143, opened 2020-06-11
Adds info about `fields.private`

dianabarsan:
medic/cht-core#6370

### DIFF
--- a/content/en/apps/guides/performance/replication.md
+++ b/content/en/apps/guides/performance/replication.md
@@ -144,4 +144,4 @@ Some reports need to be signed off by a supervisor even though that supervisor d
 
 ### Sensitive documents
 
-We won't replicate documents that are about the user when the sender is someone the user can't access. For example, if a supervisor submits a report about one of their CHPs, that CHP won't be able to see it.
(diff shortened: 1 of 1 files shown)


content/en/apps/guides/performance/replication.md
```diff
@@ -43,4 +43,4 @@ Some reports need to be signed off by a supervisor even though that supervisor d
 
 ### Sensitive documents
 
-We won't replicate documents that are about the user when the sender is someone the user can't access. For example, if a supervisor submits a report about one of their CHPs, that CHP won't be able to see it.
\ No newline at end of file
+We won't replicate reports that are `private` and are about the user when the sender is someone the user can't access. A report is private when it has a field named `private` with a value of `true`. For example, if a supervisor submits a private report about one of their CHPs, that CHP won't be able to see it.
```
SCdF, line 46: I think we should clarify the exact JSON path, since "field" here could be misinterpreted as meaning "json property", i.e. `doc.private` instead of the actual `doc.fields.private` that is needed.

dianabarsan: Amended.

Issue #138, opened 2020-06-10
Improve robots.txt and sitemap.xml to improve search engine crawling


garethbowen:
The site has a [robots.txt](https://docs.communityhealthtoolkit.org/robots.txt) but I think it needs a couple of tweaks, firstly to include `Disallow:`, and secondly to reference the sitemap.xml, eg:

```
User-agent: *
Disallow:

Sitemap: /sitemap.xml
```

We also have a [sitemap.xml](https://docs.communityhealthtoolkit.org/sitemap.xml) but I'm not sure it's up to date, for one thing it doesn't list the home page as one of the pages.

abbyad:
I see this entry in the sitemap.xml:
```xml
<loc>https://docs.communityhealthtoolkit.org/</loc>
<lastmod>2020-06-04T16:05:57-04:00</lastmod>
```

Were you expecting another home page entry?

On a related note, I am moving away from offline search to Google Search for the site, which should help with some of this as well.

garethbowen:
> I see this entry in the sitemap.xml:

So it is! Sorry about that... I must've skimmed right past it.

> Were you expecting another home page entry?

No that's perfect.

abbyad:
It's automated by Hugo, so I wasn't taking it personally :)

abbyad:
Search is now included. Will need to go through a small process to remove Google ads.

garethbowen:
Looks great. There's a tab for "discussion" which shows search results from the forum. Should that be renamed to "Forum" so it's clearer what it is?

abbyad:
The site is searchable, and ads have been removed. Some deeper pages have yet to be crawled and indexed so won't be included in the search just yet. If searching in Google with the full path of a page doesn't yield the page itself as a result, then the page has yet to be indexed.

Re Forum vs Discussion: Waiting on input from @CescaGarrett on that, and it is easily updated.

abbyad:
Heard via DM from Francesca, and she thinks _Discussion_ is more appropriate in this context.

Also, searches of deeper pages seem to be found:
![image](https://user-images.githubusercontent.com/789512/84721816-882ab980-af4f-11ea-8900-288dcbf5c273.png)

@garethbowen, with that is this issue ready to be closed?

garethbowen:
Sure!

PR #144, opened 2020-06-11
Improve Searchability

abbyad:
Search engine crawling is improved with this PR, as is internal search on the site as it now uses Google search. 

For #138


abbyad:
Need to fix the search page, converting to draft until ready.

Issue #146, opened 2020-06-12
Improve 'Architecture of CHT Instances'


SCdF:
Page: /core/overview/architecture/
Version: 2020-06-04 16:16:18  1200  1200
---

The architecture diagram is no longer accurate. It now looks like this:

![image](https://user-images.githubusercontent.com/583851/84517321-0a547d00-acc7-11ea-8aa1-838a722bf10a.png)

In text form:
 - Are we sure that gateway talks to rapidpro? Or do we here mean either medic-gateway or any other sms aggregator?
 - sentinel talks to couchdb through api not directly
 - couch2pg talks to couchdb through api not directly

MaxDiz:
@garethbowen do you own this diagram? (or do you know who does?)

garethbowen:
Merged.

PR #151, opened 2020-06-15
Update Architecture diagram

garethbowen:
#146


abbyad:
I am surprised to see the auditing layer being hit before API, as those external components hit API end points, and not the auditing layer directly. Would it make sense to include the auditing layer within API?

Also, it may help to label the CHT box as CHT Core, and then Android and Browser could be labelled as CHT Applications.

Not required to change as part of this PR though, and we can revisit as part of a visual design pass since it is often used in broader communications.

garethbowen:
> I am surprised to see the auditing layer being hit before API, as those external components hit API end points, and not the auditing layer directly. Would it make sense to include the auditing layer within API?

Auditing is a proxy which simply logs out the request and then passes it on to API. It's provided by haproxy with accompanying configuration. It's a separate process so I think it makes sense to show it and I wanted to emphasize that all requests coming in to the CHT are audited before we do anything else.

> Also, it may help to label the CHT box as CHT Core, and then Android and Browser could be labelled as CHT Applications.

I think that's confusing because the CHT box is the server, and the Android and Browser boxes are the clients. CHT Core and CHT Applications exist on both the server and the client, for example most of the cht-core repo is the client webapp. It would make more sense to say CHT Server and CHT Client. What do you think?

abbyad:
Ah, my perspective is based on API vs medic-api... so it's understandable to communicate auditing as happening before medic-api. I think we can find a representation later that captures the API as a concept since that will help communicate integrations, but I'm happy to revisit this in a later iteration.

As for Client and Server, that makes sense to me!

PR #155, opened 2020-06-16
Rename remote to training

abbyad:
The term "remote" doesn't mean as much on it's own in the URL, so renamed to "training".


MaxDiz:
merged

PR #170, opened 2020-06-22
Update phc-covid.md

helizabetholsen:
@MaxDiz let me know if these revisions make sense. I added more explicit content on PHC disruptions, as well as COVID-specific considerations.

### DIFF
--- a/content/en/apps/examples/phc-covid.md
+++ b/content/en/apps/examples/phc-covid.md
@@ -1,6 +1,6 @@
 ---
 title: "Primary Health Care Adaptations COVID-19"
-linkTitle: "PHC COVID-19"
+linkTitle: "PHC Adaptations COVID-19"
 weight: 
 description: >
   Primary health care workflow and training adaptations for COVID-19.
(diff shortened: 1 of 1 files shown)


MaxDiz:
your changes look good to me!

abbyad:
Nice work in bringing in this page!

Given that the branch has already been merged, we'll have to create a new PR with further changes, but wanted to log in some comments here:

- I see `Adaptations COVID-19` in the titles, I think it would be better as either `Adaptations for COVID-19` or `COVID-19 Adaptations`
- It would be nice to avoid having double subtitles like those shown below. Eventually we may want the content fully in markdown to be searchable, but in the meantime removing the title from the image itself would be best.
image
- > The CHT’s Primary Health Care (PHC) workflows are easily adaptable to help communities and facilities strengthen continuity of routine primary care services during COVID-19. Adapting the CHT’s PHC workflows is designed to:

  The PHC workflows are using the CHT, not workflows built into the core CHT itself, so I think it would be better to say something like `Primary Health Care workflows using the CHT are easily adaptable[...]`
- There are a few acronyms used without being spelled out (eg iCCM, ANC, even CHW). We should write them out as was done with `Primary Health Care (PHC)`.

abbyad:
Also, should this link to the [slide deck](https://docs.google.com/presentation/d/1DBbEc2R6-TwgD4pp_LjTmgGXAUzZzeVrYlrDzFdWqSI) like the other examples?

Issue #171, opened 2020-06-22
Typo in medic-conf command on contact page


mrjones-plip:
on the [contact page](https://github.com/medic/cht-docs/blob/master/content/en/apps/reference/forms/contact.md) we have a `medic-conf` call that looks like this:

```
medic-conf --local convert-contact-forms upload-contact-form
```

the `medic-conf` app barfs on this b/c `ERROR Unsupported action(s): upload-contact-form`. Turns out that we're just missing an `s` on `upload-contact-form`.  Fix is to change it to this:

```
medic-conf --local convert-contact-forms upload-contact-forms
```

PR #172, opened 2020-06-22
Fix missing 's' in medic-conf call

mrjones-plip:
adds missing `s` on `medic-conf` call

see #171


mrjones-plip:
This my first PR (for docs or cht).  Please lemme know if did anything wrong - I'm open to feedback!

mrjones-plip:
Let it be stated hence forth: no matter how small your contribution, every commit counts, even it is literally no bigger than a single character.  

I'm very please my first merge to master @ MM was simply `s` ;)

Issue #176, opened 2020-06-23
Improve 'Documentation Workflow': links in the upper right hidden too often?


mrjones-plip:
Page: /contribute/docs/workflow/
Version: 2020-06-22 23:08:51 -0400 -0400
---

Heya!  I've been using the new docs site for a week and after @abbyad's post in slack about the new workflow, I checked out this page.  It reads (emphasis mine): 

> Anyone can contribute to CHT documentation by opening an issue in the cht-docs repo or by using the “Edit this page” or “Create documentation issue” **links in the upper right corner of your window**

I literally had no idea these links in the upper right existed because I do so much reading on mobile/tablet and, apparently, on many browser window narrower than 1200px. At <1200px wide, these links simply disappear (yes @SCdF, I even checked outside my dogmatic "no javascript allowed" browser ;)  .  Given how helpful they are, maybe we should consider smooshing them into a hamburger menu instead of the `display: none!important;` treatment they get now? 

This might be a super corner case (silly developers wanting a 3 up display of browsers on one screen), but I see we also run Google Analytics which can tell us how many esoteric users like me there actually are.  We can even show stats on window size vs screen size if wanna [get real fancy](https://css-tricks.com/google-analytics-can-show-screen-resolution-%E2%89%A0-browser-window/)!

Feel free to close as "love ya, won't fix" ;)

abbyad:
Glad you brought this up! It's not clear what the best way would be to handle that. Would help in the short term to mention in documentation that those links are available when viewing in desktop mode only?

mrjones-plip:
And, oh! i see on responsive (~750px) we actually collapse the left hand side navigation into a right hand side hamburger menu. Dishy indeed! Maybe we could double down on this and first collapse the edit links into that and then append the navigation into it too?  

Otherwise, yeah, short term saying that these links are available on desktop would a good tip for sure. One step further would be to say something like, "only in desktop mode with a wider browser window". But admittedly that last part may be more confusing that clarifying.

abbyad:
Thanks, we can look at improving that. The main structure and collapsing is handled by the [Docsy theme](https://github.com/google/docsy/) that we are using, so this would be good feedback for maintainers at Google!

mrjones-plip:
Interesting! I just spent a bit too long poking around on the [docsy repo](https://github.com/google/docsy) ;) 

[...]

PR #179, opened 2020-06-24
Adds meta links when not shown in right pane

abbyad:
@mrjones-plip aside from a code review it would be great for you to also [test these changes locally](https://github.com/medic/cht-docs#building-the-documentation) before we merge.

For #176


mrjones-plip:
OMG - lerv this!  thanks @abbyad! Quick test on FF and mobile and looking good. It works on all pages except the homepage (`/`).  It looks like feedback has different treatment on the homepage so I assume the feedback/TOC include is unique and thus this is the expected behavior?

layouts/partials/page-meta-lastmod.html
```diff
@@ -0,0 +1,2 @@
+{{ T "post_last_mod"}} {{ .Lastmod.Format .Site.Params.time_format_default }}{{ with .GitInfo }}: <a  href="{{ $.Site.Params.github_repo }}/commit/{{ .Hash }}">{{ .Subject }} ({{ .AbbreviatedHash }})</a>{{end }}
```
mrjones-plip, line 1: Not that big a deal, but are very long lines like this in code the norm for `cht-docs`? I see we don't do line breaks on content (I assume folks use softwrap in their IDEs), but I wasn't sure about code/template. 

/me goes to read docs formatting docs...

abbyad: The newly added partial is actually just a version of the [existing partial from the theme](https://github.com/google/docsy/blob/master/layouts/partials/page-meta-lastmod.html), with the page links appended -- so this line is taken verbatim from the theme. We don't have any recommendation for line length in `cht-docs`, likely as you mention because the pages are mostly markdown or html which have lots of paragraph text and don't make sense to explicitly wrap. That said, if a page is heavy with new code I think formatting it with appropriate line breaks still makes sense. This partial could have used line breaks for clarity, but I'd rather leave it the way it was in the original theme file.

abbyad:
> It looks like feedback has different treatment on the homepage so I assume the feedback/TOC include is unique and thus this is the expected behavior?

Thanks for the quick review! The home page doesn't have this footer... which seems to be a separate issue, but I'll try to fix it as part of this issue too since it is somewhat related.

abbyad:
Footer now showing on the home page. @mrjones-plip can you take another quick tour to confirm it works as expected?

PR #186, opened 2020-06-25
Add Component Library

n-orlowski:


content/en/design/core/_index.md
```diff
@@ -10,9 +10,9 @@ The purpose of this guide is to document the standard aspects of our core framew
 
 Changes to the styles in this document should go through product design and be implemented throughout the app before this document is updated. 
 
-Medic in-app sample: https://github.com/medic/medic/blob/master/webapp/src/templates/partials/theme.html
+View the [CHT Component library](https://codepen.io/n-orlowski/pen/YzwVeBQ)
```
abbyad, line 13: This would be better in a medic or CHT account. Can we move it easily?

content/en/design/core/_index.md
```diff
@@ -10,9 +10,9 @@ The purpose of this guide is to document the standard aspects of our core framew
 
 Changes to the styles in this document should go through product design and be implemented throughout the app before this document is updated. 
 
-Medic in-app sample: https://github.com/medic/medic/blob/master/webapp/src/templates/partials/theme.html
+View the [CHT Component library](https://codepen.io/communityhealthtoolkit/pen/jOWLmqw)
```
abbyad, line 13: Nice addition! It would be useful to add something about how the CHT component library could and should be used. The mixed case makes me wonder if it should be CHT Component Library, or CHT component library... which do you think?

PR #188, opened 2020-06-25
Document permission property of form context

garethbowen:


abbyad:
Adding the _Blocked_ label since this seems related to https://github.com/medic/cht-core/issues/6401. @garethbowen, feel free to update if that was applied incorrectly.

garethbowen:
It's not really blocked on AT. That issue extends the permissions to contact forms as well as non-contact forms.

Issue #85, opened 2020-02-13
Create documentation describing the different CouchDB databases and what is in each


michaelkohn:
There are a number of different databases in CouchDB, but I can't find any documentation anywhere that gives an overview of what each one is, how they get named, what you can find in them, etc...

I would expect that documentation to be [here](https://github.com/medic/medic-docs/blob/master/development/db-schema.md).

In addition, for each different section, [Reports](https://github.com/medic/medic-docs/blob/master/development/db-schema.md#reports) for example, we should specify which database that document type can be found in.

<img width="1217" alt="Screen Shot 2020-02-13 at 1 12 21 PM" src="https://user-images.githubusercontent.com/10096985/74411509-83e87580-4e63-11ea-82d7-c08c2dd9cc0c.png">

esthermmoturi:
This has been fixed by the above mentioned PRs. Documentation can be found [here](https://docs.communityhealthtoolkit.org/building/guides/database/)

PR #200, opened 2020-07-02
Remove documentation: app_settings.replication

garethbowen:
This feature is no longer configurable as it's now required to be
run for the CHT to work as expected.

medic/cht-core#6519


garethbowen:
Good suggestion! I was wondering where the right place for this information is and realised we have basically no info about which db is used to store what (as Michael pointed out in [this issue](https://github.com/medic/cht-docs/issues/85). So instead of adding it to this issue I submitted a [PR for that one instead](https://github.com/medic/cht-docs/pull/201).

Is that sufficient?

garethbowen:
Merged as this feature is now deprecated.

PR #203, opened 2020-07-06
fix toc links, update style guide for same, add "read-more" to "see-also" shortcode

mrjones-plip:
We had links in the TOC that didn't jump down the page, but instead jumped to another page or another site.  Fixed the instances of this as well as updated the style guide to explicitly guide future editors.

Issue #178, opened 2020-06-24
Update style guide & content to not use links on titles, add "read-more" option to "see-also" shortcode


mrjones-plip:
**Update:** we're merging #205 into this ticket after it was already under way.  This is because `read-more` and `see-also` shortcodes are identical except for their prefix.

----

### No Links in titles

On the [home page](https://docs.communityhealthtoolkit.org/) and on the [architecture page](https://docs.communityhealthtoolkit.org/core/overview/architecture/) we use some titles (`###`, but could be `##` too) as links.  Here's a list as of now (below).  I think this is bad UX and we shouldn't do this. The reason is that these links get rendered in the TOC and are expected to jump you to that spot lower in the same page. Instead, the TOC links take you to a whole other page.

Proposal - update the the [style guide](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/) to call out to not do this and remove the instances below (and any more that may have cropped up since).  Instead of linking the `h3` we should add a `Read More` text at the end of the paragraph and link that text.

```
content/en/core/overview/architecture.md:### [CouchDB](http://couchdb.apache.org)
content/en/core/overview/architecture.md:### [api](https://github.com/medic/medic/tree/master/api)
content/en/core/overview/architecture.md:### [sentinel](https://github.com/medic/medic/tree/master/sentinel)
content/en/core/overview/architecture.md:### [PostgreSQL](https://www.postgresql.org)
content/en/core/overview/architecture.md:### [cht-core](https://github.com/medic/cht-core)
content/en/core/overview/architecture.md:### [medic-android](https://github.com/medic/medic-android)
content/en/core/overview/architecture.md:### [medic-gateway](https://github.com/medic/medic-gateway)
content/en/core/overview/architecture.md:### [medic-collect](https://github.com/medic/medic-collect)
content/en/core/overview/architecture.md:### [medic-conf](https://github.com/medic/medic-conf)
content/en/_index.md:### [Why the CHT?]({{% ref "why-the-cht" %}})
content/en/_index.md:### [CHT Applications]({{% ref "apps" %}})
content/en/_index.md:### [CHT Core Framework]({{% ref "core" %}})
content/en/_index.md:### [Design System]({{% ref "design" %}})
```

While we're in the style guide, we should be sure to avoid one of the problems originally found in this ticket which was bad links with `HAHAHUGOSHORTCODE` as the destination. The ticket has since been repurposed, but one of the comments before was:

[...]

PR #207, opened 2020-07-07
stop see-also shortcode from always showing an anchor

mrjones-plip:
we should be sure to test:
* basic `{{< see-also page="design/icons" >}}`
* basic with title  `{{< see-also page="design/icons" title="Learn about the Icon Library" >}}`
* basic with anchor  `{{< see-also page="design/icons"  anchor="about-the-icon-library"  >}}`
* basic with title and anchor  `{{< see-also page="design/icons"  title="Learn about the Icon Library" anchor="about-the-icon-library"  >}}`

### DIFF
--- a/layouts/shortcodes/see-also.html
+++ b/layouts/shortcodes/see-also.html
@@ -1,3 +1,8 @@
+<!--
+    This is technically identical to the 'read-more` template, but used differently. That said, likely
+    the styles for both should be kept in sync.
+    See https://github.com/medic/cht-docs/pull/203 for context
+-->
 {{ $page := .Get "page" }}
 {{ $anchor := .Get "anchor" }}
 {{ $save := . }}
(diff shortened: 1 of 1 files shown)


mrjones-plip:
Yes, can do!  I'll do the same preemptively on the new `read-more` template in  #203's PR

abbyad:
It would be nice to avoid keeping two shortcodes in sync... are these two shortcode _exactly_ the same, other than the prefix text? If so, there may be a cleaner way to do this. Something like a function that take the prefix, page, title, and anchors. Or perhaps simpler would be a `see-also` shortcode where default prefix "See also" is replaced by provided text (eg `prefix="Read more"`)? Some useful info passing args to templates in the [Hugo shortcode templates docs](https://gohugo.io/templates/shortcode-templates/).

mrjones-plip:
yup, yup - I had that feeling that this was an unnecessary de-normalization of code creeping up on me.  Indeed,  they're _exactly_ the same except for the prefix. 

Kinda arbitrary which ticket to pick to win, but I'm gonna jump ship on this ticket and fold it into #178.  It'll be easier to track that way with the final PR I think!

PR #222, opened 2020-07-09
fix images code sample on style guide

mrjones-plip:
just some missing `/* */`!


mrjones-plip:
thanks @n-orlowski - merged to master and closing!

Issue #211, opened 2020-07-07
Improve doc site search to make it more user friendly


MaxDiz:
Reviews from new users have highlighted confusion about the current doc site search. Eliminating the intermixing of search results from the doc site with results from outside the doc site is the largest area for user experience improvement. We discussed the potential to split search into "primary" for the doc site and "secondary" for anything CHT.

MaxDiz:
related #197

abbyad:
Assigning to @MaxDiz for AT

MaxDiz:
LGTM!

abbyad:
Thanks, closing now that it's been AT'ed and the PR has been merged.

PR #223, opened 2020-07-09
211 - Shows documentation search results first

abbyad:
The search now shows the results for the _Documentation_ first. @n-orlowski and I looked at other options of having a primary and secondary search, and in the end we were happy to have a clean way to have highlight the primary search, yet still making the secondary search available. 

For #211


MaxDiz:
Nice one! Sometimes the best improvement is the simplest one :)

Issue #212, opened 2020-07-07
Improve navigation UI


n-orlowski:
- Different active page highlight colour (#5F7681)
- Add dropdown indicator (chevron/triangle icon)
- Indent mobile nav child items for better hierarchy

Removing items below to consider in the future
<strike>
- Use top level nav items to expand child elements (see example [here](https://themes.gohugo.io//theme/hugo-theme-zdoc/docs/) for consistent desktop + mobile behaviour) and create "Overview" pages within the sections where we can move the top level content.
So the structure would be:

CHT Applications [only action is to collapse/expand child items below]
-- Overview (content on top level page gets moved here)
-- Concepts
-- Features
etc.

This makes it easier to navigate the menu on desktop and mobile without needing to load several pages

PR #228, opened 2020-07-10
212 nav

n-orlowski:
Updated nav highlight to our brand orange (our green is bit light and didn't have too much contrast) + mobile indentation
Still trying to figure out how to add dropdown chevrons 😸  

See: cht-docs #212


assets/scss/_variables_project.scss
```diff
@@ -49,3 +49,13 @@ h2, h3 {
...
+
+
+.td-sidebar-nav__section-title a {
+    color: #fff;
+}
+
+.td-sidebar-nav__section ul {
+    padding: 0;
+    margin: 0 0 0 0.5em;
+}
```
abbyad, line 61: These changes aren't showing for me locally. The theme style is being considered ahead of this project (S)CSS file. One way to get around this is to make these `!important`, which I recommend hesitantly! Another option is to have a project file for `_sidebar-tree.scss`, but you'd have to copy everything, not just add to it like we do with the `_variables_project.scss` file -- which seems worse than the two `!important` here. 

These changes also have unintended consequences:
- the titles are white on white -- seems as though you want that to be `color: $secondary` to be orange as per above.
- the margin on the `ul`s pushes all sections in, meaning that the top level page (why-the-cht) doesn't line up with it.

I'll see what we can do to fix this and we can revisit together.

abbyad:
Given that I committed the riskier change it would be good to get another reviewer. 

Also, are you still planning on adding chevrons as part of this PR?

abbyad:
The menu now has carets 🥕 

I would still like to find a way to specify the style without the `!important`s though

n-orlowski:
@abbyad Should I merge to revisit or are you going to try that in this PR?

abbyad:
Let's send it to someone with CSS experience for code review, and then if you are happy with the UX, then feel free to merge.

abbyad:
@mrjones-plip I re-assigned this issue to review since I ended up writing some code in here. It would be nice to get away from the `!important`s in CSS, if you have good ideas about that, or feel free to reassign to another teammate as you wish.

mrjones-plip:
Diving into this now!  All the features seem to work that were requested in #212, except this one:

> Use top level nav items to expand child elements

If we're gonna skip that one, I think we should update the ticket to remove it from the docket for this PR.  Oh! Maybe I misunderstood the feature request - please let me know if I need to RTFM harder!

abbyad:
Right, we didn't do that... @n-orlowski how do you feel about the current version, where the menu sections don't expand without reloading the page? Do you want to split that off for future consideration, or do that now?

n-orlowski:
Future consideration! It's a larger change for sure. I'll update the issue :)

mrjones-plip:
This was fun - thanks for letting me review this PR! 

**tl;dr:** removing all `!important`s, removing the current 3 `> ul` lines and adding this boring non-SCSS CSS at the bottom of [_variables_project.scss](_variables_project.scss) works for me in my dev environ (`font-weight` & `color` didn't seem negatively affected by removing `!important`):

```
.td-sidebar-nav > .td-sidebar-nav__section > ul { 
   padding-left:0; 
}
```

I haven't worked with SCSS before, but the need for `!important` is caused by insufficient [specificity](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity).  I kinda forgot how CSS worked in this particular case, so I'm being verbose so I don't forget why/how I came to this conclusion ;) Sorry if it's annoying!

One fix would be to let CSS do it's cascading thang by re-ordering the `@import`  sequence of `.scss` files in [main.scss](https://github.com/google/docsy/blob/053672c24cf51417c300f8ea5d32db545340746d/assets/scss/main.scss). By me moving `
@import "variables_project";` to be after `@import "sidebar-toc";` did indeed let the cascading work without `!important`.

However, since `main.scss` is in the foreign docsy repo, and I'm not sure how easy (or even possible?) it is to override that template in our local `assets` dir, let's look at fixing the specificity.  We can do that by including more classes from the DOM than were declared originally, which is from [_sidebar-tree.scss](https://github.com/google/docsy/blob/053672c24cf51417c300f8ea5d32db545340746d/assets/scss/_sidebar-tree.scss#L34) and gets compiled to this declaration:
```
.td-sidebar-nav__section > ul { 
   padding-left: .5rem; 
}
```

By updating our nested SCSS declaration so that it compiles to the above **tl;dr** statement with the `.td-sidebar-nav > ` prefixed should work! Explicitly including more classes in the declaration, and by using the direct child selector - this allworked in my testing on FF and Chrome on the desktop and Chrome on Android.  

One side note: I'm not sure how brittle this high level of specificity will be in terms how often our template structure of the TOC changes.  We could be more flexible with something more simple like `.td-sidebar-nav ul ul {}` which also works.

abbyad:
That's helpful! Given that we don't want to play with the import order figuring out the specificity will be best. I have something working and will push a new version within the next day.

abbyad:
Ready for a round of AT @n-orlowski, and review @mrjones-plip.

Review of requirements:
- [ ] Different active page highlight colour (#5F7681)
- [ ] Add dropdown indicator (chevron/triangle icon)
- [ ] Indent mobile nav child items for better hierarchy

Also
- [ ] Pages and sections should have same styling (color and weight were previously different)

assets/scss/_variables_project.scss
```diff
@@ -49,3 +52,45 @@ h2, h3 {
 {
   clear: both;
 }
+
+#td-section-nav {
```
mrjones-plip, line 56: ah - nice! i tried to this extra level of nesting to work as a "suggestion" in GH, but didn't know SCSS foo well enough.  this looks much better! I'll suck down the branch and test locally real quick.

n-orlowski:
Screen Shot 2020-07-15 at 3 10 12 PM

Thanks for your help, both!! 
Chevron not loading for me.. am I missing something?

abbyad:
Interesting... definitely needs more testing! What browser is that on?

mrjones-plip:
Yes, interesting indeed! I also note that the `padding-left:0;` is not taking effect. "Why the CHT?" and "CHT Applications" should both be flush left like so:

adj toc correct

@abbyad  - is there the concept of a cache bust query string to ensure that slightly tweaked CSS is loaded in a dev environment w/i a commit to a branch? (I assume this is well covered by hugo on merge to master/push to prod).

mrjones-plip:
I'm able to reproduce the chevrons not loading on Safari (MacOS 10.13.6 Safari 13.1.1 (13609.2.9.1.3).  However, indentation is correct!

n-orlowski:
I'm on Chrome!

n-orlowski:
Hmm our fonts are also different

mrjones-plip:
Oh weird?! It does work for me in chrome (Ubuntu 18.04 Chrome 84.0.4147.89) but indeed does NOT work for me on Chrome on mac (MacOS 10.13.6 Chrome 83.0.4103.97)

Both of that "bad" browsers don't like some of the html characters.  I swapped `23F5` -> `25B8` and `23F7` -> `25BE` and both Safari and Chrome are happy (though they're not as nice chars, and need some left pad...) @abbyad - maybe another pass picking up some [new chevrons](https://en.wikipedia.org/wiki/List_of_XML_and_HTML_character_entity_references) to ride shotgun with our new TOC?  I'll see if there's a canonical entry on which ones work and which ones don't.

me goes to research...

abbyad:
@mrjones-plip, @n-orlowski, ready for another round of cross device/browser tests!

mrjones-plip:
> Hmm our fonts are also different

Oh yes - I missed this (both your comment and the visual discrepancy).  Not sure how far we want to go down the "does it look right on linux" as that's gotta be likely far less than 1% of our user base ;)  

That all said - happy to test on our key browsers (basically Safari/Chrome/FF on Mac/Win I assume?) - so - shoulda caught this!

layouts/partials/sidebar-tree.html
```diff
@@ -36,7 +36,9 @@
 <ul class="td-sidebar-nav__section pr-md-3">
 {{ if .section.Parent }}{{/* hide root from menu */}}
   <li class="td-sidebar-nav__section-title">
-    <a  href="{{ $s.RelPermalink }}" class="align-left pl-0 pr-2{{ if not $show }} collapsed{{ end }}{{ if $active}} active{{ end }}{{ if $selected}} selected{{ end }} td-sidebar-link td-sidebar-link__section">{{ $s.LinkTitle }}</a>
+    <a  href="{{ $s.RelPermalink }}" class="align-left pl-0 pr-2{{ if not $show }} collapsed{{ end }}{{ if $active}} active{{ end }}{{ if $selected}} selected{{ end }} td-sidebar-link td-sidebar-link__section">
```
mrjones-plip, line 39: yeah, font awesome FTW! I can dig it.

Are we still using `collapsed`, `active` and `selected` classes now that we have an explicit use of the `fa-*` on it's own line? Maybe clean up and remove? 

I guess not really a big deal to leave them as the cost to render the static HTML is only done once on the server, and is nice to have classes in place for future CSS.  Not too concerned, but worth a quick question.

mrjones-plip:
Build failing aside, we're totally good on functionally: all browsers look good re. chevrons appearing and being correctly rotated or not!

I asked little question re cleanup.

layouts/partials/sidebar-tree.html
abbyad: From my understanding, `collapsed` is used for hiding the subsections, so better to keep that in -- thanks for checking!

abbyad:
The error on the "Build" step was `fatal error: concurrent map read and map write`. Not sure what caused it... rerunning all jobs worked this time around.

Issue #232, opened 2020-07-11
Hidden file in menu


abbyad:
The `_partial_functions.md` is read into two [`targets.js`](https://docs.communityhealthtoolkit.org/apps/reference/targets/) and [`tasks.js`](https://docs.communityhealthtoolkit.org/apps/reference/tasks/) pages and isn't meant to be accessed [independently](https://docs.communityhealthtoolkit.org/apps/reference/_partial_functions/). In the menu it adds a transparent item, _seen_ [here](https://docs.communityhealthtoolkit.org/apps/reference/tasks/) by hovering in the space after tasks.js in the left hand nav:
![image](https://user-images.githubusercontent.com/789512/87211789-2446a700-c2e9-11ea-8e13-b5fe798d173f.png)

We should find another way to use that file, ideally while avoiding duplication across `targets.js` and `tasks.js`.

PR #233, opened 2020-07-11
Hide item from menu

abbyad:
For #232. Hides the menu item using frontmatter. In doing so we can't use the readfile shortcode/function anymore so added a new shortcode which reads the file content.

Duplication is avoided, but drawback is that the content of the partial file is not in the TOC on the right side -- but this is not a regression. If there is a method to avoid duplication and adds the content to the TOC we should consider it. We could consider making this content a true page, and refer to it via `see-also`. The reason this wasn't chosen so far was that is it nice to have this content inline. Open to other ideas though!

### DIFF
--- a/content/en/apps/reference/_partial_functions.md
+++ b/content/en/apps/reference/_partial_functions.md
@@ -1,3 +1,8 @@
+---
+title: "Utils Functions"
+toc_hide: true
+---
+
 ## Utils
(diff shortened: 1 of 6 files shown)


mrjones-plip:
awe, i'm touched - my first request for a review ;)  I'll grab this next week!

abbyad:
That's a great idea, thanks @mrjones-plip!

abbyad:
Refactored a little more heavily... utils is now properly in the table of contents, and also brought `nools-extras.js` to it's rightful place in reference section. Ready for re-review.

PR #236, opened 2020-07-16
QA Process changes

newtewt:
Documenting some changes we as QAs have agreed to.

### DIFF
--- a/content/en/contribute/code/workflow.md
+++ b/content/en/contribute/code/workflow.md
@@ -73,28 +73,28 @@ Issues in this column are being actively worked on, which includes development, 
 Any code should be in a feature branch in each of the repositories you update. The name of the feature branch should be in the form `<issue-number>-<readable-name>`, for example `1104-inclusive-export`. Once you're satisfied with your changes:
 
 1. Submit a PR for each of the repositories. Each PR message and description will become the commit message and description so keep the message consise, describe what and why rather than how, and link to the issue in the description (eg: "medic/cht-core#123").
-2. If AT is required update the issue with AT instructions.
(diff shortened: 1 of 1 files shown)


content/en/contribute/code/workflow.md
```diff
@@ -90,9 +90,16 @@ Issues in this column are ready to be acceptance tested by a Quality Assurance e
 
 Issues in this column are in the process of being acceptance tested by a Quality Assurance engineer. To complete AT:
 
+1. Verify the steps to reproduce are accurate
+   1. Update or add comment if steps are not easily understandable. 
 1. Install the PR branch to test against.
 2. If the issue fails AT then notify the original developer and move the issue back to "In progress".
```
abbyad, line 96: Nice work, these are great improvements!

For the first point, how does that differ for bugs vs features/improvements?

Also, would be nice to use 1's throughout while we are at it. 
```suggestion
1. Verify the steps to reproduce are accurate
   1. Update or add comment if steps are not easily understandable. 
1. Install the PR branch to test against.
1. If the issue fails AT then notify the original developer and move the issue back to "In progress".
```

newtewt: That works for me. Wasn't sure if there was a standard we were adhering to since markdown should handle it.  

I'm going to involve @brad1905  and @ngaruko  in the conversation about bug/feature/improvement since it isn't just my thoughts here.

I personally feel the process should be the same regardless of type. The wording might be slightly different to encompass all scenarios.  

EX:  Add or verify steps used to reproduce or execute the bug, feature, or improvement described in the ticket.

abbyad: Ya, that makes sense!

PR #241, opened 2020-07-21
Move "replicating production data locally" from app/guides to core/guides

kennsippell:
I think the need to replicate production data locally is typically used by developers of the CHT core, rather than typically being used by app builders. This PR is just a move from the app/guides section to the core/guides section + update to inbound links.


abbyad:
To me replicating data locally is about debugging production issues, which is about debugging CHT applications. The way we've structured the doc site the division between `apps` and `core` is not based on people or their role, but rather based on the concept of what you are working with: 
- **apps**: dealing with building, deploying, maintaining CHT apps.
- **core**: dealing with developing features on cht-core. From the doc site: "This section provides an overview and reference for development of the Core Framework of the Community Health Toolkit (CHT). "

Given that, I'd prefer keeping this guide in the CHT apps section. If you think the division is confusing, we should talk more about what can be done to make it clearer.

kennsippell:
I have personally only seen this used to test core features with realistic data. But I'm likely wrong.

PR #242, opened 2020-07-21
#6251 Replication limit

latin-panda:
# Description

This PR will add documentation for the new endpoint that returns information about users who replicate document.
 Also updates the monitor endpoint documentation to include a new property about the number of users that exceeded replication limit.

medic/cht-core#6251

# License

### DIFF
--- a/content/en/apps/reference/api.md
+++ b/content/en/apps/reference/api.md
@@ -1045,6 +1045,7 @@ curl http://localhost:5988/api/v1/monitoring
 | `outbound_push.backlog` | Number | Number of changes yet to be processed by Outbound Push. |
 | `feedback.count` | Number | Number of feedback docs created usually indicative of client side errors. |
 | `conflict.count` | Number | Number of doc conflicts which need to be resolved manually. |
+| `replication_limit.count` | Number | Number of users that exceeded the replication limit of documents. |
 
 #### Errors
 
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/api.md
```diff
@@ -1232,3 +1233,44 @@ Content-Type: application/json
...
+
+## Replication Limit
+
+*Added in 3.11.0*
+Returns the quantity of documents that were replicated by each user.
+Accepts filtering by user name, when not provided, it returns all users.
+Supports GET.
+Only allowed for users with "_admin" role.
+
+### GET api/v1/log-replication-limit
```
garethbowen, line 1246: Now that I see this written down I think we should change the endpoint URL and not use the word "log". The log is how it's stored, and not relevant to the people who call this URL.

What do you think about `api/v1/users-doc-count` ?

latin-panda: Fixed.

garethbowen: This line also needs to be updated...

```suggestion
### GET api/v1/users-doc-count
```

latin-panda: Ups fixed.

content/en/apps/reference/api.md
```diff
@@ -1045,6 +1045,7 @@ curl http://localhost:5988/api/v1/monitoring
 | `outbound_push.backlog` | Number | Number of changes yet to be processed by Outbound Push. |
 | `feedback.count` | Number | Number of feedback docs created usually indicative of client side errors. |
 | `conflict.count` | Number | Number of doc conflicts which need to be resolved manually. |
+| `replication_limit.count` | Number | Number of users that exceeded the replication limit of documents. |
```
michaelkohn, line 1048: This is kind of confusing to me. From what I can tell, the `count` that it returns is the number of docs replicated by user, _not_ the number of users that exceeded the limit. Am i missing something?

PR #255, opened 2020-07-29
Initial commit of codeowners files

abbyad:
Codeowners assigned to help with the PR process. The assignments were discussed with @MaxDiz and @michaelkohn.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,27 @@
+# Default owner, used to cover all Hugo and non content files.
+# Unless a later match takes precedence, they will be requested for
+# review when someone opens a pull request.
+*                             @abbyad
+
+# Owner of GitHub Actions
+.github/workflows/            @nomulex
+
+# Owners for specific content in the docs site
+content/en/apps/              @MaxDiz
+content/en/apps/concepts      @MaxDiz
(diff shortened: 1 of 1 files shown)


abbyad:
With such diverse content in the doc site we are looking to make it easier to manage based on what we've learned so far. Given that it is hard for external or infrequent contributors to know who to assign for PRs, using the CODEOWNERS method was deemed by the docs squad to be a good way to have a default reviewer. We hope this will also lead to more consistent content, and make it easier for the doc contributors that "own" sections (eg Nicole, Sam, Michael, Max) to maintain them.

We've set it up with the CODEOWNERS as the _default_ reviewers of PR, and we are not making it mandatory for the codeowner themselves to accept the PR before merge -- meaning that the PR submitter can still select anyone, and that codeowner can re-assign as they wish. Spreading knowledge is important, and the CODEOWNERS are in better position to know who can and should make reviews and hope that they can help re-assign where appropriate. Aside from making it easier to manage the PRs, we think it will also help with the consistency of the content, which is important for the public site.

abbyad:
Thanks, we'll keep an eye on how it's going and figure out ways to continue to improve the process. I think it's a special case for the docs site, and not something we're looking to adopt across all repos, where our workflow is working well as it is.

PR #250, opened 2020-07-25
Update docker image

ronna:
Update Docker image to the latest
see: https://github.com/medic/cht-infrastructure/pull/4

PR #266, opened 2020-08-04
Add link checker in CI 

mrjones-plip:
See #250

### DIFF
--- a/.github/workflows/build-any.branch.yml
+++ b/.github/workflows/build-any.branch.yml
@@ -1,12 +1,13 @@
 on:
   push:
     branches:
-      - '*'
-      - '!master'
+      - 259-add-link-checker-to-ci # todo - remove after testing
+      # - '*' # todo - make sure this isn't commented out when done testing
+      # - '!master' # todo - make sure this isn't commented out when done testing
 name: Build Any Branch
 jobs:
   BuildAnyBranch:
-    name: 
+    name:
     runs-on: ubuntu-18.04
     steps:
       - name: Git checkout
@@ -29,5 +30,28 @@ jobs:
           hugo-version: "0.71.1"
           extended: true
 
+      - name: Install Go & Muffet
+        uses: actions/checkout@v2
+        uses: actions/setup-go@v2
+        with:
+          go-version: '1.14' # Pinning to current version as of Aug 3, 2020 for stability
+      - run: go version
+      - run: go get -u github.com/raviqqe/muffet
+
+      - name: Serve Hugo site
+        run: hugo server
+        run: sleep 5 # taken from another muffet CI recipe - this makes sure hugo is full spun up and ready
+
+      - name: Run Muffet link checkier
+        run: |
+          muffet http://localhost:1313\
+          --ignore-fragments \
+          -exclude ".*github\.com.*|.*demo\.app\.medicmobile\.org.*|.*download\.docker\.com.*|" \
+          -exclude ".*www\.npmjs\.com/org/medic/team/developers.*|.*github\.com/medic/cht-docs/issues/new.*|.*localhost:5984.*" \
+
+      # todo - remove this comment:
+      # muffet -f  -e ".*github\.com.*" -e  ".*demo\.app\.medicmobile\.org.*"  -e  ".*download\.docker\.com.*" -e ".*www\.npmjs\.com/org/medic/team/developers.*" -e ".*github\.com/medic/cht-docs/issues/new.*" -e ".*localhost:5984.*"  http://localhost:1313
+
+      # is this needed now that we build above when running `hugo server`?
       - name: Build
         run: hugo


mrjones-plip:
@abbyad  - tapping you for a review of this PR, but i see @nomulex did the original architecture (and he's back next week), so feel free to pass back to him if that makes more sense!

nomulex:
@mrjones-plip  do you still need a review here. I have some questions on this section that was deleted.
```yaml
- name: Install Node depencies	      
        uses: actions/setup-node@v1	
        with:	
          node-version: '12.x'	
      - run: npm install	
      - run: npm install -g postcss-cli	
      - run: npm i -D autoprefixer
```

README.md
```diff
@@ -16,15 +18,25 @@ To install, follow the [installation instructions for your Operating System](htt
...
    - `hugo server`
    - If there are any errors, check that you have the latest version of Hugo, and are using the extended version.
+1. Preview your site in your browser at: http://localhost:1313/
+1. As you make changes to the site, your browser will automatically reload your local dev site so you can easily and quickly see your changes. 
 
-3. Preview your site in your browser at: http://localhost:1313/
+## Optional Link Checking
 
-## Continuous Deployment
+We validate that all links on the docs site work (do not 404) using a tool called [Muffet](https://github.com/raviqqe/muffet) along with  [Actions](https://github.com/features/actions).  If you're creating a lot of new links, or editing a lot of existing links, you may optionally run Muffet locally before pushing your commits.  This will save a lot of time of pushing a build, waiting for the Action to run, finding you have a broken link, fixing it, pushing a new change etc.
```
abbyad, line 29: Not sure if the double spaces were typos, a typing preference, or both :)

We generally have single spaces after periods so have made that correction in this suggestion. Also, the last sentence was confusing to read at first... tried to make it clearer, but could use yet another pass:

```suggestion
We validate that all links on the docs site work (do not 404) using a tool called [Muffet](https://github.com/raviqqe/muffet) along with [Actions](https://github.com/features/actions). If you're creating a lot of new links, or editing a lot of existing links, you may optionally run Muffet locally before pushing your commits. Running Muffet locally can save time by exposing broken links before pushing a build since you can avoid waiting for the Action to run, finding you have a broken link, fixing it, and pushing a new change.
```

README.md
```diff
@@ -16,15 +18,25 @@ To install, follow the [installation instructions for your Operating System](htt
...
+  - Test the links with the [`muffet.sh`](https://github.com/medic/cht-docs/blob/master/.github/scripts/muffet.sh) script.  If you're in the root of this repo, that'd be: `./.github/scripts/muffet.sh` 
+  
+It should take about 60 seconds depending on your Internet connection. If Muffet returns no output, you have no broken links, congrats! 
+
+_Note_: The `muffet.sh` script here is the identical script we run on GitHub. If you simply run `muffet http://localhost:1313` you will hit GitHub's rate limiting and get lots of [429's](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429). Our script intentionally reduces concurrency and excludes some repetitive GitHub URLs.
+
+## Continuous Deployment
 
-![CHT Documentation Site Build](https://github.com/medic/cht-docs/workflows/CHT%20Documentation%20Site%20Build/badge.svg)
+All changes to `master` branch run a [GitHub action](.github/workflows/ci.yml) to first check for any broken links (per above) and then deploy to the documentation site: [docs.communityhealthtoolkit.org](https://docs.communityhealthtoolkit.org)
```
abbyad, line 42: Would be nice to link "per above" to the proper section and make it easy to the reader to find it.

README.md
```diff
@@ -16,15 +18,25 @@ To install, follow the [installation instructions for your Operating System](htt
...
-
-2. Build your site:
+1. Build your site:
    - `hugo server`
    - If there are any errors, check that you have the latest version of Hugo, and are using the extended version.
+1. Preview your site in your browser at: http://localhost:1313/
+1. As you make changes to the site, your browser will automatically reload your local dev site so you can easily and quickly see your changes. 
 
-3. Preview your site in your browser at: http://localhost:1313/
+## Optional Link Checking
```
abbyad, line 27: Want to make sure that it's clear that _local_ link checking is optional... Is there a better way to bring in the optional sub header just before the install instructions?

```suggestion
## Link Checking [Optional]
```

mrjones-plip:
>  do you still need a review here. I have some questions on this section that was deleted.

@nomulex  - yes, I'd love a technical review please! Marc said he'd review my updates to the readme.

Per your question, those lines to install node dependencies weren't deleted, they were [moved here](https://github.com/medic/cht-docs/pull/266/files#diff-e9f950f17198d3d5e3122a44230a09b9R57). As well, I reformatted them into a single, multi-line `run` statement.

README.md
mrjones-plip: I feel like given this document is entirely about doing dev, that's it's obvious you're doing link checking against local host. We don't say that you install `hugo` to do local dev, nor call out that you're building docs to do local dev, but both cite `localhost:1313`

That said, I don't feel strongly about this ;)  If you think we need to change some language here to call this out, I'm game.

README.md
mrjones-plip: I'd say this old dog is having trouble learning new tricks ;) My typing class 30 years ago (on actual typewriters!!) has me doing double spaces after periods.  The other spaces...are just mistakes?

Either way, thanks for the space fixes.  And thanks for the re-wording - agreed this is clearer what's going on.

nomulex:
Great @mrjones-plip . I see that hugo build now only applies to the master branch. I recall the initial design requirement was for it to happen for all branches as put [here](https://github.com/medic/cht-docs/blob/master/.github/workflows/build-any.branch.yml#L18)  and deploy was for master only.  The idea was to detect build failures early enough.  If this is no longer needed then everything else looks great. Nice Job!

README.md
mrjones-plip: yup yup - agreed! see 05a8355 .

 However, given you're already scrolled down to the bottom of the page, the net effect is that the page doesn't actually move at all.  It'll be nice for copying and pasting deep link though!

mrjones-plip:
> I see that hugo build now only applies to the master branch.

@nomulex  - thanks for the review!  If you look at [line 10](https://github.com/medic/cht-docs/blob/259-add-link-checker-to-ci/.github/workflows/ci.yml#L10) you'll see we actually apply this to all branches. And then we conditionally only apply to master [from 52](https://github.com/medic/cht-docs/blob/259-add-link-checker-to-ci/.github/workflows/ci.yml#L52) on only to master using `if` statements like so:

```yml
- name: Master Branch Only - Install Node 12.x
  if: ${{ github.ref == 'refs/heads/master' }}
  uses: actions/setup-node@v1
  with:
      node-version: '12.x'
```

The effect is that every commit to not master will always have the build checked, including the new `muffet` link checker.  However, only the `master` branch will actually get pushed live (if build doesn't fail of course ;).

> everything else looks great. Nice Job!

Awesome, thanks thanks!  If it all looks good and my explanation above makes sense, feel free to mark this review as complete.

nomulex:
Apologies, I guess I did not ask my clarifying questions clearly. So when I last worked on this the hugo build [here](https://github.com/medic/cht-docs/blob/259-add-link-checker-to-ci/.github/workflows/ci.yml#L42), needed the npm dependencies [here](https://github.com/medic/cht-docs/blob/259-add-link-checker-to-ci/.github/workflows/ci.yml#L57) to run.  [This](https://github.com/medic/cht-docs/blob/master/.github/workflows/build-any.branch.yml#L32) step comes after installing npm dependencies not before  like it is [here](https://github.com/medic/cht-docs/blob/259-add-link-checker-to-ci/.github/workflows/ci.yml#L42)..

mrjones-plip:
Ah - I see, thanks for clarifying! 

Actually the npm dependencies are not needed to run `hugo` and are not needed to run `hugo server&`.  They're only needed for when we run `env HUGO_ENV="production" hugo --minify`. This is because that's the only call that uses the `postcss-cli` and `autoprefixer` npm libraries.

If I'm still missing something (sorry if so!), maybe it'd be easier to hop on a google meet call to sort it out real time? That said, I see it's late where you are, so this can wait 'til tomorrow if need be! No big rush on this.

mrjones-plip:
@abbyad  - I think all the changes you requested to the `readme.md` are tended to - lemme know if not!

abbyad:
@mrjones-plip, I think you make a good point here: https://github.com/medic/cht-docs/pull/266#discussion_r468049199

The instructions listed in the README are for devs... but it is not clear _why_ and _how_ the current headers relate. @MaxDiz what do you think about restructuring it as part of another issue?

mrjones-plip:
> The instructions listed in the README are for devs... but it is not clear why and how the current headers relate. @MaxDiz what do you think about restructuring it as part of another issue?

I actually wonder, since we have [developer instructions in our docs](https://docs.communityhealthtoolkit.org/contribute/code/workflow/), if we should add more to the current "Getting Started" section there? We don't explicitly call out the difference between making edits on GH directly (or in your own fork) vs setting up hugo locally.  

So yeah, my vote would be to spin up a new ticket to spice up the [the Basics section](https://docs.communityhealthtoolkit.org/contribute/docs/workflow/#basics) to include this.  Or, yeah, the readme ;)

mrjones-plip:
@garethbowen , @abbyad & @nomulex  - Much thanks for all your help with this PR!  The first full [build against master](https://github.com/medic/cht-docs/runs/972041073?check_suite_focus=true) seemed to go OK!

MaxDiz:
creates a new issue: https://github.com/medic/cht-docs/issues/275 based on [comment above](https://github.com/medic/cht-docs/pull/266#issuecomment-671622184). Feel free to add to it

Issue #267, opened 2020-08-04
Include headers for mobile view of workflows in reference application


MaxDiz:
In https://github.com/medic/cht-docs/pull/239 we completed an overhaul of the reference app documentation, including replacing static workflow images with HTML versions. The mobile rendering of the those HTML workflows do not currently have headers. Add headers to make them more user friendly.

example:
![image](https://user-images.githubusercontent.com/46463417/89325348-487a7700-d64e-11ea-9d86-07c2df34e264.png)

abbyad:
With bootstrap I couldn't find an easy way to differentiate between these two screen size that are smaller than `xs`, so opted for a 45 degree turn on the arrow, which seemed to work well with both. We can revisit this later if there are concerns with it.

| Right Arrow | SE Arrow |
|---|---|
| ![image](https://user-images.githubusercontent.com/789512/89417547-ce54fb80-d6fc-11ea-9772-e6380ebe1fcc.png) | ![image](https://user-images.githubusercontent.com/789512/89439037-9b6c3100-d717-11ea-8b05-371d3c39a598.png) |

| Right Arrow | SE Arrow |
|---|---|
| ![image](https://user-images.githubusercontent.com/789512/89417603-e462bc00-d6fc-11ea-9bcb-6dd10cb0d13d.png) | ![image](https://user-images.githubusercontent.com/789512/89439199-d2dadd80-d717-11ea-9308-44bca2037560.png) |

PR #268, opened 2020-08-05
Improve workflow headers

abbyad:
Improves header alignment in desktop view, and adds header for all mobile views.

For #267.


mrjones-plip:
Cool cool! I'll hop on this later today.

abbyad:
Thanks, I want to make the path checker a little more tolerant so moving to a draft for now.

mrjones-plip:
All good! Ping me when it's ready.

abbyad:
Ready again, now with a more robust path checker which doesn't dictate file structure.

mrjones-plip:
We have only 2 of these tables in use currently on the "Maternal and Newborn Health Reference App" page ([dev URL](http://localhost:1313/apps/examples/anc/)):

```bash
grep -r ' workflow-table' content/en
content/en/apps/examples/anc.md:{{% workflow-table
content/en/apps/examples/anc.md:{{% workflow-table
```

I see no regressions on desktop or mobile!  On to code/syntax/css review now.

abbyad:
That's a weird one!

I can reproduce it _only_ in a dev setup, as in when file changes are triggering a rebuild of the site. I wasn't able to figure out a way for a fresh `hugo server` after any change to behave in an unexpected way. I would expect `hugo server --disableFastRender` to work the same way as a fresh run and catch errors, but it doesn't. 

This scenario can be reproduced in a simplified form too:

## 1. Create these files

### layouts\shortcodes\error.html
```html
{{ errorf "ERROR: %s" .Position }}
We shouldn't be here! 
First argument: "{{ .Get 0}}"
```

### content\en\test.md (blank)
```md
```

## 2. Run `hugo server --disableFastRender`: Build passes, as expected.

## 3. Edit md file as shown, see build fail, as expected

### content\en\test.md 
```md
This shortcode _should_ prevent builds. {{< test-error arg >}}
```

## 4. Resave the exact same file, see build pass, NOT expected

## 5. In `test.md` change `arg` to `anything`, see build pass, NOT expected

By using the shortcode in the example we can see that this isn't a simple partial/shortcode caching issue -- the page is generated with the shortcode argument.

**Given that our GH Action should catch issues like this, I don't think this needs to block this PR.** 

We can follow up with Hugo via their [discourse forum](https://discourse.gohugo.io/) first, and then [GH issues](https://github.com/gohugoio/hugo/issues).

mrjones-plip:
@abbyad  - thanks for the fix on the comment line wrap and thanks for checking out that crazy file exists checking race condition! 

lgtm

PR #287, opened 2020-08-20
Releasing 3.9.1 and 3.8.2

dianabarsan:
medic/cht-core#6592
medic/cht-core#6593

### DIFF
--- a/content/en/core/overview/supported-software.md
+++ b/content/en/core/overview/supported-software.md
@@ -13,8 +13,8 @@ Once a version is no longer supported it will not receive any further releases a
 
 | Version | Status | Release date | End of life |
 |----|----|----|----|
-| 3.9.x | Current | 19-Jun-2020 | TBA |
-| 3.8.x | Current | 11-Feb-2020 | 19-Sep-2020 |
+| 3.9.x | Current | 20-Aug-2020 | TBA |
+| 3.8.x | Current | 20-Aug-2020 | 20-Nov-2020 |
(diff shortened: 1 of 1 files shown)


content/en/core/overview/supported-software.md
```diff
@@ -13,8 +13,8 @@ Once a version is no longer supported it will not receive any further releases a
 
 | Version | Status | Release date | End of life |
 |----|----|----|----|
-| 3.9.x | Current | 19-Jun-2020 | TBA |
-| 3.8.x | Current | 11-Feb-2020 | 19-Sep-2020 |
+| 3.9.x | Current | 20-Aug-2020 | TBA |
```
dianabarsan, line 16: This changed after we moved to cht-docs and doesn't include every minor we release. 
In this format, when releasing a new minor, do we update the release date or just the EOL date?

garethbowen: Actually I don't think we should update either. I don't want to extend the EOL because if we had a bug every 2 months or so then the minor would never be EOLed. I don't think we should change the release date, but maybe we need to change the column title to "initial release date"?

Given that neither is changing I guess we should also remove the todo item from the patch release issue template...

dianabarsan:
Closing in favor of https://github.com/medic/cht-core/pull/6603

content/en/core/overview/supported-software.md
dianabarsan: Cool. done!

PR #289, opened 2020-08-20
Rapidpro docs guides

michaelkohn:
This adds the RapidPro "Guides" page for Integrations.  It also links the 'features' page to the new 'guides' page.

It still needs some App Services TLC but I'd like to publish this version and have them look at it from there.


michaelkohn:
@mrjones-plip can you have a look and let me know why Mr Muffet doesn't like my link?

mrjones-plip:
yeah - will do! gimme a sec and I'll report back

mrjones-plip:
```
 o__       
 ,>/_       
(*)`(*)
```

w00t! 

only reason to rush it is if it's something systemic I wanna get ahead of it before my day ends.

mrjones-plip:
OK - not systemic \o/  ! The problem reminds of an old WordPress plugin I wrote that [makes every page a 404](http://sidecar.tv/the-404er/). Turns out us humans don't care that the URL in question (`https://rapidpro.io/api/v2/flow_starts`) is indeed returning a 403, but computers do.  That's Just Not OK according to `muffet` (a good thing!):

really a 403

This because RapidPro is doing a slight of hand where when a human goes to the API endpoint and isn't authenticated (403 not allowed) they get what looks like a documentation page (indeed it is).  So....what to do? I think we really have three options (two really):

 - find a different URL to use that's not on `https://rapidpro.io/api/v2`. Maybe the [API explorer](https://rapidpro.io/api/v2/explorer/) or a [higher level "start flow"](https://help.nyaruka.com/en/articles/1911210-starting-a-flow) page?
 - exclude the URL from `muffet` by [updating the exclude section](https://github.com/medic/cht-docs/blob/master/.github/scripts/muffet.sh#L22) to include RapidPro's API pages 
 - something else I'm not thinking of 

Long term I think we should be extremely conservative with adding items to the exclude list and when we do, be extremely granular so as to not hinder future link checking on the same site.

Issue #293, opened 2020-08-27
related content link on building workflows page goes to itself


mrjones-plip:
on the [building workflows page](https://docs.communityhealthtoolkit.org/apps/concepts/workflows/) under "Related Content" there's a link titled "CHT Applications > Concepts > Workflows" that goes to the same page.  We should remove it.

PR #294, opened 2020-08-27
remove link in related

mrjones-plip:
per  #293


mrjones-plip:
@n-orlowski  - with both Max and Marc out are you allowed to approve a docs change do you know?  If not, we'll just wait - no biggie!

mrjones-plip:
Gareth reports that only code owners can approve, so I'll punt this over to Marc and he'll get it some time when he gets back from holiday.

PR #295, opened 2020-08-31
fix see-also for hydration documentation

michaelkohn:
Not rendering properly due to see-also still using `%` instead of `<`

https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#cross-referencing-content

### DIFF
--- a/content/en/core/overview/db-schema.md
+++ b/content/en/core/overview/db-schema.md
@@ -95,7 +95,7 @@ Generally when contacts are **used** in the app they are first "hydrated", with 
 }
 ```
 
-{{% see-also page="apps/guides/data/hydration" title="Document hydration" %}}
+{{< see-also page="apps/guides/data/hydration" title="Document hydration" >}}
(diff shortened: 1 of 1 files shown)


dianabarsan (approved):
Cool. How come the syntax changed?

michaelkohn:
¯\_(ツ)_/¯ Not sure.

mrjones-plip:
@dianabarsan  - when we [implemented the feature](https://github.com/medic/cht-docs/issues/178) to have both "Read More" or "See Also", it was an intentional deprecation. I forget the specifics, but Marc and I were chatting about it and the ability to have the Read More/See Also feature which knowingly broke the `{{%` was decided to be a net win. 

Sorry if we didn't do a good enough job socializing the change after it went live. If there's steps you know we could/should have taken - please lemme know!

garethbowen:
@michaelkohn This can be merged, right?

PR #298, opened 2020-09-01
Remote login

n-orlowski:
Add to concepts, shorten on training example, add see-also shortcode to relevant pages


content/en/apps/examples/training.md
```diff
@@ -42,19 +42,14 @@ The CHT’s onboarding and training capabilities offer a remote way to provide e
 
 ### Remote Login by App
 
-When creating users, the admin has the option to send a user their credentials via SMS using a [magic link](https://hackernoon.com/magic-links-d680d410f8f7). This generates a new, random and complex password with a 24-hour expiry. If no gateway is set up, the message may be sent via another messaging app. 
+When creating users, the admin has the option to send a user their credentials via SMS using a [magic link](https://hackernoon.com/magic-links-d680d410f8f7). This generates a new, random and complex password with a 24-hour expiry.
```
garethbowen, line 45: I think it would be better to drop the phrase "magic link" as well as the hyperlink to "hackernoon.com". I don't expect the people reading this to know what that is, and it's sort of an implementation detail. We can just say "link", right?

content/en/apps/examples/training.md
```diff
@@ -42,19 +42,14 @@ The CHT’s onboarding and training capabilities offer a remote way to provide e
...
 
-![admin-magiclink](admin-magiclink.png)
+By clicking the magic link to log in, the user is able to enter their project's instance directly, bypassing the need to enter their username and password. 
 
-By clicking the magic link to log in, the user is able to enter their project's instance directly, bypassing the need to enter their username and password. If the app is not installed on their phone, it will open in their default browser.
+![magic-link](magic-link.png)
```
garethbowen, line 49: It looks like we have this image twice - can we use the same copy for both so if we update one they both get updated?

content/en/apps/examples/training.md
```diff
@@ -42,19 +42,14 @@ The CHT’s onboarding and training capabilities offer a remote way to provide e
...
 
-![login.png](login.png)
+{{< see-also page="apps/concepts/access/#remote-login" >}}
```
garethbowen, line 51: The content here is pretty much a duplicate of what's in this `see-also` link - why not just link to the other page rather than copying a bunch of the content over?

garethbowen (commented):
I've made a few comments. Also the link checker seems to be failing but I can't tell whether that's due to your changes or not. 

@mrjones-plip Any tips?

mrjones-plip:
>  the link checker seems to be failing but I can't tell whether that's due to your changes or not. 
> @mrjones-plip Any tips?

Oh, this is interesting! This is the error (for posterity):

```bash
 All Branches - Run Muffet link checker
    GOROOT: /opt/hostedtoolcache/go/1.14.7/x64
Run ./.github/scripts/muffet.sh
  ./.github/scripts/muffet.sh
  shell: /bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    GOROOT: /opt/hostedtoolcache/go/1.14.7/x64
dial tcp4 127.0.0.1:1313: connect: connection refused
##[error]Process completed with exit code 1.
```

so that smells like GH Actions thought it started the instance via `hugo` but then when `muffet` tried to reach the server on localhost `127.0.0.1:1313` it failed.  Let's hope it's ephemeral! I'll hit it with the "retry" stick, see what happens and report back.

content/en/apps/examples/training.md
n-orlowski: Yep!

mrjones-plip:
What in the world!? Indeed, I was right that `hugu` had failed, but you wouldn't know it by looking at the GH actions:

server hugo site OK

I happened to be watching and saw the failures in the build, which you can see when you expand the step:

server hugo NOT OK

So there's three bad links on `apps/concepts/access.md`, `apps/examples/training.md` and `apps/reference/api.md`:

* [see-also] REF_NOT_FOUND: Ref "apps/examples/training/": "apps/concepts/access.md": page not found
* [see-also] REF_NOT_FOUND: Ref "apps/concepts/access/#remote-login": "apps/examples/training.md": page not found
*  [see-also] REF_NOT_FOUND: Ref "apps/concepts/access/#remote-login": "apps/reference/api.md": page not found

If you fix those three, hugo should be able to start in GH Actions and link checker should run OK. But hrm - why isn't this being caught locally? Maybe the doc writers `hugo` isn't catching these broken `see-also` links and is building OK?!? I need to test this. 

This is a bug in the GH Action - it should fail on the "All Branches - Serve Hugo site" step but isn't, so it's a false positive that it's failing on the next step, " All Branches - Run Muffet link checker".  If I can reproduce, I'll file a bug to fix this. 

Thanks for calling this out!

mrjones-plip:
Ah, ok! sorry to be so chatty on this ticket, but indeed the build is failing locally. This is good! Now e just need to get GH actions to do the same thing

@n-orlowski be sure your build is working before you commit/push.  This is what I'm seeing locally, let me know if it's not failing for you!

```bash
➜  cht-docs git:(remote-login-updates) ../hugo server 
Building sites … ERROR 2020/09/01 16:12:08 [see-also] REF_NOT_FOUND: Ref "apps/concepts/access/#remote-login": "apps/examples/training.md": page not found
ERROR 2020/09/01 16:12:08 [see-also] REF_NOT_FOUND: Ref "apps/examples/training/": "apps/concepts/access.md": page not found
ERROR 2020/09/01 16:12:09 [see-also] REF_NOT_FOUND: Ref "apps/concepts/access/#remote-login": "apps/reference/api.md": page not found
Built in 776 ms
Error: Error building site: logged 3 error(s)
```

content/en/apps/examples/training.md
n-orlowski: Made this more succinct and kept the meat in the Access page

n-orlowski:
@mrjones-plip thanks for looking into this! Seems to work now that I've removed the "See also" shortcodes..

n-orlowski:
Is the process to add the `see-also` shortcodes after these changes are merged? How do I avoid running into this again?

mrjones-plip:
@n-orlowski  - sweet!  To be clear, this is because, for example,  `apps/examples/training/` isn't a valid link in your see also: `{{< see-also page="apps/examples/training/" >}}`.  This is because the trailing slash `/` denotes you're trying to load `index.md` in that dir (`apps/examples/training/index.md`). If you want to link to `training.md` then you drop the trailing slash and you're good to go!

`{{< see-also page="apps/examples/training" >}}`

So feel free to add them back in if you want (at least this one, I didn't check the other two).

Aaaaandd - when I was testing this, I saw that when I was running `hugo` it barfed the first time I put this intentionally bad `see-also` in, and then came correct when I fixed it.  But then it did NOT barf when i broke it again.  Was this experience such that your build was working when you committed it? That's bad!

n-orlowski:
@mrjones-plip Ahhhh I see!! Thanks for clarifying, I'll add them back. Tbh Hugo has been barfing for a while for me and I've sort of blacked out so I'm not 100% sure but I think that is what happened 😅

content/en/apps/examples/training.md
```diff
@@ -42,19 +42,7 @@ The CHT’s onboarding and training capabilities offer a remote way to provide e
...
-By clicking the magic link to log in, the user is able to enter their project's instance directly, bypassing the need to enter their username and password. If the app is not installed on their phone, it will open in their default browser.
-
-![login.png](login.png)
-
-To recover a password, the user needs to contact the admin so that they may regenerate a new magic link and repeat the workflow. 
-
-{{% alert title="Note" %}}
-The magic link workflow will not work for users who want to use multiple devices or for multiple users on one device.
-{{% /alert %}}
+Users may log into their app with a link sent to them via SMS. The link allows the user to directly enter their app, bypassing the need to enter their username and password.
```
garethbowen, line 45: I think we should add a link from here to the apps/concepts/access.md file.

garethbowen (changes requested):
It looks like the `content/en/apps/examples/training/admin-magic-link.png` image is not used anywhere - can we just delete it now?

mrjones-plip:
@n-orlowski  - yeah, glad that was helpful! I'm also happy debug your local hugo branch for a 100% less barfing if you need any help. If hugo isn't happy locally, you'll push, wait for GH Actions to run, it will fail your build and you'll have to rinse and repeat until it succeeds to build.  This is a pain!

garethbowen (approved):
Awesome!

Issue #311, opened 2020-09-14
Document export and processing telemetry data


mrjones-plip:
## Overview 
With the Sep 11 2020 go live of [CMoHM deployment](https://github.com/medic/config-moh-mali/tree/master/covid-surveillance) (private repo) we were trying to watch the [known performance issue](https://github.com/medic/config-moh-mali/issues/56) (again, private repo) and it came up that there's no documented process in the [docs site](https://docs.communityhealthtoolkit.org/) for how to get the telemetry data out CouchDB for use in external apps/scripts.  Further, it might be nice to have some prototypical usage of the resulting JSON file.

We do have two pages in the docs site related to this:
* where the telemetry data [is in CouchDB](https://docs.communityhealthtoolkit.org/apps/guides/database/#medic-user-username-meta)
* the [keys and what they mean](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/) in the resulting JSON data

## Related

It should be noted that the currently in use script ([`get_users_meta_docs.js`](https://github.com/medic/cht-core/blob/master/scripts/get_users_meta_docs.js)) is looking to be deprecated by exporting the data via couch2pg:

* [Add ability to replicate medic-users-meta database to PostgreSQL](https://github.com/medic/medic-infrastructure/issues/180)
* [Telemetry data not synced to postgres (private repo)](https://github.com/medic/medic-infrastructure/issues/180)
* [Analyse partners telemetry closely](https://github.com/medic/cht-core/issues/5993)

If any of these efforts are going to come to fruition in weeks not months, we may just want to wait and then 

## tl;dr
we should add below info, likely to [the telemetry page](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/) or to a new page linked from there.  The `plot_fp_time.py` is not publicly available (taken from private [medic-impact repo](https://github.com/medic/medic-impact/blob/master/learning_roadmap/telemetry/270_family_planning_visit_duration/README.md)), so we should either remove this reference or write a simple `.py` script to get folks started:

> To get telemetry data for a client, you will need to use [this script](https://github.com/medic/cht-core/blob/master/scripts/get_users_meta_docs.js) from the [cht-core](https://github.com/medic/cht-core/) repository. To run it from the root directory of the repo, use the following command, filling in the admin credentials and server address for the deployment:
> 
> ```
[...]

Issue #299, opened 2020-09-01
GH Action isn't failing on `hugo server` when there's a build error


mrjones-plip:
In PR #298, a commit was made that correctly wouldn't build.  This is what Hugo did when run locally on the developer machine:

```bash
➜  cht-docs git:(remote-login-updates) ../hugo server 
Building sites … ERROR 2020/09/01 16:12:08 [see-also] REF_NOT_FOUND: Ref "apps/concepts/access/#remote-login": "apps/examples/training.md": page not found
ERROR 2020/09/01 16:12:08 [see-also] REF_NOT_FOUND: Ref "apps/examples/training/": "apps/concepts/access.md": page not found
ERROR 2020/09/01 16:12:09 [see-also] REF_NOT_FOUND: Ref "apps/concepts/access/#remote-login": "apps/reference/api.md": page not found
Built in 776 ms
Error: Error building site: logged 3 error(s)
```

This is correct!  However, when GH Actions ran CI, and hit the same error [when running `hugo server`](https://github.com/medic/cht-docs/blob/00f288aecf31f948bf5e9b280dbe0912ef8dd091/.github/workflows/ci.yml#L42), it did not fail as it should have at the  "All Branches - Serve Hugo site" step. This makes the next step " All Branches - Run Muffet link checker" have a false positive:

![server hugo NOT OK](https://user-images.githubusercontent.com/8253488/91915257-e4f46100-ec6e-11ea-9fae-22667dcf8979.png)

We should ensure we fail the build in GH Actions before running `muffet` so doc authors know how to fix their builds.

mrjones-plip:
@n-orlowski  & @garethbowen  - bad builds will now break with in hopefully logical ways instead of [illogical ways like before](https://github.com/medic/cht-docs/pull/298#pullrequestreview-480135106)!  You can see an [example here](https://github.com/medic/cht-docs/runs/1074283936).

tl;dr - there's a new " All Branches - Test For Broken Builds in Hugo" section and it will error with things like `ERROR [see-also] REF_NOT_FOUND: Ref "apps/reference/app-settings/user-roles/": "apps/concepts/users.md": page not found` which I think is clear. Lemme know if we should [update docs](https://docs.communityhealthtoolkit.org/contribute/docs/workflow/) or do any other outreach!

FYI @abbyad .

PR #312, opened 2020-09-14
Add telemetry export steps 

mrjones-plip:
per #311

@ecsalomon  & @sacul-git - when you get a sec, can you [look at what I did](https://github.com/medic/cht-docs/blob/311-add-telemetry-exfiltrating-steps/content/en/apps/guides/performance/telemetry.md#exfiltrating-data-to-json)? I assume y'all know this a lot better than I do ;)  Any feedback very welcome!


sacul-git:
This is great! I feel like having this in the docs will be really helpful for anyone interesting in diving into their telemetry data. I'm wondering if we can omit the SSH and the scp steps to keep it short and sweet, if people are using this to further analyse their telemetry data, they're likely going to want to pull it locally directly.

mrjones-plip:
Thanks for the feedback @sacul-git ! 

I'm open to removing the `ssh` and `scp` steps!  Is it common to be able to remotely access the CouchDB such that those steps aren't needed?  I assumed that CouchDB would either only be listening on 127.0.0.1 or the public IP would be firewalled off from anything that wasn't on it's subnet.  How do you access the MoH Mali telemetry, for example?  

And, yeah, I was smitten at first w/ the use of "Exfiltrate", but the more I used it the more awkward it felt. With your feedback confirming this, I'll convert to a more accessible "Export" which I think is less polarized and more common. Fixed in d02867f .

sacul-git:
>I assumed that CouchDB would either only be listening on 127.0.0.1 or the public IP would be firewalled off from anything that wasn't on it's subnet. How do you access the MoH Mali telemetry, for example?

Not sure if this depends on how CouchDB was set up, but I'm able to exfiltrate (:wink:) the Mali telemetry data locally with the admin credentials without tunnelling in to their instance.

mrjones-plip:
Ah - sweet! That's great to know, thanks. I hope you had your hacker hoodie up though when exfiltrating, as you're for sure right about [the negative connotations](https://en.wikipedia.org/wiki/Data_Exfiltration) ;)  !

I've remove the `ssh` and `scp` steps (and added a call out to use the correct `COUCH_DB_SERVER` in  a3b06d4 and easier to read final version [is still here](https://github.com/medic/cht-docs/blob/311-add-telemetry-exfiltrating-steps/content/en/apps/guides/performance/telemetry.md#export-data-to-json).

ecsalomon:
LGTM! Thanks for documenting this!

I scanned the script quickly because it seems silly to use interactive mode to get all the docs if there another mode.... And it looks like there is a batch mode. I haven't tried it, but maybe that saves a step?

mrjones-plip:
@ecsalomon  - ah good point! I don't have any telemetry data in my dev instance (and am yet to ask around on how to load some in), but passing `--mode batch` in and then piping to a file by appending ` > output.json` at the end, does indeed save a step!  

If we can verify this gets the same thing as the interactive mode, I say we go for it!

ecsalomon:
Checking...

ecsalomon:
Aye, that works and looks nice (sorry for not checking sooner; I don't have any use for telemetry data at the moment, so thought it would be better not to download and delete if it could be avoided).

mrjones-plip:
Awesome ! Thanks a bunch for testing that.  I've reduced it down to 4 steps in 30d58e0 and preview [is still here](https://github.com/medic/cht-docs/blob/311-add-telemetry-exfiltrating-steps/content/en/apps/guides/performance/telemetry.md#export-data-to-json).

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,15 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
 | `dbInfo.idb_attachment_format` | The format of database attachments. |
 | `dbInfo.db_name` | The name of the local database. |
 | `dbInfo.auto_compaction` | Whether or not auto compaction is set. |
-| `dbInfo.adapter` | The database adapter being used. |
\ No newline at end of file
+| `dbInfo.adapter` | The database adapter being used. |
+
+## Export data to JSON
+
+While you can view the telemetry in the database with a tool like [Fauxton](https://couchdb.apache.org/fauxton-visual-guide/), you may also want to export the data to JSON to further analyze or visualize it.  To do so:
```
abbyad, line 104: I'd either remove the fauxton bit, or guide folks to how/where to see telemetry data in fauxton.

```suggestion
{{% alert title="Note" %}}
Telemetry data can be viewed directly in your browser with [Fauxton](https://couchdb.apache.org/fauxton-visual-guide/) at `https://{{CHT_INSTANCE_URL}}/_utils`, and navigating the `medic-users-meta` database.
{{% /alert %}}

Follow the steps below to export all telemetry in JSON for further analysis or visualization:
```

abbyad (changes requested):
Nice work! I left comments inline, but aside from a minor path error they are all stylistic suggestions!

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,15 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
+
+1. `cd` to a directory where you want to run the export script
+1. Ensure all that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
+1. Get a current copy of the script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js`
+1. Using your CouchDB admin username (`USERNAME`), password (`PASSWORD`) and server URL (`COUCH_DB_SERVER`), run this command: `$ node scripts/get_users_meta_docs.js --mode batch --type telemetry https://USERNAME:PASSWORD@COUCH_DB_SERVER/medic-users-meta > telemetry.json`
+
+This will save a file named `telemetry.json` containing all the telemetry data in the current directory. If you do this exporting regularly, you can skip steps 2 and 3.
```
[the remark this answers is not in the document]
mrjones-plip: Agreed! Fixed in b494c5a

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,15 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
+| `dbInfo.adapter` | The database adapter being used. |
+
+## Export data to JSON
+
+While you can view the telemetry in the database with a tool like [Fauxton](https://couchdb.apache.org/fauxton-visual-guide/), you may also want to export the data to JSON to further analyze or visualize it.  To do so:
+
+1. `cd` to a directory where you want to run the export script
+1. Ensure all that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
+1. Get a current copy of the script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js`
+1. Using your CouchDB admin username (`USERNAME`), password (`PASSWORD`) and server URL (`COUCH_DB_SERVER`), run this command: `$ node scripts/get_users_meta_docs.js --mode batch --type telemetry https://USERNAME:PASSWORD@COUCH_DB_SERVER/medic-users-meta > telemetry.json`
```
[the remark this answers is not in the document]
mrjones-plip: cool cool - sample added in 09b4d3b

content/en/apps/guides/performance/telemetry.md
mrjones-plip: Ah - the "note" is a nice touch - thanks! Fixed in 23e949b

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,15 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
-| `dbInfo.adapter` | The database adapter being used. |
\ No newline at end of file
+| `dbInfo.adapter` | The database adapter being used. |
+
+## Export data to JSON
+
+While you can view the telemetry in the database with a tool like [Fauxton](https://couchdb.apache.org/fauxton-visual-guide/), you may also want to export the data to JSON to further analyze or visualize it.  To do so:
+
+1. `cd` to a directory where you want to run the export script
+1. Ensure all that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
```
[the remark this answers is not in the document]
mrjones-plip: I agree that a `package.json` would be good, but I think it should be w/ the `get_users_meta_docs.js` in `cht-core`, so let's skip it for now and just ensure it's in the newly added prerequisites - done! (per b494c5a)

mrjones-plip:
@abbyad - Thanks again for all the feedback! They should all be addressed, except maybe this one?

> The `scripts` directory may not exist with the steps as they are now

did you mean in the `cht-core` [here](https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js), such that we shouldn't do the `curl` call to download it? I wasn't sure what the suggested fix was here.

content/en/apps/guides/performance/telemetry.md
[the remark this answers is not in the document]
abbyad: Sounds good. Fixing a minor typo:

```suggestion
1. Ensure that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
```

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,22 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
+## Export data to JSON
+
+{{% alert title="Note" %}}
+Telemetry data can be viewed directly in your browser with [Fauxton](https://couchdb.apache.org/fauxton-visual-guide/) at `https://{{CHT_INSTANCE_URL}}/_utils`, and navigating the `medic-users-meta` database.
+{{% /alert %}}
+Follow the steps below to export all telemetry in JSON for further analysis or visualization:
+
+1. Open a command line terminal in the directory where you want to save the export files
+1. Ensure all that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
+1. Get a current copy of the script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js`
```
abbyad, line 111: First mention of the script, so worth noting what script it is:

```suggestion
1. Get a current copy of the export script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js` 
```

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,22 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
+{{% alert title="Note" %}}
+Telemetry data can be viewed directly in your browser with [Fauxton](https://couchdb.apache.org/fauxton-visual-guide/) at `https://{{CHT_INSTANCE_URL}}/_utils`, and navigating the `medic-users-meta` database.
+{{% /alert %}}
+Follow the steps below to export all telemetry in JSON for further analysis or visualization:
+
+1. Open a command line terminal in the directory where you want to save the export files
+1. Ensure all that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
+1. Get a current copy of the script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js` 
+
+Then to export the data run this command, replacing `USERNAME`, `PASSWORD`, `COUCH_SERVER` and `COUCH_PORT`  with your own values: `node get_users_meta_docs.js --mode batch --type telemetry https://USERNAME:PASSWORD@COUCH_DB_SERVER:COUCH_PORT/medic-users-meta > telemetry.json`.
```
abbyad, line 113: I would expect this to be step 4. Also minor fixes to keep the vars consistent.
```suggestion
1. To export the data run this command, replacing `USERNAME`, `PASSWORD`, `COUCHDB_SERVER` and `COUCHDB_PORT`  with your own values: `node get_users_meta_docs.js --mode batch --type telemetry https://USERNAME:PASSWORD@COUCHDB_SERVER:COUCHDB_PORT/medic-users-meta > telemetry.json`. 
```

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,22 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
+Follow the steps below to export all telemetry in JSON for further analysis or visualization:
+
+1. Open a command line terminal in the directory where you want to save the export files
+1. Ensure all that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
+1. Get a current copy of the script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js` 
+
+Then to export the data run this command, replacing `USERNAME`, `PASSWORD`, `COUCH_SERVER` and `COUCH_PORT`  with your own values: `node get_users_meta_docs.js --mode batch --type telemetry https://USERNAME:PASSWORD@COUCH_DB_SERVER:COUCH_PORT/medic-users-meta > telemetry.json`. 
+
+For example, if your user was `admin`, your password was `pass`, your server was `localhost` and your port was `5984`, you would run `bash
+node get_users_meta_docs.js --mode batch --type telemetry https://admin:pass@localhost:5984/medic-users-meta > telemetry.json`
```
abbyad, line 116: Keep in present if possible. Also, you use `bash` here, but not in the steps above... is it needed?

```suggestion
Keep in present if possible and code line on its own line. Also, you use `bash` here, but not in the steps above... is it needed?

```suggestion
For example, if your user is `admin`, your password is `pass`, your server is `localhost`, and your port is `5984`, you would run the following:
`node get_users_meta_docs.js --mode batch --type telemetry https://admin:pass@localhost:5984/medic-users-meta > telemetry.json`
```

abbyad (changes requested):
Looking good! Some minor fixes then ready for merge.

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,22 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
 | `dbInfo.auto_compaction` | Whether or not auto compaction is set. |
-| `dbInfo.adapter` | The database adapter being used. |
\ No newline at end of file
+| `dbInfo.adapter` | The database adapter being used. |
+
+## Export data to JSON
+
+{{% alert title="Note" %}}
+Telemetry data can be viewed directly in your browser with [Fauxton](https://couchdb.apache.org/fauxton-visual-guide/) at `https://{{CHT_INSTANCE_URL}}/_utils`, and navigating the `medic-users-meta` database.
+{{% /alert %}}
```
abbyad, line 106: Do you need an empty line after the close of the alert to display well in Hugo?

mrjones-plip: I checked and it works any which way. As it is now:

```
and navigating the `medic-users-meta` database. 
{{% /alert %}}
Follow the steps below to export all telemetry in JSON for further analysis or visualization:
```

With an extra line:

```
and navigating the `medic-users-meta` database. 
{{% /alert %}}

Follow the steps below to export all telemetry in JSON for further analysis or visualization:
```

With no line extra lines:

```
and navigating the `medic-users-meta` database. {{% /alert %}}Follow the steps below to export all telemetry in JSON for further analysis or visualization:
```

Unless there's a best practice I'm missing, I'll keep it as is!

mrjones-plip: Actually...an extra line after the `{{% /alert %}}` does indeed make it more legible in code ;)

mrjones-plip:
Thanks @abbyad! all these should be fixed in  242fc8c.  Also merged master in 8a517e5 to pull in the fix from #299 so that the build won't break any more on GH Actions ;)

content/en/apps/guides/performance/telemetry.md
```diff
@@ -97,4 +97,22 @@ When the aggregate doc is created the Telemetry service also includes a snapshot
...
+
+To export all telemetry in JSON for further analysis or visualization, first meet these three prerequisites:
+
+1. Open a command line terminal in the directory where you want to save the export files
+1. Ensure that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
+1. Get a current copy of the export script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js` 
+
+To export the data run this command, replacing `USERNAME`, `PASSWORD`, `COUCHDB_SERVER` and `COUCHDB_PORT`  with your own values: `node get_users_meta_docs.js --mode batch --type telemetry https://USERNAME:PASSWORD@COUCHDB_SERVER:COUCHDB_PORT/medic-users-meta > telemetry.json`.
```
abbyad, line 114: Opening the terminal doesn't need to be stated as a prerequisite. It is fine to include as a step if running the export is an explicit step 4, but as it is now I would just remove this step all together, and add a note to the export line. Given that there is an example, I don't think you need to explicitly state all the vars either. 

Also, I don't think we have anything in the doc style guide, but it is nice to have commands on a separate line to make it easier to read and copy & paste.

```suggestion
1. Ensure that both [`node`](https://nodejs.org/en/) and [`npm`](https://www.npmjs.com/get-npm) are installed and that the needed `node` libraries are installed: `npm install inquirer pouchdb-core fs path minimist pouchdb-adapter-http`
1. Get a current copy of the export script: `curl -s -o get_users_meta_docs.js https://raw.githubusercontent.com/medic/cht-core/master/scripts/get_users_meta_docs.js` 

To export the data open a terminal in the folder where you want to save the export, and run this command:
 `node get_users_meta_docs.js --mode batch --type telemetry https://USERNAME:PASSWORD@COUCHDB_SERVER:COUCHDB_PORT/medic-users-meta > telemetry.json` 
```

abbyad (changes requested):
One more change for clarification, then ready to go!

mrjones-plip:
Tidied up "first meet these three prerequisites" -> "first meet these prerequisites" and otherwise did feedback in ee3122d 

fix trailing backtick  def5d34

abbyad (approved):

Issue #313, opened 2020-09-15
GH Actions breaks on "Test For Broken Builds"


mrjones-plip:
It appears that all builds are breaking on "All Branches - Test For Broken Builds in Hugo" because `postcss-cli` is failing to be installed in the prior step "All Branches - Install npm modules for Hugo" despite it reporting it's successfully been installed.  Here's an excerpt from [this build](https://github.com/medic/cht-docs/runs/1119898619?check_suite_focus=true) showing it installed the module:

```
 npm install
  npm install -g postcss-cli
  npm i -D autoprefixer

[SNIP]

+ postcss-cli@7.1.2
added 113 packages from 101 contributors in 4.393s

```

But then later we try to run `hugo` to make sure the branch has a valid build, it reports that:

> Error: Loading PostCSS Plugin failed: Cannot find module 'postcss'

```
Run hugo
  hugo
  shell: /bin/bash -e {0}
Error: Loading PostCSS Plugin failed: Cannot find module 'postcss'
Require stack:
- /home/runner/work/cht-docs/cht-docs/node_modules/autoprefixer/lib/utils.js
- /home/runner/work/cht-docs/cht-docs/node_modules/autoprefixer/lib/browsers.js
- /home/runner/work/cht-docs/cht-docs/node_modules/autoprefixer/lib/autoprefixer.js
- /opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/postcss-cli/node_modules/import-from/index.js
- /opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/postcss-cli/node_modules/import-cwd/index.js
- /opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/postcss-cli/node_modules/postcss-load-config/src/options.js
- /opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/postcss-cli/node_modules/postcss-load-config/src/index.js
- /opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/postcss-cli/index.js
- /opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/postcss-cli/bin/postcss

(@/home/runner/work/cht-docs/cht-docs/themes/docsy/postcss.config.js)
    at load (/opt/hostedtoolcache/node/12.18.3/x64/lib/node_modules/postcss-cli/node_modules/postcss-load-config/src/plugins.js:28:11)
[...]

PR #314, opened 2020-09-15
Fix CI postcss error

mrjones-plip:
I'm gonna make a tshirt "@newtewt is my hero" (but let's hope it doesn't [infect all of GH](https://en.wikipedia.org/wiki/Samy_%28computer_worm%29) ;). With newtewt's help, he verified my initial hunch of `autoprefixer` versions upgrading out from under us was indeed the problem.  

Fix was to add a `package.json` and pin the versions. I went ahead pinned both `autoprefixer` and `postcss-cli`.

see  #313

@abbyad  or @nomulex - I don't think we need ya both to do a review.  First come, first served?  I'm easy. 

PR is ready for ya!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,19 +28,13 @@ jobs:
           hugo-version: "0.71.1"
           extended: true
 
-      # Run "build" to catch errors as early as possible in CI
-      - name: All Branches - Show Hugo verison
-        run: hugo version
-
       - name: All Branches - Install Node 12.x
         uses: actions/setup-node@v1
         with:
           node-version: '12.x'
 
       - name: All Branches - Install npm modules for Hugo
-        run: |
-          npm install
-          npm install -g postcss postcss-cli autoprefixer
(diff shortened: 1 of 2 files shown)


abbyad:
I will leave this review to @nomulex so that the PR can benefit of his expertise.

mrjones-plip:
Thanks @abbyad! 

@n-orlowski - while your build is still failing on GH, feel free to work with Max on the review since I see it's not broken when run on local dev machines. I just see that it's Alex's end of day where he is, so we might not get a fix out until tomorrow.

nomulex (approved):
LGTM
May be good to put a comment at the npm install stage  indicating that the packages are installed through packages.json so that if anyone whats to add any package, they go through that or if they want to upgrade the version of a package. Other than that. I see this passed [here](https://github.com/medic/cht-docs/runs/1125667582?check_suite_focus=true) so it's good to go.

mrjones-plip:
Thanks Alex!

PR #315, opened 2020-09-17
concepts

n-orlowski:
Revised images in the Concepts section for consistency and to include desktop instances


mrjones-plip:
The fact that all the images are linked to the full size versions is SUPER helpful.  When I was viewing the app and install screenshots on a phone before, and couldn't click to see the larger version, it was very hard to read.

I did notice that the transparent "click the app!" screenshots are rendering poorly on desktop and mobile.  Here's the desktop version (note grey background with grey place holder icons - may not be visible depending on your monitor):



And then, because I think I have a default dark theme on my phone, the mobile UI (Android 10 Brave) renders it with a black background:

n-orlowski:
@mrjones-plip thanks for flagging! they were rendering differently for me and I didn't realize they were missing a solid background 👍

MaxDiz (approved):
LGTM - Really nice updates to images!

PR #334, opened 2020-10-05
Adds XLSForm meta fields

abbyad:
Taken from https://github.com/medic/medic-docs/blob/91badc2a347f86fd2b42b1c9ab9527b5982ee9c0/configuration/forms.md#supported-xlsform-meta-fields

### DIFF
--- a/content/en/apps/reference/forms/app.md
+++ b/content/en/apps/reference/forms/app.md
@@ -63,8 +63,8 @@ Since writing raw XML can be tedious, we suggest creating the forms using the [X
 
 | element | description |
 |---|---|
-| `start` | A timestamp of when the form entry was started, which is the time when the form is fully loaded. |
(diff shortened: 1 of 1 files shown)


abbyad:
@newtewt, you AT'ed this [back in 3.7](https://github.com/medic/cht-core/issues/5768#issuecomment-528425141), can you please confirm that it still works as expected, and review this docs change?

newtewt:
I know this is a copy from the previous docs. Do you think it would be helpful to define exactly when those time stamps should take place?

abbyad:
That makes sense, I'll update.

abbyad:
Thanks for the comment @newtewt, now ready for review.

newtewt (approved):

garethbowen (approved):

PR #336, opened 2020-10-07
Add telemetry sample payload.

### LINKED ISSUE #202: Add JSON snippet for telemetry data
It's hard to imagine what telemetry data looks like without seeing a snippet. We should add an example somewhere around here:  https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/#

benkags:
Adds a sample of telemetry doc.
Resolves #202


michaelkohn (approved):
LGTM!

Issue #342, opened 2020-10-09
Add new targets tutorial


smbuthia:
Add a new tutorial that shows how to build CHT application targets.

PR #349, opened 2020-10-14
add targets tutorial

smbuthia:
Closes https://github.com/medic/cht-docs/issues/342


content/en/apps/tutorials/targets.md
```diff
@@ -0,0 +1,111 @@
...
+    appliesTo: 'reports',
+    appliesToType: ['assessment'],
+    date: 'now'
+  }
+];
+```
+
+### 2. Define a Monthly Target Widget
+
+Edit the `target.js` file and add another target widget definition object to define a monthly widget as shown below:
```
kitsao, line 63: `targets.js`

kitsao (changes requested):
Like the precision. 1 inline comment.

content/en/apps/tutorials/targets.md
smbuthia: Updated.

kitsao (approved):

PR #350, opened 2020-10-14
adding editable mockups

n-orlowski:
Renamed folders to reflect conversation with @smbuthia for better separation of design tutorials and design resources and added section for editable mockups


smbuthia (approved):
Looks good to me.

PR #351, opened 2020-10-17
Update using-windows.md

ronna:
update the README url from medic repo to cht-core repo (the redirect works fine)


garethbowen (approved):
Thanks!

PR #352, opened 2020-10-17
Update docker-setup.md

ronna:
update medic-os image from medic-os:cht-3.7.0-rc.1 to medic-os:cht-3.9.0-rc.2

update haproxy image from haproxy:rc-1.16 to haproxy:rc-1.17

### DIFF
--- a/content/en/core/guides/docker-setup.md
+++ b/content/en/core/guides/docker-setup.md
@@ -43,7 +43,7 @@ version: '3.7'
 services:
   medic-os:
     container_name: medic-os
-    image: medicmobile/medic-os:cht-3.7.0-rc.1
+    image: medicmobile/ medic-os:cht-3.9.0-rc.2
     volumes:
       - medic-data:/srv
     ports:
@@ -60,7 +60,7 @@ services:
 
   haproxy:
     container_name: haproxy
-    image: medicmobile/haproxy:rc-1.16
+    image: medicmobile/haproxy:rc-1.17
     volumes:
       - medic-data:/srv    
     environment:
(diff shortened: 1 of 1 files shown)


content/en/core/guides/docker-setup.md
```diff
@@ -43,7 +43,7 @@ version: '3.7'
 services:
   medic-os:
     container_name: medic-os
-    image: medicmobile/medic-os:cht-3.7.0-rc.1
+    image: medicmobile/ medic-os:cht-3.9.0-rc.2
```
garethbowen, line 46: I think you added an extra space here that would break.

```suggestion
    image: medicmobile/medic-os:cht-3.9.0-rc.2
```

Also, maybe we should just say use the latest so it never gets out of date again, eg:

```suggestion
    image: medicmobile/medic-os:latest
```

content/en/core/guides/docker-setup.md
```diff
@@ -60,7 +60,7 @@ services:
 
   haproxy:
     container_name: haproxy
-    image: medicmobile/haproxy:rc-1.16
+    image: medicmobile/haproxy:rc-1.17
```
garethbowen, line 63: Same here - just go with the latest I think.

```suggestion
    image: medicmobile/haproxy:latest
```

content/en/core/guides/docker-setup.md
```diff
@@ -149,7 +149,7 @@ In your compose file, change the ports under medic-os:
 services:
   medic-os:
     container_name: medic-os
-    image: medicmobile/medic-os:cht-3.7.0-rc.1
+    image: medicmobile/medic-os:cht-3.9.0-rc.2
```
garethbowen, line 152: And again...

garethbowen (changes requested):
Thanks, very helpful!

I've added some comments (one typo, and a few suggestions) inline.

content/en/core/guides/docker-setup.md
garethbowen, line 152: Just one last change here and it'll be good to go!

```suggestion
    image: medicmobile/medic-os:latest
```

garethbowen (changes requested):
Thanks again, almost there!

garethbowen (approved):
Nice one, thanks!

PR #354, opened 2020-10-22
Added the rest of the notation types

newtewt:
We didn't have these other notation types documented.

### DIFF
--- a/content/en/apps/guides/data/csv-to-docs.md
+++ b/content/en/apps/guides/data/csv-to-docs.md
@@ -66,7 +66,7 @@ Converting that CSV file to JSON docs with the `csv-to-docs` action would genera
 }
 ```
 
-#### Special notations
+### Special notations
 
 ##### Specifying property type
 
@@ -97,7 +97,23 @@ A special column type, `excluded`, is used for excluding a column from the final
 
 This can be useful if using a column for doc references.
 
-#### Including another doc
+#### Available types
+
+| type | outcome |
+| --------------------- | ----------------- |
+| date          | Creates a date using Date() in javascript |
+| rel-date      | Creates a date with the addition of number days to the current date. A negative number results in a future date. |
+| timestamp     | Sets the string passed in as a timestamp number |
+| rel-timestamp | Creates a timestamp that is offset by miliseconds against the current(NOW) timestamp. A negative number results in a future timestamp |
+| int           | Parses the value as an int using Number.parseInt() |
+| bool          | sets boolean value based on string passed |
+| string        | Sets the value as the string being interpurted. Can be omitted if values are only strings. |
+| float         | Sets value to Number.parseFloat() |
+
+
+
+
+### Including another doc
 
 Often times database documents need to include or refer to other documents in the database. This can be achieved with queries across CSV files, which is done by specifying a query in the column header. The query specifies the doc type (`person` or `place`) and matching condition.
 
--- a/content/en/apps/guides/data/csv-to-docs.md
+++ b/content/en/apps/guides/data/csv-to-docs.md
@@ -102,9 +102,9 @@ This can be useful if using a column for doc references.
 | type | outcome |
 | --------------------- | ----------------- |
 | date          | Creates a date using Date() in javascript |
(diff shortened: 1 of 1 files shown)


content/en/apps/guides/data/csv-to-docs.md
```diff
@@ -66,7 +66,7 @@ Converting that CSV file to JSON docs with the `csv-to-docs` action would genera
 }
 ```
 
-#### Special notations
+### Special notations
 
 ##### Specifying property type
```
abbyad, line 71: Updating the level here too:
```suggestion
#### Specifying property type
```

content/en/apps/guides/data/csv-to-docs.md
```diff
@@ -97,7 +97,23 @@ A special column type, `excluded`, is used for excluding a column from the final
...
 This can be useful if using a column for doc references.
 
-#### Including another doc
+#### Available types
+
+| type | outcome |
+| --------------------- | ----------------- |
+| date          | Creates a date using Date() in javascript |
+| rel-date      | Creates a date with the addition of number days to the current date. A negative number results in a future date. |
+| timestamp     | Sets the string passed in as a timestamp number |
```
abbyad, line 106: Is this in ms or s? Worth adding the unites here.

content/en/apps/guides/data/csv-to-docs.md
```diff
@@ -97,7 +97,23 @@ A special column type, `excluded`, is used for excluding a column from the final
...
+| rel-timestamp | Creates a timestamp that is offset by miliseconds against the current(NOW) timestamp. A negative number results in a future timestamp |
```
abbyad, line 107:
```suggestion
| rel-timestamp | Creates a timestamp that is offset by milliseconds against the current(NOW) timestamp. A negative number results in a future timestamp |
```

content/en/apps/guides/data/csv-to-docs.md
```diff
@@ -97,7 +97,23 @@ A special column type, `excluded`, is used for excluding a column from the final
...
+| int           | Parses the value as an int using Number.parseInt() |
+| bool          | sets boolean value based on string passed |
```
abbyad, line 109: Would be good to give the example options for this one:
```suggestion
| bool          | Sets boolean value based on string passed, either `"true"` or `"false"` |
```

content/en/apps/guides/data/csv-to-docs.md
```diff
@@ -97,7 +97,23 @@ A special column type, `excluded`, is used for excluding a column from the final
...
+| string        | Sets the value as the string being interpurted. Can be omitted if values are only strings. |
```
abbyad, line 110: What does it mean that these can be omitted? Is it the default?

```suggestion
| string        | Sets the value as the string being interpreted. Can be omitted if values are only strings. |
```

content/en/apps/guides/data/csv-to-docs.md
```diff
@@ -97,7 +97,23 @@ A special column type, `excluded`, is used for excluding a column from the final
...
+| float         | Sets value to Number.parseFloat() |
```
abbyad, line 111:
```suggestion
| float         | Sets value using Number.parseFloat() |
```

abbyad (changes requested):
Great to have this included!

Some notes inline, and should do an extra pass to review the hierarchy. For instance we should lose a `#` for `##### Excluding column` and `##### Including value from another doc`.

content/en/apps/guides/data/csv-to-docs.md
newtewt: My thought was that Special notations was the section and the others were sub categories underneath that. Do you think they should be retained as the way they are?

abbyad: I think we agree here... Special notations has 3 #, this subsection should have 4, not 5.

newtewt: oh! I was seeing 4 and not 5. I see the problem now.

newtewt:
@abbyad  I believe all requests are resolved now.

abbyad (approved):
Looks great! Nice work on this @newtewt!

PR #359, opened 2020-10-29
update JSON snippet for people API

michaelkohn:
The example JSON snippet specifies  `"type": "contact"` but "contact" isn't valid, it should be `person`.

### DIFF
--- a/content/en/apps/reference/api.md
+++ b/content/en/apps/reference/api.md
@@ -443,7 +443,7 @@ Content-Type: application/json
 {
   "name": "Hannah",
   "phone": "+2548277210095",
-  "type": "contact",
+  "type": "person",
   "contact_type": "patient",
   "place": {
     "name": "CHP Area One",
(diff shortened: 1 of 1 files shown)


garethbowen (changes requested):
I think it was correct before, however this example does rely on a configured hierarchy with a `patient` type.

The modern style is the `type` of people and places should always be "contact", and the `contact_type` should be one of the configured types. Changing `type` to "person" works because of the backwards compatibility with the old style, but then you shouldn't specify a `contact_type` at all.

I think the example should be left as is, but perhaps a note added at the top that the example only works with a configured contact hierarchy which includes a `patient` person type.

michaelkohn:
Ahh gotcha. I haven't seen configurable hierarchy in use yet so I'm not familiar with it (and the [contact_types documentation](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/hierarchy/#app_settingsjson-contact_types) lacks examples/snippets so it's a bit hard to make sense of).  I'll delete this PR and create an issue to improve the documentation to clarify.

Issue #362, opened 2020-11-02
Update docker-compose  install steps in CHT Local Environment Setup


mrjones-plip:
On the [CHT Local Environment Setup page](https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/) we should add some more info:
* What version of CHT core you're installing (currently the 3.9rc1 docker image, but we should bump it to 3.9rc2 and ha proxy to rc-1.17, see docker hub for [medic-os](https://hub.docker.com/r/medicmobile/medic-os/tags) and [haproxy](https://hub.docker.com/repository/docker/medicmobile/haproxy/tags) (NB - do not hard code to `latest` as this could be untested))
* How to handle when port 80 is in use
   ```
   ERROR: for medic-os  Cannot start service medic-os: driver failed programming external connectivity on endpoint medic-os (189d8[SNIP]b1b): Error starting userland proxy: listen tcp 0.0.0.0:80: bind: address already in use
   ```
  Oop - I see this is discussed [here on the Docker Setup page](https://docs.communityhealthtoolkit.org/core/guides/docker-setup/). We should explicitly link to this page for how to resolve port conflicts    
* ~~How to upgrade the instance to latest (currently 3.10) - either through GUI via admin options or through horti~~ - covered already in [the FAQ on this page](https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/#frequently-asked-questions).

While we're in here - let's update both this page and the "[Docker Image Setup](https://docs.communityhealthtoolkit.org/core/guides/docker-setup/)" to use `curl` to get the latest `docker-compose.yml`  This will help [our multiple instances of `docker-compose.yml` problem](https://github.com/medic/cht-docs/issues/317) a bit.

For QA testing steps, [see PR](https://github.com/medic/cht-core/pull/6707).

Issue #317, opened 2020-09-17
Unify our install instructions and docker-compose.yml's


mrjones-plip:
We have multiple locations that tell you how to install CHT core. A problem with having so many locations with instructions is that we need to keep them all current and in sync.  We should consider deleting duplicates and linking to one canonical set of instructions:

* [Developer set up in cht-core](https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md)
* [Self hosted in cht-infrastructure](https://github.com/medic/cht-infrastructure/tree/master/self-hosting)
* [Self hosted in docs](https://docs.communityhealthtoolkit.org/apps/guides/hosting/self-hosting/) - slated to be replaced by cht-infrastructure one above
* [Local Environment Setup in docs](https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/) - `docker-compose.yml` linked to is latest ([aka master](https://github.com/medic/cht-core/blob/master/docker-compose.yml))
* [Local Environment Setup in cht-core](https://github.com/medic/cht-core/blob/d4340f661e325f8d1ad1067926311a0dd111ee67/INSTALL.md)  - `docker-compose.yml` linked to is from Dec 2019 ([aka d4340f661e](https://github.com/medic/cht-core/blob/d4340f661e325f8d1ad1067926311a0dd111ee67/docker-compose.yml))
* [EC2 in docs](https://docs.communityhealthtoolkit.org/apps/guides/hosting/ec2-setup-guide/)

We also have many versions of our `docker-compose.yml` file. These also should be deleted and linked to a single canonical source:

* [medic-docs](https://github.com/medic/medic-docs/blob/ef134df0bb9e5e586319bfe45218fe575ee63643/installation/public-docker-image-setup.md#use-docker-compose) (precursor to cht-docs? can be ignored? says "read only" clearly)
* [cht-docs](https://github.com/medic/cht-docs/blob/033934230ec555a78dadeaedb958477cd77ac571/content/en/core/guides/docker-setup.md)
* [cht-infrastructure](https://github.com/medic/cht-infrastructure/blob/789d3e070d199f3a26cb593876592d923113432b/self-hosting/main/docker-compose.yml)
* [cht-core](https://github.com/medic/cht-core/blob/43782050cd7495f4681c15cfe0a4d4fda7779b66/docker-compose.yml)
[...]

PR #363, opened 2020-11-02
remove hard coded yml file, update steps, add links 

mrjones-plip:
per #362

### DIFF
--- a/content/en/apps/tutorials/local-setup.md
+++ b/content/en/apps/tutorials/local-setup.md
@@ -10,7 +10,7 @@ relatedContent: >
 ---
 
 {{% pageinfo %}}
-This tutorial will take you through setting up a local environment to build and test CHT applications. This includes setting up the necessary tools to download and run the CHT public docker image as well as a command line interface tool to manage and build CHT apps.
+This tutorial will take you through setting up a local environment to build and test CHT applications on CHT version 3.9.1. This includes setting up the necessary tools to download and run the CHT public docker image as well as a command line interface tool to manage and build CHT apps.
 
 By the end of the tutorial you should be able to:
 
@@ -28,6 +28,8 @@ By the end of the tutorial you should be able to:
 *Docker* is a tool designed to make it easier to create, deploy, and run applications by using containers.  
 
 *Containers* allow a developer to package up an application with all of the parts it needs, such as libraries and other dependencies, and deploy it as one package.
+
+To read more about these concepts, see our [Docker Setup guide]({{< relref "core/guides/docker-setup" >}}).
 
 ## Required Resources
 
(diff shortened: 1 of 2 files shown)


mrjones-plip:
FYI @garethbowen  and @smbuthia

content/en/apps/tutorials/local-setup.md
```diff
@@ -10,7 +10,7 @@ relatedContent: >
 ---
 
 {{% pageinfo %}}
-This tutorial will take you through setting up a local environment to build and test CHT applications. This includes setting up the necessary tools to download and run the CHT public docker image as well as a command line interface tool to manage and build CHT apps.
+This tutorial will take you through setting up a local environment to build and test CHT applications on CHT version 3.9.1. This includes setting up the necessary tools to download and run the CHT public docker image as well as a command line interface tool to manage and build CHT apps.
```
newtewt, line 13: Wish we had a way to make this dynamic on the version number.  It's a manual maintenance item and in docs sites that usually means it gets forgotten.

content/en/core/guides/docker-setup.md
```diff
@@ -35,47 +35,10 @@ CPUs: 2
 
 ## Use Docker-Compose:
 
-In the location you would like to host your configuration files, create a file titled <project_name>-medic-os-compose.yml with the following contents:
+In the location you would like to host your configuration files, create a file titled <project_name>-medic-os-compose.yml with the following contents. In this example we're creating a file called `test-docs-medic-os-compose.yml` via `curl`:
```
newtewt, line 39: My only thought is that not every OS you can run docker on has curl. Might add an href and a note about right clicking to save as but it's up to you really.  Also, maybe a ticket for another day. We specify a name for the file. If the assumption is the reader is first time using docker-compose they likely won't have any other compose files. 

`<a href="https://raw.githubusercontent.com/medic/cht-core/master/docker-compose.yml" >Right Click to Save the docker file. </a> `

newtewt (approved):

content/en/core/guides/docker-setup.md
mrjones-plip: yeah - this is totally worth an update.  Same for the [local set up page](https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/) too (deep link to raw yml).  Stay tuned

content/en/apps/tutorials/local-setup.md
mrjones-plip: There's [custom variables in hugo](https://www.mikedane.com/static-site-generators/hugo/variables/) which we could try and leverage. However, maybe when ever we update the docker instance we have a mini release process that updates this text?  Currently it looks to be only in one place.  Alternately, we could say something along the lines of "See our [Supported Versions](https://docs.communityhealthtoolkit.org/core/overview/supported-software/) page" and then denote which version the docker image is on and _this_ to the docker image release process?  That way we'd only have to reference the one `supported-software` page.

 Also, maybe we solve it when we solve when [we unify our install docs](https://github.com/medic/cht-docs/issues/317)?

content/en/core/guides/docker-setup.md
mrjones-plip: Done! see 3cc65ec

newtewt: solid!

content/en/apps/tutorials/local-setup.md
newtewt: works for me.

newtewt (approved):

PR #364, opened 2020-11-03
remove unused Use Kitematic subtitle

mrjones-plip:
there was a subtitle on the [docker page](https://docs.communityhealthtoolkit.org/core/guides/docker-setup/#use-kitematic-gui-for-docker-tools) that had no content.  we should remove it until we have content for it!

### DIFF
--- a/content/en/core/guides/docker-setup.md
+++ b/content/en/core/guides/docker-setup.md
@@ -90,10 +90,6 @@ Clean data volume:
 
 After following the above three commands, you can re-run `docker-compose up` and create a fresh install (no previous data present)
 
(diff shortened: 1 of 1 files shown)


michaelkohn:
I don't know what Kinematic is or what it's used for, but if it's useful content, we should go ahead and create an issue to add it.

mrjones-plip:
Thanks @michaelkohn! I opened [this ticket](https://github.com/medic/cht-docs/issues/365) to track the effort to add content back in at some later date.

michaelkohn (approved):
Tested and confirmed it is no longer there and didn't break anything else ;-)

Issue #367, opened 2020-11-03
Remove per commit link checking


mrjones-plip:
Back in #259 , we added a link checker that ran in GitHub Actions on every commit to a PR.  This is great, as it catches bad links before they go live. However, it also catches bad links unrelated to the PR, which then the author is blocked on fixing a link they may know nothing about.  Further, the link checker sometimes has a false positive on a link and simply re running it "fixes" the "bad" link.

The solution then, per a suggestion from @garethbowen , is to remove the per PR link checking.  Then at a later date we can add back in [nightly link checking](https://github.com/medic/cht-docs/issues/366).

Issue #366, opened 2020-11-03
Add nightly link checking


mrjones-plip:
Now that we've [removed per commit link checking](https://github.com/medic/cht-docs/issues/367) on PRs, we should add nightly link checking on the entire repo.  This will enable us to catch [link rot](https://en.wikipedia.org/wiki/Link_rot) overtime. In the original link checker effort (see #259), we used `muffet`.  This is still a good tool!  We should just run them [on a schedule](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#scheduled-events) with something like:

```yaml
on:
  schedule:
    # * is a special character in YAML so you have to quote this string
    # this example runs at 3:15am every day:
    - cron:  '15 3 * * *'
```

What we have right now runs on every push:

```yaml
on:
  push:
    branches:
      - '*' # branch checking is below so all this is one CI file
```

And then we check for if we're on master or not to only run certain parts of the CI workflow:

```yaml
- name: Master Branch Only - Deploy to GH pages
   if: ${{ github.ref == 'refs/heads/master' }}
```

Check out the resulting PR from #367 as that will show what we need to add back in to the nightly runs.

mrjones-plip:
This is what I yanked out of `ci.yml` in #365 :

```yaml

      # Pinning to current version as of Aug 3, 2020 for stability
      - name: All Branches - Install Go v1.14
        uses: actions/setup-go@v2
        with:
          go-version: '1.14'

      # Pinning to current version as of Aug 3, 2020 for stability
      - name: All Branches - Install Muffet v1.3.3
        run: env GO111MODULE="on" go get github.com/raviqqe/muffet@1.3.3

      # run in background so CI doesn't hang waiting for "ctrl + c".  Sleep so hugo server is ready
      - name: All Branches - Serve Hugo site
        run: |
          hugo server&
          sleep 5

      - name: All Branches - Run Muffet link checker
        shell: bash
        run: ./.github/scripts/muffet.sh
```

mrjones-plip:
When ever this gets tended to, we should likely refactor the filenames so that they're logical.  Maybe move `ci.yml` to be `branch-checks.yml` and the new one is `nightly-link-check.yml` or something?  Just make it logical ;)

mrjones-plip:
@MaxDiz and @abbyad  - while i've actively yanked out link checking becaues it's blocked some folks from merging PRs - I'm letting this one naturally fall into when ever we want to schedule it.

[...]

PR #368, opened 2020-11-03
remove per PR link checking 

mrjones-plip:
per #367.  We'll add it back in on #366

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,26 +40,6 @@ jobs:
       - name: All Branches - Test For Broken Builds in Hugo
         run: hugo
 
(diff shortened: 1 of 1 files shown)


mrjones-plip:
@nomulex  - one for old times sake ;)  Should be easy as we're just not doing stuff now (we'll do more later ;)

nomulex (approved):
This looks okay to me. 
For the link checker nightly build, we  may need to put it  retry logic for those links that work on retry , also if it fails say 3 times we need to  have a way to alert those that are concerned.

mrjones-plip:
@nomulex  - awesome, thanks for the review.

PR #389, opened 2020-12-11
Quintiles config

newtewt:
Added how to configure the quintiles recording to the people profile.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The documentation is built using Markdown pages, which can be converted into a n
 
 ## Installing Hugo
 
-The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.71.1.
+The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.53 and later.
 
 To install, follow the [installation instructions for your Operating System](https://gohugo.io/getting-started/installing/), and be sure to get the extended version.
 
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The documentation is built using Markdown pages, which can be converted into a n
 
 ## Installing Hugo
 
-The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.53 and later.
+The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.71.1.
 
 To install, follow the [installation instructions for your Operating System](https://gohugo.io/getting-started/installing/), and be sure to get the extended version.
 
--- a/content/en/apps/guides/forms/wealth-quintiles.md
+++ b/content/en/apps/guides/forms/wealth-quintiles.md
@@ -20,23 +20,9 @@ Note that the `can_write_wealth_quintiles` permission is needed for the user tha
 
 | type | name | label | relevant | appearance | calculate | ... |
 |---|---|---|---|---|---|---|				
-|begin group |	inputs |	Patients |||		./source = 'user'	field-list
-|string |	source |	Source | | hidden
-|string |	source_id |	Source ID | hidden |	
-|begin group |	contact |	NO_LABEL|				
-|db:person |	_id |  |	 |		db-object 
-|begin | group |	parent | NO_LABEL 			
-|begin group |	contact	| NO_LABEL		
-|string |	phone |	CHW Phone | | hidden
-|string |	name |	CHW Name | | hidden	
-|end group |						
-|end group |					
-|end group |						
-|end group |						
-|integer|	NationalQuintile|	National Quintile|		
-|integer|	UrbanQuintile	| Urban Quintile				
-|calculate|	place_id|	ID|	||			../inputs/contact/_id|
-
+| string | place_id | | | select-contact type-household
+| integer | NationalQuintile | National Quintile		
+| integer | UrbanQuintile | Urban Quintile
 
 ### Sample Report
 
--- a/content/en/apps/guides/forms/wealth-quintiles.md
+++ b/content/en/apps/guides/forms/wealth-quintiles.md
@@ -8,7 +8,7 @@ relatedContent: >
 
 ---
 
-Introduced in 2.16.0
+_Introduced in v2.16.0_
 
 Household surveys with questions about the home, possessions, and access to safe drinking water have been used to create equity score and improve the targeting of health services. This guide will cover how quintile information from a household survey can be used in customizing care for individual household members. For example, the equity quintile can be used to increase the number of pregnancy follow-ups for women in specific households, or to display specific notes or questions within patient forms.
 
--- a/content/en/apps/guides/forms/wealth-quintiles.md
+++ b/content/en/apps/guides/forms/wealth-quintiles.md
@@ -1,0 +1,61 @@
+---
+title: "Tracking Wealth Quintiles"
+linkTitle: "Quintiles"
+weight: 
+description: >
+  How to track wealth quintiles on the profile of each family member in the household
+relatedContent: >
+
+---
+
+
+A few things are needed to get the data copied onto the family members. 
+
+1. A report submitted with either `doc.fields.NationalQuintile` or `doc.fields.UrbanQuintile` or both. 
+1. An offline user that has the permission of `can_write_wealth_quintiles`. 
+1. Once the form is submitted the contact will then get the fields `"wealth_quintile_national"` and ` "wealth_quintile_urban"` if they are provided in the form.
+
+
+### Sample Report
+
+
+``` json
+{
+  "form": "family_survey",
+  "type": "data_record",
+  "content_type": "xml",
+  "fields": {
+    "place_id": "23c7ab6e-6ea5-4f9f-9ffa-f00a1e30278d",
+    "place_name": "Medic Area",
+    "NationalQuintile": "5",
+    "UrbanQuintile": "3"
+  }
+}
+```
+
+### Sample Contact
+
+``` json 
+{
+  "_id": "c148d0c7-b200-47f6-a891-3f55c78eda9b",
+  "_rev": "3-746a7c8ecb15dc1f43bfa13eba2afbbe",
+  "type": "person",
+  "name": "Jane Smith",
+  "notes": "",
+  "sex": "female",
+  "date_of_birth_method": "approx",
+  "date_of_birth": "1987-12-10",
+  "phone": "",
+  "alternate_phone": "",
+  "external_id": "",
+  "reported_date": 1607530340715,
+  "parent": {
+    "_id": "23c7ab6e-6ea5-4f9f-9ffa-f00a1e30278d",
+    "parent": {
+      "_id": "8ed3600f-23e2-4446-8ff4-2528067085e7"
+    }
+  },
+  "wealth_quintile_national": "5",
+  "wealth_quintile_urban": "3"
+}
+```
--- a/content/en/apps/guides/forms/wealth-quintiles.md
+++ b/content/en/apps/guides/forms/wealth-quintiles.md
@@ -8,6 +8,7 @@ relatedContent: >
 
 ---
 
+Introduced in 2.16.0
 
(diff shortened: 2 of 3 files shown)


content/en/apps/guides/forms/wealth-quintiles.md
```diff
@@ -0,0 +1,61 @@
...
+
+---
+
+
+A few things are needed to get the data copied onto the family members. 
+
+1. A report submitted with either `doc.fields.NationalQuintile` or `doc.fields.UrbanQuintile` or both. 
+1. An offline user that has the permission of `can_write_wealth_quintiles`. 
+1. Once the form is submitted the contact will then get the fields `"wealth_quintile_national"` and ` "wealth_quintile_urban"` if they are provided in the form.
+
```
abbyad, line 17: Additional context would be helpful here. Also, I think this can be simplified a bit:

```suggestion
Household surveys with questions about the home, possessions, and access to safe drinking water have been used to create equity score and improve the targeting of health services. This guide will cover how quintile information from a household survey can be used in customizing care for individual household members. For example, the equity quintile can be used to increase the number of pregnancy follow-ups for women in specific households, or to display specific notes or questions within patient forms.

For quintile information from a household survey to be used for people in that household, a form about a place must contain `NationalQuintile` and/or `UrbanQuintile` fields at the top level. When that form is submitted all the people belonging to that place will get the corresponding `wealth_quintile_national` and `wealth_quintile_urban` fields. Those fields can then be used by forms, tasks, and contact profiles accordingly.

Note that the `can_write_wealth_quintiles` permission is needed for the user that is submitting the form.
```

content/en/apps/guides/forms/wealth-quintiles.md
```diff
@@ -0,0 +1,61 @@
...
+
+``` json
+{
+  "form": "family_survey",
+  "type": "data_record",
+  "content_type": "xml",
+  "fields": {
+    "place_id": "23c7ab6e-6ea5-4f9f-9ffa-f00a1e30278d",
+    "place_name": "Medic Area",
```
abbyad, line 29: The `place_name` isn't required and it may be confusing here.
```suggestion
```

README.md
```diff
@@ -10,7 +10,7 @@ The documentation is built using Markdown pages, which can be converted into a n
 
 ## Installing Hugo
 
-The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.53 and later.
+The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.71.1.
```
abbyad, line 13: Should this be in a separate PR? Also, what brought on this change? Are versions after 71.1 suitable too?

content/en/apps/guides/forms/wealth-quintiles.md
```diff
@@ -0,0 +1,61 @@
...
+
+A few things are needed to get the data copied onto the family members. 
+
+1. A report submitted with either `doc.fields.NationalQuintile` or `doc.fields.UrbanQuintile` or both. 
+1. An offline user that has the permission of `can_write_wealth_quintiles`. 
+1. Once the form is submitted the contact will then get the fields `"wealth_quintile_national"` and ` "wealth_quintile_urban"` if they are provided in the form.
+
+
+### Sample Report
+
```
abbyad, line 20: I think a sample XLSForm would be nice here.

abbyad (changes requested):
Nice work, it is great to have this included!

I added some comments to try to bring a bit more context in.

content/en/apps/guides/forms/wealth-quintiles.md
```diff
@@ -0,0 +1,61 @@
...
+``` json
+{
+  "form": "family_survey",
+  "type": "data_record",
+  "content_type": "xml",
+  "fields": {
+    "place_id": "23c7ab6e-6ea5-4f9f-9ffa-f00a1e30278d",
+    "place_name": "Medic Area",
+    "NationalQuintile": "5",
+    "UrbanQuintile": "3"
```
garethbowen, line 31: Ideally would these be strings, numbers, or ints?

README.md
newtewt: We build the app on github actions using 71.1 but we state .53 as what we support. I believe this should match so that someone developing doesn't get confused when a feature doesn't work in older or newer version.  There's also a bug or improvement in versions earlier than 71.1 that causes the error below. 

```
Building sites … ERROR 2020/10/01 10:52:02 [en] REF_NOT_FOUND: Ref "apps/features/admin/": "~/dev/cht-docs/content/en/apps/guides/integrations/rapidpro.md:82:222": page not found
ERROR 2020/10/01 10:52:02 [en] REF_NOT_FOUND: Ref "apps/reference/app-settings/": "~/dev/cht-docs/content/en/apps/reference/api.md:701:44": page not found
Built in 808 ms
Error: Error building site: logged 2 error(s)
```

newtewt: It could be a separate PR if need be.

content/en/apps/reference/forms/app.md
```diff
@@ -160,6 +160,10 @@ The data used by this function needs to be added to CouchDB. The example below s
 }
 ```
 
+### CHT Special Fields
+
+Set either `NationalQuintile` and/or `UrbanQuintile` on a form to associate this values to the people where the form was submitted. These fields should direct descendants of the fields on a report.
```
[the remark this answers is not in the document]
newtewt: I am taking this suggestion but removing the bit about it being a "place" form.

content/en/apps/guides/forms/wealth-quintiles.md
newtewt: 👍

content/en/apps/guides/forms/wealth-quintiles.md
newtewt: I'm not 100% sure as I just used a form that was already built. I'm going to generate a sample form as requested above and try the different data types.

newtewt: @garethbowen  setting the type on the xlsform doesn't set the actual type. It always seems to be a string. I had prompted a question about storing the specific type to the report but did not get an answer. Do we have a way to store the data in a specific type?

content/en/apps/guides/forms/wealth-quintiles.md
newtewt: added

content/en/apps/reference/forms/app.md
```diff
@@ -160,6 +160,10 @@ The data used by this function needs to be added to CouchDB. The example below s
 }
 ```
 
+### CHT Special Fields
```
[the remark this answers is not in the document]
newtewt: updated to this

content/en/apps/guides/forms/wealth-quintiles.md
garethbowen: Not sure... Don't let me block this PR. If I find a way I can update later...

README.md
abbyad: Noting which version we use for CI is nice, and would be good to be excplicit about that so that people could match up their local environment accordingly. I think that is separate from the minimum version needed to build locally though. Would it work to recommended 0.71.1 since that is what we use for CI, but the minimum for local install is 0.53?

As for those errors, I think if you drop the trailing slash those references should work.

newtewt: removed from this pr

newtewt:
@abbyad  could you have another review?

michaelkohn:
I have a few questions...

- Seems like it's the first time I've seen us purposely use camel case field naming convention. Is that because it is a "CHT Special Field" or is there other reasoning/standard behind that?
- I assume the copying down to people all happens client side, not server side, correct? (ie it's not a sentinel thing)
- Do we have any content/documentation/sample forms/links covering how quintiles are normally calculated? 
- Technically speaking, do we allow _any_ data type for the quintile value? And I assume that if the value on the survey evaluated to NULL or empty string, it just get's copied to all the family members.
- While I understand this is mostly intended for households/families, technically speaking it works at any level of the hierarchy, correct? So if you had a form at the CHW Area level with these special fields, all `person` records whose `parent` is that CHW Area would get updated.
- Do we give any (soft?) errors when uploading a form with `NationalQuintile` / `UrbanQuintile` but there are no users with the `can_write_wealth_quintiles` permission?

michaelkohn:
Also... while I've seen a number of different ways to approach this, I like it when we include when (ie what version of cht-core) the feature was introduced. Here are some examples:

**_Introduced in v3.4.0_** [Telemetry](https://docs.communityhealthtoolkit.org/apps/guides/performance/telemetry/)
**_Only available in 3.7.0 and above_** [Purging](https://docs.communityhealthtoolkit.org/apps/guides/performance/purging/)
**_Version Notes_** [DHIS2](https://docs.communityhealthtoolkit.org/apps/features/integrations/dhis2/#version-notes)
**_In x.x.x_** [.outbound](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/outbound/#in-39)

For quintile stuff, I'd probably just go with "Introduced in..." for now, unless there is already a backlog of known improvements or even ideas on how this feature might evolve.

newtewt:
@michaelkohn 

> Seems like it's the first time I've seen us purposely use camel case field naming convention. Is that because it is a "CHT Special Field" or is there other reasoning/standard behind that? 

I believe this might have been before we implemented any sort of standard. It was first implemented in 2.16

> I assume the copying down to people all happens client side, not server side, correct? (ie it's not a sentinel thing)

Yes this is client side. 

> Do we have any content/documentation/sample forms/links covering how quintiles are normally calculated?

Their is a lot of work done to calculate this info but I don't believe this feature is actually used much if at all. I'd leave this question to be answered by another person.

> Technically speaking, do we allow any data type for the quintile value? And I assume that if the value on the survey evaluated to NULL or empty string, it just get's copied to all the family members.

This feature can be thought of as seeing a field value on a report, copy that to a field on the contact. There isn't much logic otherwise.  As far as I can tell if the value was null or blank string it would get copied. 

> While I understand this is mostly intended for households/families, technically speaking it works at any level of the hierarchy, correct? So if you had a form at the CHW Area level with these special fields, all person records whose parent is that CHW Area would get updated.

Yes I believe this is correct, if the person submitting is offline and has the role, they would update.

> Do we give any (soft?) errors when uploading a form with NationalQuintile / UrbanQuintile but there are no users with the can_write_wealth_quintiles permission?

We do not.  This feature is very basic in what it does. 

As for versioning, we do that a lot for newer features but not so much for things that have been around for a long time. I can add that note to this but it seems there is some gap on a standard for this.  I personally like when documentation sites are versioned with the code they document. Then there is no remembering to add a note. Also would warrant the question of whether we need to keep those "Introduced In" notes if they're outside the range of supported versions.

content/en/apps/guides/forms/wealth-quintiles.md
```diff
@@ -0,0 +1,82 @@
...
+title: "Tracking Wealth Quintiles"
+linkTitle: "Quintiles"
+weight: 
+description: >
+  How to track wealth quintiles on the profile of each family member in the household
+relatedContent: >
+
+---
+
+Introduced in 2.16.0
```
abbyad, line 11: Other pages have this text italicized, so doing the same here for consistency.

```suggestion
_Introduced in v2.16.0_
```

content/en/apps/guides/forms/wealth-quintiles.md
```diff
@@ -0,0 +1,82 @@
...
+|string |	phone |	CHW Phone | | hidden
+|string |	name |	CHW Name | | hidden	
+|end group |						
+|end group |					
+|end group |						
+|end group |						
+|integer|	NationalQuintile|	National Quintile|		
+|integer|	UrbanQuintile	| Urban Quintile				
+|calculate|	place_id|	ID|	||			../inputs/contact/_id|
+
```
abbyad, line 39: I think it would be clearer to replace the boilerplate with a select field.
```suggestion
| string | place_id | | | select-contact type-household
| integer | NationalQuintile | National Quintile		
| integer | UrbanQuintile | Urban Quintile
```

abbyad (changes requested):
Looking good! A couple minor changes suggested, then seems good to go.

newtewt:
Committed those changes @abbyad

abbyad (approved):
Looks great, nice work!

PR #391, opened 2020-12-14
Change countdown field type from text to note

mrjones-plip:
On our [form widget's docs](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#cht-xform-widgets), we say:

> **Countdown Timer**: A visual timer widget that starts when tapped/clicked, and has an audible alert when done. To use it create a `string` field with an `appearance` set to `countdown-timer`.

I find this to be incorrect. The net result is that you have an unused `text` input field like so:

timer extra input

I find it better to not have this input and just make it a `note` instead:

timer no extra

This was discovered while designing the [MNCH app](https://github.com/medic/cht-core/issues/6727), example [.xlsx here](https://docs.google.com/spreadsheets/d/1NF3ys70X0JWM6wIGnlIklM0v05YVjJzs0LJf9KV3JRY/edit#gid=1910514219&range=B71).


mrjones-plip:
@binokaryg - do you have experience deploying this timer? I thought you might such that you'd be a good person to review this PR.  Let me know if you don't - no big deal - I can grab some one else!

binokaryg (approved):
I can confirm that I have only used the `note` field for a countdown timer before.

mrjones-plip:
Awesome - thanks @binokaryg !

PR #393, opened 2020-12-15
minor corrections and changes to grammar for contact & users-1 tutorial

andy5995:


andy5995:
I noticed on this page that sometimes "right hand"  and "left hand" is hyphenated, and sometimes not. Which is preferred?

smbuthia (approved):
Looks good to me. Thank you for contributing the improvements!

The preference on whether to hyphenate or not has not been documented in the [style guide](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#language) yet and I think it would be a good addition.

Issue #397, opened 2020-12-18
[style guide] whether or not to hyphenate right and left hand


andy5995:
Whether or not to hyphenate right and left hand (side) recently came up in https://github.com/medic/cht-docs/pull/393 @smbuthia suggested it should be documented.

I assume you'd want "left-hand" and "right-hand" for cases such as these?

>You should see the newly created Health Facility appear on the left hand side and when you select it, you will see details of the Health Facility appear on the right hand side.

>On the right hand you should see the CHW Area name, the primary contact of the CHW Area, and the Health Facility that the CHW Area belongs to.

https://www.merriam-webster.com/dictionary/right-hand

I'll open up a PR for this.

MaxDiz:
Thanks Andy! Merged via https://github.com/medic/cht-docs/pull/398

PR #398, opened 2020-12-18
sets a guideline for using 'right-hand' or 'left-hand'

andy5995:
for #397


MaxDiz (approved):
nice one! Thanks Andy

PR #403, opened 2020-12-27
hyphenates left- and right-hand

andy5995:
(completes #397)


MaxDiz (approved):
LGTM! Thanks @andy5995

PR #408, opened 2021-01-12
supervision index fix

MaxDiz:
Images were not rendering correctly and noticed error in naming convention of index.


n-orlowski (approved):
LGTM :)

PR #409, opened 2021-01-12
Grammatical fixes and improvements

garethbowen:
I was triggered by the incorrect use of "it's" then got carried away.


mrjones-plip (approved):
`it's` vs `its` ftw!  lgtm

Issue #411, opened 2021-01-15
Document workaround to "fatal error: pipe failed" on MacOS


mrjones-plip:
Some MacOS users are experiencing [this issue](https://github.com/google/docsy-example/issues/89) when they try and launch `hugo server`:

```
hugo server
Start building sites … 

                   | EN | NO  
-------------------+----+-----
  Pages            | 47 | 76  
  Paginator pages  |  0 |  0  
  Non-page files   |  3 |  1  
  Static files     | 38 | 38  
  Processed images | 10 |  2  
  Aliases          |  4 |  0  
  Sitemaps         |  2 |  1  
  Cleaned          |  0 |  0  

Built in 3306 ms

fatal error: pipe failed
```

The work around is to increase the number of file handlers:

```
sudo launchctl limit maxfiles 65535 200000
ulimit -n 65535
sudo sysctl -w kern.maxfiles=100000
sudo sysctl -w kern.maxfilesperproc=65535
```

We should add this to the [readme page](https://github.com/medic/cht-docs/blob/master/README.md) so folks know about the work around. 

As well, we may consider offering a bash file with the [`.command` suffix](https://stackoverflow.com/a/5126257) for users who perfer a double click approach:

```shell
#!/bin/bash
cd /Users/USER_NAME_HERE/Documents/Github/cht-docs
sudo launchctl limit maxfiles 65535 200000
ulimit -n 65535
sudo sysctl -w kern.maxfiles=100000
sudo sysctl -w kern.maxfilesperproc=65535
hugo server
```

PR #412, opened 2021-01-15
Update readme to address macos error 

mrjones-plip:
per #411

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,33 @@ To install, follow the [installation instructions for your Operating System](htt
 1. Preview your site in your browser at: http://localhost:1313/
 1. As you make changes to the site, your browser will automatically reload your local dev site so you can easily and quickly see your changes. 
 
+## MacOS Users
+
+MacOS users encountering the `fatal error: pipe failed` error when running `hugo server` need to run these commands:
+
+```shell
+sudo launchctl limit maxfiles 65535 200000
+ulimit -n 65535
+sudo sysctl -w kern.maxfiles=100000
+sudo sysctl -w kern.maxfilesperproc=65535
+```
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -24,6 +24,33 @@ To install, follow the [installation instructions for your Operating System](htt
...
+sudo launchctl limit maxfiles 65535 200000
+ulimit -n 65535
+sudo sysctl -w kern.maxfiles=100000
+sudo sysctl -w kern.maxfilesperproc=65535
+hugo server
+```
+
+Save this in file with the `.commmand` suffix (e.g. `cht-docs-server.command`) to enable an easy double clicking to start the server. 
+
+Note - be sure to enter your MacOS user password in the terminal when prompted.
```
abbyad, line 52: Any time I see "Note [...]" I think it is worth formatting as a note (or rewording).

abbyad: But then I realize sometimes that its is a GH readme, and not for the documentation site!

abbyad (changes requested):
Good to have this included!

I am in favour of easily digestable readmes, and concerned that this new section is given a lot of visual importance. I'd suggest having a note in the `Building the Documentation` section that mentions the error and links to a troubleshooting file with the workaround.

README.md
mrjones-plip: ;)

mrjones-plip:
Excellent feedback to have this take up less visual feedback - thanks! I've updated this PR, but you can easily see the in the [branch's readme here](https://github.com/medic/cht-docs/tree/411-document-pipe-failed-macos-error#building-the-documentation).

abbyad (approved):
Looks great like this!!

Issue #415, opened 2021-01-22
Improve 'tasks.js'


abbyad:
Page: /apps/reference/tasks/
Version: 2020-10-19 08:47:55 -0700 -0700
---

`resolvedIf` missing from https://docs.communityhealthtoolkit.org/apps/reference/tasks/#tasksjs-1

Also there is a typo: `use functions kepts in a separate`

garethbowen:
Merged.

PR #416, opened 2021-01-22
Fixes required fields and examples

abbyad:
For #415

### DIFF
--- a/content/en/apps/reference/tasks.md
+++ b/content/en/apps/reference/tasks.md
@@ -21,7 +21,7 @@ Tasks are configured in the `tasks.js` file. This file is a JavaScript module wh
 
 | property | type | description | required |
 |---|---|---|---|
-| `name`| `string` | A unique identifier for the task. Used for querying task completeness. | no |
+| `name`| `string` | A unique identifier for the task. Used for querying task completeness. | yes, unique |
 | `icon` | `string` | The icon to show alongside the task. Should correspond with a value defined in `resources.json`. | no |
 | `title` | `translation key` | The title of the task (labeled above). | yes |
 | `appliesTo` | `'contacts'` or `'reports'` | Do you want to emit one task per report, or one task per contact? This attribute controls the behavior of other properties herein. | yes |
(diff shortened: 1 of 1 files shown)


mrjones-plip (approved):
lgtm!

content/en/apps/reference/tasks.md
```diff
@@ -21,7 +21,7 @@ Tasks are configured in the `tasks.js` file. This file is a JavaScript module wh
 
 | property | type | description | required |
 |---|---|---|---|
-| `name`| `string` | A unique identifier for the task. Used for querying task completeness. | yes, unique |
+| `name`| `string` | A unique identifier for the task. Used for querying task completeness. | no |
```
garethbowen, line 24: Our [task schema validation](https://github.com/medic/medic-conf/blob/d52eb7be4c5c71e600235e41705bff8f6bfa8c3a/src/lib/validate-declarative-schema.js#L97) lists name as a required field so medic-conf will fail if this isn't specified.

garethbowen (changes requested):
Nice one! One comment inline about inconsistency between the schema validation and this change.

content/en/apps/reference/tasks.md
abbyad: Ah, seems as though that changed with medic-conf/pull/275, likely so that tasks can be tracked from a data analysis perspective:

> This change adds two additional constraints to the declarative configuration schema:
> 
> - Task.name is now a required unique attribute
> - Task.event[n].id is now a required unique attribute when there is more than one event

I will clarify in this PR to match what is existing. We can revisit the value of these fields for the next set of improvements to the declarative config.

abbyad, line 24:
```suggestion
| `name`| `string` | A unique identifier for the task. Used for querying task completeness. | yes, unique |
```

content/en/apps/reference/tasks.md
```diff
@@ -104,6 +118,7 @@ module.exports = [
       }
     }],
     events: [ {
+      id: 'postnatal-visit',
```
abbyad, line 121: This isn't actually needed given that there is only one event.

garethbowen (approved):
Awesome.

Issue #418, opened 2021-01-22
Come with a way for easy to edit/author tables for use in MNCH docs


mrjones-plip:
For the [MNCH ticket](https://github.com/medic/cht-core/issues/6727), we're going to be adding some tables for when to give supplements/shots. They're currently in gdocs like this:

![ad hoc](https://user-images.githubusercontent.com/8253488/105558863-86cf2580-5cc4-11eb-87d4-4720cd7054ae.png)

PR #419, opened 2021-01-22
Add docs for tabular schedule markdown, CSS and shortcode for same

mrjones-plip:
Original [examples off different options here](https://user-images.githubusercontent.com/8253488/105559239-79666b00-5cc5-11eb-9a86-dfbc236c4b76.png), but we're going for circle emoji in markdown table with CSS sauce.

See #418


mrjones-plip:
FYI @michaelkohn

MaxDiz (approved):
The tabular schedule looks great, nice work!

content/en/contribute/docs/style-guide.md
```diff
@@ -192,6 +192,29 @@ Documentation pages should be written in [Markdown notation](https://www.markdow
...
+
+{{% alert title="Note" %}}
+The `div` HTML snippet must come before markdown table in order for it to affect the table styling
+{{% /alert %}}
+
+A sample row in markdown:
+
+```markdown
+| Deworming  | |   | 🟢 | 🟢 | 🟢 | 🟢 |  | 🟢 | 🟢 | 🟢 |
+```
```
abbyad, line 207: It would seem helpful for the example to include the html portion as well. Is it worth just using the code for the example?

```suggestion
```markdown
<div class="schedule-table"></div>

|| 6 mo | 12 mo | 18 mo | 2 yr | 2.5 yr | 3 yr | 3.5 yr | 4 yr | 4.5 yr | 5 yr |
|------------|--|--|--|--|--|--|--|--|--|--|
| Deworming  | |   | 🟢 | 🟢 | 🟢 | 🟢 |  | 🟢 | 🟢 | 🟢 |
| Vitamin A  | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 |
```
```

File: content/en/contribute/docs/style-guide.md [hunk 2]
```diff
@@ -192,6 +192,29 @@ Documentation pages should be written in [Markdown notation](https://www.markdow
 | **bold** | `**bold**` | `<b>bold</b>` or `<strong>bold</strong>` |
 | table | `|...|` using [markdown tables](https://www.markdownguide.org/extended-syntax/#tables) | `<table><tr><td>...</td></tr></table>` | 
 
+### Tabular schedules
+
+Lists of supplement or vaccination schedules should be in markdown and use the green circle emoji (🟢). Leave cells empty when no action is needed. Add the `<div class="schedule-table"></div>` HTML before the table to make columns narrower.
```
abbyad, line 197: This could be presented as more generic, and use vaccinations as an example.

```suggestion
Displaying the occurence of events over time in a workflow is often done using a table. To keep these consistent we recommend using the tabular schedule format. 

For example, here is a sample vaccination schedules:
<div class="schedule-table"></div>
|| 6 mo | 12 mo | 18 mo | 2 yr | 2.5 yr | 3 yr | 3.5 yr | 4 yr | 4.5 yr | 5 yr |
|------------|--|--|--|--|--|--|--|--|--|--|
| Deworming  | |   | 🟢 | 🟢 | 🟢 | 🟢 |  | 🟢 | 🟢 | 🟢 |
| Vitamin A  | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 |

To acheive this use a markdown table with the green circle emoji (🟢) to mark events, leaving cells empty when no action is needed. The HTML `<div class="schedule-table"></div>` is used before the markdown table so that built-in styling can be applied to the table. Here is the code for the above example:

```markdown
<div class="schedule-table"></div>
|| 6 mo | 12 mo | 18 mo | 2 yr | 2.5 yr | 3 yr | 3.5 yr | 4 yr | 4.5 yr | 5 yr |
|------------|--|--|--|--|--|--|--|--|--|--|
| Deworming  | |   | 🟢 | 🟢 | 🟢 | 🟢 |  | 🟢 | 🟢 | 🟢 |
| Vitamin A  | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 |
``` 
```

**Review by abbyad — CHANGES_REQUESTED** [review]
This is awesome to have! 

I made a _small_ suggestion, which looks like a big change, but is really just about presenting the generic use, then an example with its code.

File: content/en/contribute/docs/style-guide.md (hunk 2 above)
Comment (abbyad) [reply content/en/contribute/docs/style-guide.md:197]: Note that the closing markdown ` ``` ` is not properly shown in the suggested change.

**Comment by abbyad** [comment]
After some reflection, I think the best approach here is to create a new shortcode. The shortcode could in turn be implemented with the CSS method here.

We already have `workflow-table` shortcode. A `schedule-table` short code could wrap the markdown table more neatly than requiring a doc writer to add HTML. For example:

```markdown
{{% schedule-table %}}
|| 6 mo | 12 mo | 18 mo | 2 yr | 2.5 yr | 3 yr | 3.5 yr | 4 yr | 4.5 yr | 5 yr |
|------------|--|--|--|--|--|--|--|--|--|--|
| Deworming  | |   | 🟢 | 🟢 | 🟢 | 🟢 |  | 🟢 | 🟢 | 🟢 |
| Vitamin A  | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 | 🟢 |
{{% /schedule-table %}}
```

mrjones-plip:
@abbyad - cool, i'm down with a shortcode! Lemme give it a go and then get back to you with either questions or an updated PR

mrjones-plip:
@abbyad  - I resolved the now moot suggestions as I've now incorporated them in my latest commit, which includes the new pancy-fants shortcode (see 37c552e)

mrjones-plip:
Hmmm...we don't need a closing short code? I had it in there in 37c552e but then removed it in bd8fdf2 - happy to go back to having if you think that's best! Was going for "less is more" approach.

abbyad:
Nice work with your (first?) shortcode!

As for the closing code, it is a (literal) coincidence that it works -- the shortcode outputs the `div`, has no inner content, and then the next bit of content is the table from the original markdown file. Having the two pieces next to each other is how this CSS trick works, so it is fortunate. Although it works now, it could break pages in the future if you decide to use a different implementation of the shortcode. I think the concept of wrapping the entire table is clearer, and probably best to document it as such. We can still be happy for the times it works when folks don't put a closing tag!

assets/scss/_variables_project.scss
```diff
@@ -81,14 +81,19 @@ h2, h3 {
...
+      color: $primary;
     }
     &__section.selected {
       color: $secondary;
     }
   }
 }
 
+.td-main .td-content {
+  .schedule-table + table td  {
```
abbyad, line 93: Is the `td-main` and `td-content` required? If we think the specificity is valuable enough to include, should we include `workflow` from below in there as well?

In terms of naming, given that we have "workflow" which is also a table with a shortcode, should we also align the naming (eg just `schedule` to match the `workflow` example)? Seems worthwhile to consider now before we start using it.

abbyad (changes requested):
Looks great with the shortcode. Only minor comments inline.

assets/scss/_variables_project.scss
mrjones-plip: cool cool! I've decreased specificity and renamed it, both to match workflow. We can always increase specificity later.

mrjones-plip:
Thanks for the feedback! I've reverted to using closing tag and updated docs and .html files accordingly!

abbyad (approved):
Nice work!!

PR #421, opened 2021-01-26
Remove task references from targets schema

smbuthia:
Minor update to remove some task references in the schema.


garethbowen (approved):
Nice one, thanks!

Issue #429, opened 2021-02-10
Improve 'Generic'


abbyad:
Page: /apps/features/integrations/generic/
Version: 2020-12-16 14:39:28  0100  0100
---

The term "generic integration" makes sense from the CHT side, but from an implementer or app developer perspective it would be more clearly understood as "custom" integration. After a quick internal poll there was wide representation of roles in the 6 assents, and no dissents.

michaelkohn:
Should this be closed?

PR #430, opened 2021-02-10
Rename generic integrations to be custom integrations

abbyad:
As per #429

This may be our first use of aliases... @mrjones-plip is this the best way to do redirects when changing the file name? 

Also, should the best practice be to change all references too (one of one done in this PR)?

### DIFF
--- a/content/en/apps/features/integrations/custom.md
+++ b/content/en/apps/features/integrations/custom.md
@@ -1,16 +1,18 @@
 ---
-title: "Generic"
+title: "Custom"
 weight: 1
 description: >
    Integrate with any system using RESTful APIs
-keywords: generic
+keywords: generic, custom, integrations
+aliases:
+    -    /apps/features/integrations/generic/
 relatedContent: >
   apps/features/integrations/dhis2
   apps/features/integrations/rapidpro
   apps/features/integrations/openmrs
 ---
 
-The CHT Core Framework includes functionality that allows sharing data with any API-based system. App Builders have configured CHT integrations with OpenMRS, KenyaEMR, Bahmni, DHIS2, RapidPro, Apache NiFi, OpenHIM, custom EMRs, and several other systems.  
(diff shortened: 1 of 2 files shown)


mrjones-plip:
@abbyad - It's neat to see us have such mature and extensive content we have this problem ; ) 

How I read the [hugo docs on aliases](https://gohugo.io/content-management/urls/#aliases), we're using it correctly in this PR. Nice work! 

I'm a bit bummed it serves a `200` and then does the redirect in HTML with a `meta` tag (eg `<meta http-equiv="refresh" content="0; url=http://localhost:1313/apps/features/integrations/custom/" />`), but :shrug: , search engines and browsers alike are smart enough to parse this and follow along and it's a nominal payload so only _slightly_ slower than a proper 301.

As far a best practice to change all links or not, hugo forces our hand on this matter. When you change a document name (and thus a URL) you MUST update all  links or else the site won't compile.  This is what I see when I added back in an intentionally outdated link to `[Generic Integrations]({{< ref "apps/features/integrations/generic" >}}) ` in this branch:

```
Change detected, rebuilding site.
2021-02-11 20:53:58.151 -0800
Source changed "/home/mrjones/Documents/MedicMobile/cht-docs/content/en/apps/features/integrations/custom.md": WRITE
ERROR 2021/02/11 20:53:58 [en] REF_NOT_FOUND: Ref "apps/features/integrations/generic": "/home/mrjones/Documents/MedicMobile/cht-docs/content/en/apps/features/integrations/custom.md:19:25": page not found
Total in 152 ms
ERROR 2021/02/11 20:53:58 Rebuild failed:
ERROR 2021/02/11 20:53:58 Logged 1 error(s)
```

mrjones-plip (approved):
lgmt!

abbyad:
Thanks for reviewing and the thorough testing! It's good to know about the references all needing to change. It's probably easier to maintain it like that in long run, so it's good that they force our hand!

PR #432, opened 2021-02-11
fix group_assessment variable access

### LINKED ISSUE #433: Task sample code results in a task that never appears
Page: /apps/tutorials/tasks/
Version: 2021-01-22 07:00:56 -0800 -0800
---
`parseInt` will always return NaN
```
appliesIf: function(contact, report) {
      return report && report.fields.group_assessment && parseInt(report.fields.group_assessment) > 3;
    }
```
while what we're after is the group_assessment's field `cough_duration`

freddieptf:
`parseInt(report.fields.group_assessment) > 3;` returns NaN and the task never appears. What we're actually after is `report.fields.group_assessment.cough_duration`. Fixes #433


smbuthia (approved):
Good catch!

PR #442, opened 2021-03-03
change Medic 3.5 to cht-core 3.5

mrjones-plip:
Per @isaacholeman's [request in slack](https://medic.slack.com/archives/CUXB9SVT5/p1614743413001400).

### DIFF
--- a/content/en/apps/reference/app-settings/outbound.md
+++ b/content/en/apps/reference/app-settings/outbound.md
@@ -7,7 +7,7 @@ description: >
 relevantLinks: >
 ---
 
-*Outbound is only available in cht-core 3.5.0 and above*
+*Outbound is only available in CHT Core 3.5.0 and above*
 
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/app-settings/outbound.md
```diff
@@ -7,7 +7,7 @@ description: >
 relevantLinks: >
 ---
 
-*Outbound is only available in Medic 3.5.0 and above*
+*Outbound is only available in cht-core 3.5.0 and above*
```
abbyad, line 10: In most other places we mention versions we don't specify cht-core (or medic), but rather just say "available in X". The extra clarity is helpful though, and for consistency should use "CHT Core" rather than the "cht-core".

```suggestion
*Outbound is only available in CHT Core 3.5.0 and above*
```

mrjones-plip: Awesome - thanks for the style guide help there @abbyad  - much obliged.

garethbowen (approved):
Nice

PR #446, opened 2021-03-05
Fix contact task action schema definition and example code typo

mrsarm:
Small corrections:

- Fix schema definition of `actions[n].form`, required only if `actions[n].type` is `'report'`
- Remove wrong `;` in task example

Ticket: https://github.com/medic/medic-conf/issues/358


dianabarsan (approved):
Nice :) Thanks for updating!

garethbowen (approved):
Nice one.

PR #448, opened 2021-03-10
Remove unused FAQ section

mrjones-plip:
Ever since launch of the docs site, we've had [this FAQ section](docs.communityhealthtoolkit.org/core/faq/) that is empty.  I think since it's because it's still empty and not being used, we should remove it!


garethbowen (approved):
Agreed. Thanks!

Issue #450, opened 2021-03-15
link for update_clinics on `facility_reference` goes to muting


newtewt:
The link for `update_clinics transtions` on the table row for `facility_reference` goes to muting config. There is no real details on the `update_clinics` other than what is in the available transitions.

garethbowen:
Merged.

PR #451, opened 2021-03-15
changed the reference that the link is potining to.

newtewt:
https://github.com/medic/cht-docs/issues/450


garethbowen (approved):
Thanks!

PR #466, opened 2021-04-08
Adding documentation for default task resolve checking method

### LINKED ISSUE #465: Documentation for default task resolve checking method
Edit the documentation for medic/medic-conf#107

binokaryg:
Marked `resolvedIf` as optional (conditionally) and added a section on using the default resolve check method
#medic/medic-conf#107

Issue #465

### DIFF
--- a/content/en/apps/reference/tasks.md
+++ b/content/en/apps/reference/tasks.md
@@ -229,7 +229,7 @@ If the `resolvedIf` is undefined in an action of type `report`, then `resolvedIf
 The `defaultResolvedIf` method returns `true` if it finds any report assigned to the contact that matches the `form` defined in the action of type `report`. Only the reports submitted during a specific time period are considered:
 - For a contact-based task, the period is the same as the task window period i.e. when the task is visible.
 - For a report based task, the period is determined between `start` and `end` as:
-  - `start`: the latest date between start of the task's window and one millisecond after the report's reported date
+  - `start`: the latest date between start of the task window and one millisecond after the report's reported date
   - `end`: end of the task window
 
 You can also use `this.definition.defaultResolvedIf` inside the `resolvedIf` definition and optionally add more conditions:
--- a/content/en/apps/reference/tasks.md
+++ b/content/en/apps/reference/tasks.md
@@ -223,54 +223,22 @@ module.exports = {
 };
 ```
 
-## Default resolvedIf method
-If you skip defining `resolvedIf` section in the task definition, it will default to defaultResolvedIf method which is equivalent to this:
+### Default resolvedIf method
+If you skip defining `resolvedIf` section in the task definition, it will default to the `defaultResolvedIf` method. You can only skip the section if you have at least one action with `report` type.
+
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/tasks.md
```diff
@@ -223,6 +223,55 @@ module.exports = {
 };
 ```
 
+## Default resolvedIf method
+If you skip defining `resolvedIf` section in the task definition, it will default to defaultResolvedIf method which is equivalent to this:
```
latin-panda, line 227: Maybe explaining the why, that if actions type is `report` and `resolvedIf`isn't defined then `defaultResolvedIf` will be executed in its place. 

Also that `defaultResolvedIf` is available in the task scope to be used and how to use it (code sample).

You can also explain with few words what `defaultResolvedIf` validates (conditions|rules) so it's clear the outcome, no need of pasting it's code since it can change slightly and be out of date easy, better that if someone is interested of looking at the implementation then they can go see the open source. 
Specially because `defaultResolvedIf` won't be modified from outside but just used.

content/en/apps/reference/tasks.md
```diff
@@ -223,6 +223,55 @@ module.exports = {
...
+You can also use `this.defaultResolvedIf` directly:
+
+```js
+resolvedIf: this.defaultResolvedIf,//Same as skipping resolvedIf
+```
+
+It is also possible to specify a resolving form type:
+
+```js
+resolvedIf: function (c, r, event, dueDate) {
```
latin-panda, line 268: Please replace 'c' for 'contact' and 'r' for 'report', trying to fix the one letter names for more descriptive names. Thanks!

Remember to add semi-colon at the end of each JS line. 

And use `cons` and `let` instead of `var`, basically leaving the best best practices in the code sample for others to replicate 💪

latin-panda (changes requested):
Nice! I've left few comments inline :) Thanks!

content/en/apps/reference/tasks.md
```diff
@@ -223,6 +223,23 @@ module.exports = {
 };
 ```
 
+### Default resolvedIf method
+If you skip defining `resolvedIf` section in the task definition, it will default to the `defaultResolvedIf` method. You can only skip the section if you have at least one action with `report` type.
```
latin-panda, line 227:
```suggestion
If the `resolvedIf` is undefined in an action of type `report`, then `resolvedIf` is going to default to `defaultResolvedIf` function.
```

content/en/apps/reference/tasks.md
```diff
@@ -223,6 +223,23 @@ module.exports = {
...
+
+The `defaultResolvedIf` method looks for reports assigned to the contact, and returns `true` if it finds any report that matches the `form` defined in your first action with `report` type. Only the reports submitted during a specific time period are considered:
```
latin-panda, line 229:
```suggestion
The `defaultResolvedIf` method returns `true` if it finds any report assigned to the contact that matches the `form` defined in the action of type `report`. Only the reports submitted during a specific time period are considered:
```

latin-panda (approved):
Nice! @binokaryg I've added some suggestions in wording. See comments above, let me know your thoughts :) 

Note: When ticket was passed AT, then this PR can be merged to main branch

Issue #468, opened 2021-04-12
Move initial design tutorials from Google docs


smbuthia:
@leah-ngaari wrote some [design tutorials on Google docs](https://docs.google.com/document/d/1vWg5aF8qg_iGFWVHtCw8wsi1kEK0luBEaV1Ad-oC4rg/edit#). Move these tutorials to CHT docs.
cc @katanu

PR #469, opened 2021-04-15
Add initial design quick guides

smbuthia:
#468


n-orlowski:
@smbuthia looking at the persona guides I think we should align to what [we currently have in the docs](https://docs.communityhealthtoolkit.org/design/personas/chw-janet/), ex. About, Values, Responsibilities, Needs, Motivations, Day in the Life (either update the template or fill in our current personas with this information).

As for the order of the pages, I think it makes sense that we order them in the chronological process. If it's currently the case that's fine but I imagine it would go something like:

1. Designing Interviews
2. Creating Personas
3. Empathy Map
4. Problem Statement
5. Mapping Hierarchy

content/en/design/guides/empathy-map.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Brief Overview Of Key Concepts
+
+Insight refers to a remarkable realization that can help you to solve the current design challenge you are facing.
+
+## Steps
+
+### 1. Fill Out the Empathy Map
+
+Lay the four quadrants out on a table, draw them on paper or on a whiteboard.
```
n-orlowski, line 33: Include a visual template here?

content/en/design/guides/mapping-hierarchy.md
```diff
@@ -0,0 +1,56 @@
...
+
+User persona is a generalized character that embodies a particular type of user.
+
+User roles are the activities that the user personas are expected to carry out.
+
+Hierarchy represents the reporting structures
+
+## Prerequisites/Required Resources
+
+User personas profiles
```
n-orlowski, line 29: Link to persona page

smbuthia: Done. I also added a `see also` that links to the other page on Personas.

smbuthia:
Thanks for the notes @n-orlowski. The order was the same except for 3 and 5 which were interchanged. I've updated that.

I'm not too sure about the layout and content of the personas page since I was mostly following what Leah had drafted. @katanu what template or format do service designers currently use?

smbuthia:
@n-orlowski, I've created a [separate issue](https://github.com/medic/cht-docs/issues/483) to address the persona template issue so that this PR is unblocked.

n-orlowski (approved):

Issue #475, opened 2021-04-19
Broken links


abbyad:
From builds on #330 we see these errors:
```
http://localhost:1313/apps/reference/contact-page/
	404	https://github.com/medic/medic/tree/master/webapp/src/js/filters
http://localhost:1313/apps/concepts/interoperability/
	404	https://www.who.int/classifications/icd/factsheet/en/
http://localhost:1313/core/overview/translations/
	404	https://github.com/medic/cht-core/blob/master/webapp/src/js/app.js#L39
```

PR #476, opened 2021-04-19
Fixes broken links

abbyad:
For https://github.com/medic/cht-docs/issues/475


content/en/apps/reference/contact-page.md
```diff
@@ -34,7 +34,7 @@ Each field that can be shown on a contact's profile is defined as an object in t
...
 | `value` | `string` | The value shown for the field. | yes |
-| `filter` | `string` | The display filter to apply to the value, eg: `{ value: '2005-10-09', filter: 'age' }` will render as "11 years". Common filters are: `age`, `phone`, `weeksPregnant`, `relativeDate`, `relativeDay`, `fullDate`, `simpleDate`, `simpleDateTime`, `lineage`, `resourceIcon`, `translate`. For the complete list of filters, and more details on what each does, check out the code in [`medic/webapp/src/js/filters` dir](https://github.com/medic/medic/tree/master/webapp/src/js/filters). | no |
+| `filter` | `string` | The display filter to apply to the value, eg: `{ value: '2005-10-09', filter: 'age' }` will render as "11 years". Common filters are: `age`, `phone`, `weeksPregnant`, `relativeDate`, `relativeDay`, `fullDate`, `simpleDate`, `simpleDateTime`, `lineage`, `resourceIcon`, `translate`. For the complete list of filters, and more details on what each does, check out the code in [`medic/webapp/src/js/filters` dir](https://github.com/medic/cht-core/tree/3.10.x/webapp/src/js/filters). | no |
```
craig-landry, line 37: Was the `3.10.x` intentional here? Presumably we'd need to update this link over time as new versions are released.

content/en/core/overview/translations.md
```diff
@@ -22,7 +22,7 @@ New languages must be added and configured in several places:
 - *In GitHub*
   - Create a new [ddocs/medic/_attachments/translations/messages-XX.properties](https://github.com/medic/cht-core/tree/master/ddocs/medic/_attachments/translations) file, replacing XX with the 2 or 3 letter language code.
   - Add the language to the [`LOCAL_NAME_MAP` in api](https://github.com/medic/cht-core/blob/master/api/src/translations.js#L8). Use the language code for the key, and the local name followed by the English name for the language in brackets, eg: "fr: 'Français (French)'".
-  - Import the moment language pack in the [root app.js file](https://github.com/medic/cht-core/blob/master/webapp/src/js/app.js#L39). If moment doesn't provide the required language pack you may need to contribute it upstream to the moment library.
+  - Import the moment language pack in the [main.ts file](https://github.com/medic/cht-core/blob/3.11.x/webapp/src/ts/main.ts#L23). If moment doesn't provide the required language pack you may need to contribute it upstream to the moment library.
```
craig-landry, line 25: ...same question here about referencing tagged versions or the master branch, where `master` is used on line 24 in this file.

craig-landry (commented):
Looks mostly fine, but I did notice a mix of links to the master branch or tagged/release code. I can think of reasons to do either way and don't have a huge preference one way or another, but consistency seems worthwhile, whatever route is taken.

content/en/apps/reference/contact-page.md
abbyad: Sorry, the note about this is hidden in the commit message: "Temporary fix until filters documented fully in docs."

The code got completely refactored in 3.11, so we don't have somewhere easy to point to. We should properly document the filters individually in these docs, rather than link to code anyhow, but in the meantime linking to 3.10.x would be fine.

content/en/core/overview/translations.md
abbyad: This is another situation where the location changed. The other one should be replaced by documentation, but in this case the reference to code is valid. What's the best way to keep this from going stale? Keep it versioned and have a todo to update docs in release steps, or link to master and put a comment in that file to update the docs if the location of moment changes?

abbyad:
Ya, I'd generally like to get out of the habit of linking to code. Linking to master seems more frail, so seems as though we should link to the latest version branch where absolutely needed, and in the release steps make sure that they are revisited.

content/en/apps/reference/contact-page.md
craig-landry: Gotcha.

For the refactor, these are now "pipes" and the code has been moved/refactored to the `pipes` directory for 3.11+

```suggestion
| `filter` | `string` | The display filter to apply to the value, eg: `{ value: '2005-10-09', filter: 'age' }` will render as "11 years". Common filters are: `age`, `phone`, `weeksPregnant`, `relativeDate`, `relativeDay`, `fullDate`, `simpleDate`, `simpleDateTime`, `lineage`, `resourceIcon`, `translate`. For the complete list of filters, and more details on what each does, check out the code in [`webapp/src/ts/pipes` dir](https://github.com/medic/cht-core/tree/3.11.x/webapp/src/ts/pipes). | no |
```

craig-landry:
I agree with all your points. For now, this all works to keep it moving and make it correct. I added one update with the new location for the filters. Approving now so as not to block.

craig-landry (approved):

PR #485, opened 2021-05-02
7068 Remove documentation about setting the user's language via the API

jkuester:
medic/cht-core#7068

The associated code changes are included on: https://github.com/medic/cht-core/pull/7081

### DIFF
--- a/content/en/apps/reference/api.md
+++ b/content/en/apps/reference/api.md
@@ -727,7 +727,6 @@ will be undefined.
 | token_login | no | Boolean | A boolean representing whether or not the Login by SMS should be enabled for this user. | 3.10.0 | 
 | fullname | no | String | Full name  | 
 | email | no | String | Email address  |
-| language | no | String | Language preference. e.g. "sw" for Swahili |
 | known | no | Boolean | Boolean to define if the user has logged in before. Used mainly to determine whether or not to start a tour on first login. |
 
 ##### Login by SMS
@@ -788,10 +787,7 @@ Content-Type: application/json; charset=utf-8
     "id": "org.couchdb.user:admin",
     "rev": "10-6486428924d11781c107ea74de6b63b6",
     "type": "admin",
-    "username": "admin",
-    "language": {
-      "code": "en"
-    }
+    "username": "admin"
   },
   {
     "id": "org.couchdb.user:demo",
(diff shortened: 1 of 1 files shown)


dianabarsan (approved):
This is looking good.

jkuester:
@dianabarsan, now that the [code has been merged](https://github.com/medic/cht-core/pull/7081), should I be good to merge this PR too?  (I am not sure if there is any special AT that needs to be done for these doc changes)

dianabarsan:
@jkuester yes, merge away. The CI ... wait ... CI doesn't work for external contributors on cht-docs? 
Normally, the CI would be sufficient to be sure nothing is broken. Since the CI didn't start in this case, please run the "hugo" command locally to check that everything is in order before merging. (https://github.com/medic/cht-docs/blob/master/.github/workflows/ci.yml#L40)

cc @mrjones-plip

dianabarsan:
Oh, now I get the question: it was the "Blocked: waiting for AT" label that was confusing? 
Some context for that: https://medic.slack.com/archives/C024KTGRW/p1622557871066100

jkuester:
For the record, I was able to successfully run the `hugo` command locally.

mrjones-plip:
@dianabarsan  and @jkuester  - PR is merged already, but do we need a `cht-docs` ticket to research if GH CI is working correctly for internal vs external contributions? I can imagine there might be some security measures in place so that [Mallory](https://en.wikipedia.org/wiki/Alice_and_Bob#Cast_of_characters
) can't run arbitrary code in a privileged CI environment, but that would more apply to `cht-core` than lil' ol' `cht-docs`.

Either way, lemme know!

dianabarsan:
Thanks for the reply, @mrjones-plip . We do need a ticket for external contributors. 
In CHT-Core, what I've observed is that a user with write permissions has to click a button to initiate CI for an external contribution PR, so it becomes part of the code review process and is safe, as long as the reviewer makes sure no arbitrary dangerous code exists in the PR. I don't know if this is a setting in GitHub or just how things work by default.

dianabarsan:
added issue: https://github.com/medic/cht-docs/issues/525

mrjones-plip:
Awesome, thanks so much for filing that ticket @dianabarsan !  I'll follow up there.

PR #486, opened 2021-05-10
Document dropping of Android 4.4 support

garethbowen:


dianabarsan (approved):
Cool! 
So we won't even support Node 14, which is to be maintained until 2023?

garethbowen:
@dianabarsan It's a bit of a moot point because once the release artefact is the Docker image then we'll be able to dictate what Node version is required in that container so it'll be an internal dependency and I expect we'll just always go to the latest LTS. I thought about changing it to "N/A" or dropping the column but it's useful for historical versions.

PR #488, opened 2021-05-10
Revert "Add initial design quick guides"

smbuthia:
Reverts medic/cht-docs#469


mrjones-plip:
@smbuthia - this PR isn't needed, right?  Since it's failing for [everyone starting ~12 hours ago](https://github.com/medic/cht-docs/actions) (specifically [here](https://github.com/medic/cht-docs/actions/runs/826888660) and [here](https://github.com/medic/cht-docs/actions/runs/826888915) and [here](https://github.com/medic/cht-docs/actions/runs/826889101)), we just need to add a fix and then we can merge any new content that was failing to merge prior to the fix. 

Your new content didn't break anything ; )

If you agree, we can just close this PR and not merge it.

smbuthia:
Makes sense to me. I'll go ahead and close it.

Issue #496, opened 2021-05-14
SMS object keys are required to be caps


newtewt:
Our docs don't mention that the keys need to be capitalized but they are.

Issue #495, opened 2021-05-14
Forms docs shows as array but it is an object


newtewt:
Forms docs shows as an array but is an object of objects.

tmatheas:
@newtewt If this issue is fixed, please close the issue.

PR #497, opened 2021-05-14
change to object and added note about capitalization

newtewt:
#496 
#495

### DIFF
--- a/content/en/apps/reference/app-settings/forms.md
+++ b/content/en/apps/reference/app-settings/forms.md
@@ -10,7 +10,7 @@ relevantLinks: >
 keywords: workflows interop json-forms
 ---
 
-JSON forms are defined in either the `base_settings.json` or the `app_settings/forms.json` file and compiled in to the `app_settings.json` file with the `compile-app-settings` action in the `medic-conf` tool. JSON Forms are used for parsing reports from formatted SMS, SIM applications, and Medic Collect. JSON form definitions are also used for interoperability with third-party systems. Each form is defined as an JSON form object according to the following schema. The key for each object is required to be capitalized. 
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/app-settings/forms.md
```diff
@@ -10,9 +10,9 @@ relevantLinks: >
 keywords: workflows interop json-forms
 ---
 
-JSON forms are defined in either the `base_settings.json` or the `app_settings/forms.json` file and compiled in to the `app_settings.json` file with the `compile-app-settings` action in the `medic-conf` tool. JSON Forms are used for parsing reports from formatted SMS, SIM applications, and Medic Collect. JSON form definitions are also used for interoperability with third-party systems. Each form is defined as an JSON form object according to the following schema. 
+JSON forms are defined in either the `base_settings.json` or the `app_settings/forms.json` file and compiled in to the `app_settings.json` file with the `compile-app-settings` action in the `medic-conf` tool. JSON Forms are used for parsing reports from formatted SMS, SIM applications, and Medic Collect. JSON form definitions are also used for interoperability with third-party systems. Each form is defined as an JSON form object according to the following schema. The key for each object is required to be capitalized.
```
abbyad, line 13: Would be good to clarify that it is the entire string that needs to be uppercase, and I also think uppercase will reasonate more than capitalized to say that. 

```suggestion
JSON forms are defined in either the `base_settings.json` or the `app_settings/forms.json` file and compiled in to the `app_settings.json` file with the `compile-app-settings` action in the `medic-conf` tool. JSON Forms are used for parsing reports from formatted SMS, SIM applications, and Medic Collect. JSON form definitions are also used for interoperability with third-party systems. Each form is defined as an JSON form object according to the following schema. The key for each object must be unique and all characters must be uppercase. 
```

abbyad (changes requested):
Good to have these changes included!

Suggestion inline, but it may be a matter of preference. If you feel strongly I am happy to approve as is.

newtewt:
@abbyad  I think your suggestion makes sense. I committed and just need an approval.

abbyad (approved):
Looks great, thanks for making the corrections that will surely save someone headaches and time down the line!

PR #500, opened 2021-05-17
Update transitions.md

njogz:
An update to the `update_clinics` documentation necessitated by this [pull request](https://github.com/medic/cht-core/pull/7031).


content/en/apps/reference/app-settings/transitions.md
```diff
@@ -34,7 +34,7 @@ The following transitions are available and executed in order.
...
+| update_clinics | Adds a contact's info to a new data record. This is used to attribute an incoming SMS message or report to the appropriate contact. The `rc_code` value on the contact is used to match to the value of the form field set as the `facility_reference` in the [JSON form definition]({{< ref "apps/reference/app-settings/forms#app_settingsjson-forms" >}}). This matching is useful when reports are sent on behalf of a facility by unknown or various phone numbers. If `facility_reference` is not set for a form, the contact match is attempted using the sender's phone number. If a form is not public and a match is not found then the `sys.facility_not_found` error is raised. Ammessage can be configured to be sent out whenever this error is raised. |
```
dianabarsan, line 37: I think we should add more details about how the configuration should look like, under which key it should be found, what happens if no configuration is provided and so on. Please have a look at the other transitions that have configuration and detailed descriptions, for example: https://docs.communityhealthtoolkit.org/apps/reference/app-settings/transitions/#accept-patient-reports

dianabarsan (changes requested):
I think we should add a new section for `update_clinics` to detail how the configuration looks like and how it behaves. And also outline that the behavior has changed in this version.

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -34,7 +34,7 @@ The following transitions are available and executed in order.
...
+| update_clinics | Adds a contact's info to a new data record. This is used to attribute an incoming SMS message or report to the appropriate contact. The `rc_code` value on the contact is used to match to the value of the form field set as the `facility_reference` in the [JSON form definition]({{< ref "apps/reference/app-settings/forms#app_settingsjson-forms" >}}). This matching is useful when reports are sent on behalf of a facility by unknown or various phone numbers. If `facility_reference` is not set for a form, the contact match is attempted using the sender's phone number. If a form is not public and a match is not found then the `sys.facility_not_found` error is raised. A message can be configured to be sent out whenever this error is raised. |
```
dianabarsan, line 37: Please update the first column here to have a link to the section.

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -575,3 +575,48 @@ Supported `events_types` are:
...
+
+### update_clinics
+Adds a contact’s info to a data record so as to attribute an incoming SMS message or report to the appropriate contact. 
+
+#### Configuration
+As of version 3.12 you can add configuration to send a message whenever a contact match fails while running this transition. Configuration is stored in the `update_clinics` field of app_settings.json as a list of objects connecting forms to messages. Every object should have this structure:
+
+| Property | Description |
+|---|---|
+| `form` | Form code. **Required** |
```
dianabarsan, line 587: What does "required" mean here? Does the transition fail to load if this field is not set?

content/en/apps/reference/app-settings/transitions.md
dianabarsan: Also, please change the phrasing of: 
```
A message can be configured to be sent out whenever this error is raised
```
To: 
```
A message will be sent out whenever this error is raised. 
```

dianabarsan (changes requested):
Nice, just a couple of small details.

content/en/apps/reference/app-settings/transitions.md
njogz: Removed it to avoid confusion.

dianabarsan (approved):
Nice one!

Issue #494, opened 2021-05-13
Improve 'Africa’s Talking SMS Aggregator'


shreyakb:
Page: /apps/guides/messaging/africas-talking/
Version: 2021-04-09 06:59:14  0300  0300
---

Flagging that we need to update Medic branding in the image (remove "Mobile"). Thanks!

![image](https://user-images.githubusercontent.com/7915295/118085876-bb383000-b3e0-11eb-9ee9-1f0ddd649ee9.png)

garethbowen:
@abbyad You were the last one to touch this image. Do you have the source?

abbyad:
Thanks for highlighting @shreyakb! 

I converted this page to markdown but wasn't involved in the creation of the original. I think I found the [original](https://docs.google.com/presentation/d/1vW3lVD5cfBJd-wn5KTah1zGJ1La9n96URsNjxLHjEjM) in drive though.

abbyad:
The gateway section needed a fair bit of reorganization now that we have multiple gateways. As part of that reorganization the image was not as relevant, escpecially given that it was not technically accurate as it was. If someone feels as though conceptual images would help please open a new issue.

PR #503, opened 2021-05-18
Reorganize gateway guides

abbyad:
Triggered by https://github.com/medic/cht-docs/issues/494, and then realized that much of this section predates having more than one gateway. I decide to tackle it together to try to sort out this section.

As it relates to the original issue, the diagram was removed since it wasn't adding much, and wasn't technically accurate. If we think a diagram explaining the flow is important we should put one at the top level.

Aliases were added to pages that moved to avoid broken/direct search links, and I suspect some internal references may need to be updated. Will move from Draft once those are resolved.


mrjones-plip (approved):
Lots of links changed for the refactor - GH CI + Hugo did the hard part for me when your build failed and then you fixed it ;)

Approving on the condition @shreyakb and company are OK not having [the image](https://github.com/medic/cht-docs/blob/3f2c758253d6c03e1361f6d78242ff812d5e78fe/content/en/apps/guides/messaging/africas-talking/sms-aggregator-workflow.png) on the top of the "[Africa’s Talking SMS Aggregator](https://docs.communityhealthtoolkit.org/apps/guides/messaging/africas-talking/)" page.   I'm not sure how much this diagram adds, so deferring to others!

abbyad:
Ya, I was happy that CI did the hard part!

That said, they don't automatically check aliases being added for changed filenames. Double checking using the [Files changed](https://github.com/medic/cht-docs/pull/503/files) and searching for `md →` helped for that.

PR #504, opened 2021-05-19
Adds documentation about client-side muting.

dianabarsan:
medic/cht-core#6737

### DIFF
--- a/content/en/apps/reference/app-settings/transitions.md
+++ b/content/en/apps/reference/app-settings/transitions.md
@@ -13,7 +13,7 @@ When sentinel detects a document has changed it runs transitions against the doc
 ## Configuration
 
 By default all transitions are disabled. They can be enabled by configuring the `transitions` property to have a key with the transitions name and a `truthy` value. 
-As of version 3.12.0, some transitions can partially run on the client, for offline users. To op out from client-side transitions, add a `"client: false"` property to the transition configuration.
+As of version 3.12.0 some transitions will partially run on the client for offline users. To opt out from client-side transitions, add a `"client: false"` property to the transition configuration.
 
 ```json
 {
--- a/content/en/apps/reference/app-settings/transitions.md
+++ b/content/en/apps/reference/app-settings/transitions.md
@@ -359,32 +359,32 @@ Implements muting/unmuting of persons and places. Supports multiple forms for ea
 As of 3.12.0, client-side muting only runs on new reports or new contacts, before they are saved in the local database:
 
 - updates the target contact and all its descendants<sup>[10]</sup>, setting the `muted` property equal to the device's current `date` in ISO format<sup>[8]</sup>. 
-- adds/updates the `muting_history`<sup>[11]</sup> property on every updated contact, to keep track of all the updates that have been processed on the client, as well as the last known server-side state of the contact and sets the `last_update` property to `client`
-- updates the report doc to add a `client_transitions` property to track which transitions have run client-side
+- adds/updates the `muting_history`<sup>[11]</sup> property on every updated contact, to keep track of all the updates that have been processed client-side, as well as the last known server-side state of the contact and sets the `last_update` property to `client_side`
+- updates the report doc to add a `client_side_transitions` property to track which transitions have run client-side
 
 Server-side:
 
 - updates the target contact and all its descendants<sup>[1]</sup>, setting the `muted` property equal to the current `date` in ISO format<sup>[2]</sup>. If the contact was already muted by a client, the `muted` date will be overwritten. The client-side `muting_history` will have a copy of the client-side muting date.
 - adds a `muting_history` entry to Sentinel `info` docs for every updated contact<sup>[7]</sup>
 - updates all connected registrations<sup>[3]</sup>, changing the state of all unsent<sup>[4]</sup> `scheduled_tasks` to `muted`
-- as of 3.12.0, updates the contact's client-side `muting_history` to set the `last_update` property to `server` and update the `server` section with the current date and muted state. 
-- as of 3.12.0, if the report was processed on the client, all "following" muting/unmuting events that have affected the same contacts will be replayed. This means the transition _could_ end up running multiple times over the same report<sup>[9]</sup>. 
+- as of 3.12.0, updates the contact's client-side `muting_history` to set the `last_update` property to `server_side` and update the `server_side` section with the current date and muted state. 
+- as of 3.12.0, if the report was processed client-side, all "following" muting/unmuting events that have affected the same contacts will be replayed. This means the transition _could_ end up running multiple times over the same report<sup>[9]</sup>. 
 
 ##### Unmuting action:
 
 As of 3.12.0, client-side unmuting only runs on new reports before they are saved in the local database:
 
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/app-settings/transitions.md
```diff
@@ -12,20 +12,22 @@ When sentinel detects a document has changed it runs transitions against the doc
 
 ## Configuration
 
-By default all transitions are disabled. They can be enabled by configuring the `transitions` property to have a key with the transitions name and a `truthy` value, eg:
+By default all transitions are disabled. They can be enabled by configuring the `transitions` property to have a key with the transitions name and a `truthy` value. 
+As of version 3.12.0, some transitions can partially run on the client, for offline users. To op out from client-side transitions, add a `"client: false"` property to the transition configuration.
```
garethbowen, line 16:
```suggestion
As of version 3.12.0 some transitions will partially run on the client for offline users. To opt out from client-side transitions, add a `"client: false"` property to the transition configuration.
```

garethbowen (approved):
Great work. One minor suggestion inline - approved.

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -352,24 +354,47 @@ No custom configuration for `generate_patient_id_on_people`.
...
+
+[1] Contacts that are already in the correct state are skipped. This applies to updates to the contact itself, updates to the Sentinel `muting_history` and to the connected registrations (registrations of a contact that is already in the correct state will not be updated).  
+[2] The date represents the moment Sentinel has processed the muting action  
+[3] target contact and descendants' registrations  
+[4] `scheduled_tasks` being in either `scheduled` or `pending` state  
+[5] Because the muted state is inherited, unmuting cascades upwards to the highest level muted ancestor. If none of the ancestors is muted, unmuting cascades downwards only.  
+[6] `scheduled_tasks` which are due today or in the future. All `scheduled_tasks` with a due date in the past will remain unchanged.  
+[8] The date represents the device's date when the report is processed.  
+[9] Replaying is required due to how PouchDB <-> CouchDB synchronization does not respect the order in which the documents have been created, to ensure that contacts end up in the correct muted state.    
+[10] The updated contacts are limited to the contacts available on the device.
```
abbyad, line 397: I hadn't noticed our use of footnotes before, and find it quite hard to find/use given that they aren't cross linked (eg it isn't clear where to go to see the footnotes, and 7 is separate). It may be outside of the scope of this PR to fix, but just highlighting as we are adding to them here.

dianabarsan: Yea, I think with plain markdown files it wasn't possible to superscript and also link, maybe we have some options now.

abbyad: I'll create a separate issue for the footnotes.

abbyad: Closing the loop: https://github.com/medic/cht-docs/issues/509

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -381,6 +406,23 @@ Entries in `muting_history` contain the following information:
 | date | Date in ISO Format |
 | report_id | An `_id` reference to the report that triggered the action |
 
+
+##### [11] Client-side Muting history as of 3.12.0
```
kennsippell, line 410: worth noting in contacts schema page?
https://docs.communityhealthtoolkit.org/core/overview/db-schema/#contacts-persons-and-places

dianabarsan: Unsure. This is an implementation detail, I wouldn't want it: a. edited or b. used for any other purpose by data-analysis and such.

kennsippell: it is in the contact doc, right? it is going to get used :( 
if that's not a good idea - maybe that's what you should say in the schema documentation?

dianabarsan: It is in the contact doc, correct. We need this metadata to insure some sort of consistency with out of order syncing and muting. The alternative would have been to have another doc type to sync to just contain this metadata.

dianabarsan: Server-side Muting history exists in infodocs, and can be used for dashboards.

kennsippell: > The alternative would have been to have another doc type to sync to just contain this metadata

Sounds not that nice.

I'm not really saying you shouldn't put the data here. But I'm saying that this data is very public and discoverable, so it is likely it will be used.

dianabarsan: I can't guarantee that this data structure won't change over time. It being an implementation artifact, I don't want to block future development from changing this structure to accommodate new functionalities or bug fixes because data analysis is already using it. 
How do you think it's best communicated that this data is not guaranteed to have this format or exist at all?

kennsippell: most people writing data analytic queries are doing it from xform data. with that data, all the internal calculations/intermediate steps/UI strings are even exposed publicly. it is very common for a query to depend on something fragile which either a) breaks the query, or b) makes it more difficult to edit the form later.

we would benefit from a convention of "private vs public data", but i believe we'd be creating it here. today, everything is fair game cc @derickl

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -381,6 +406,23 @@ Entries in `muting_history` contain the following information:
...
+Each time the client changes the `muted` state of a contact, an entry is added to a `muting_history` property on the contact's doc. The `last_update` entry is also changed to `client`.
+The `muting_history` property contains the following information:
+
+| Property | Values | Description | 
+| --- | ---- | ---- |
+| last_update | `server` or `client` | Updated every time a service updates the contact, with the corresponding value | 
+| server | Object |  |
+| server.muted | `true` or `false` | Last known server muting state |
+| server.date | Date in ISO format | Last known server muting/unmuting date | 
+| client | Array | Client-side muting/unmuting events list. <br> New events are pushed at the end of this list and it should never be re-ordered. <br>The list represents the "chronological" order in which the reports that triggered muting were created. |
```
abbyad, line 420: Just a note to update this based on the result of this conversation.

dianabarsan: Editing this would require quite a large refactor, I was just planning to edit the config.

dianabarsan: This part: https://github.com/medic/cht-docs/pull/504/files#diff-03618adbb0b596d7e314daf185764395f0b911b479b72ac9954722007b776c93R25

abbyad: Ah, I see. Would other potential future client-side transitions go here as well? Does having client here and client_side on the config make it more confusing? I'll let you make a call on that with app developers.

dianabarsan: It depends, right now this is an implementation detail for muting specifically, which is in itself a complex piece of work. It's used exclusively to have consistent muting while having out of order syncs. 
App developers or anyone really should be using these.

dianabarsan: But I'll do the change anyway :P

PR #506, opened 2021-05-25
Updates openmrs integration docs

### LINKED ISSUE #252: Improve 'OpenMRS'
Page: /apps/features/integrations/openmrs/
Version: 2020-06-03 15:13:10 -0400 -0400
---

We should add content to empty page.

kitsao:
Resolves #252


abbyad:
The content here seems more apt for a guide or example. @smbuthia would you agree?

derickl (commented):
It would be great if you included an illustration of how the different pieces work together

smbuthia:
I agree with Marc. I think this would be a great guide. It would probably be a better fit [here](https://docs.communityhealthtoolkit.org/apps/guides/integrations/).

PR #510, opened 2021-06-03
[To be deprecated] Documentation of RDToolkit integration

latin-panda:
Documentation for RDToolkit integration.

Ticket: https://github.com/medic/cht-core/issues/6981
CHT Core PR: https://github.com/medic/cht-core/pull/7093
CHT Android PR: https://github.com/medic/medic-android/pull/175


latin-panda:
Hi @craig-landry can you please proofread this documentation? Thanks!

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+     * Auto filled by the CHT integration.
+* `rdtoolkit_time_started` | Text
+     * The date and time when the RDT started.
+     * Auto filled by the CHT integration.
+* `rdtoolkit_time_resolved` | Text
+     * The date and time when the RDT result is ready to capture.
+     * Auto filled by the CHT integration.
+
+### Sample form for provisioning RDT
+![Provisioning RDT form](rdtoolkit-povision-form.png)
```
craig-landry, line 57: Does this image reference work? I would have expected a path of some sort to the image.

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+    * Auto filled by CHT integration.
+* `rdtoolkit_time_resolved` | Text
+    * The date and time when the RDT result is ready to capture.
+    * Auto filled by CHT integration.
+* `rdtoolkit_time_read` | Text
+    * The date and time when RDT result was captured.
+    * Auto filled by CHT integration.
+
+### Sample form for capturing RDT results
+![Capturing RDT result form](rdtoolkit-capture-form.png)
```
craig-landry, line 104: Same question about this image and if it works as expected.

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+
+The following is an example of a task implementation:
+
+```javascript
+{
+  name: 'rdtoolkit-capture-results',
+  icon: 'icon-follow-up',
+  title: 'task.rdtoolkit.capture.title',
+  appliesTo: 'reports',
+  appliesToType: ['rdtoolkit-provision'], // form
```
craig-landry, line 118: This `// form` comment looks like it could use some elaboration.

craig-landry (changes requested):

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+description: >
+  Integration for the Rapid Diagnostic Toolkit (RDToolkit)
+keywords: RDToolkit RDT
+
+---
+
+_Introduced in v3.12.0_
+
+CHT has an integration with [Rapid Diagnostic Toolkit (RDToolkit)](https://github.com/dimagi/rd-toolkit) which is an Android app developed by Dimagi that facilitates the use of Rapid Diagnostic Test (RDT) in global health care. 
+The integration is only available in the CHT Android app and consist of two Enketo widgets (rdtoolkit-provision and rdtoolkit-capture) that are used in XForms to provision RDT and capture results.
```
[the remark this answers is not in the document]
latin-panda: Fixed

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+    * Patient's first name and / or last name
+    * Auto filled when using the contact lookup field.
+* `rdtoolkit_session_id` | Text
+    * The unique ID for the RD Test.
+    * Auto filled by CHT integration.
+* `rdtoolkit_filter` | Text
+    * The group of RDT that users can select when provisioning RDT by using RDToolkit app.
+    * The available options can be found in the [RDToolkit app source code](https://github.com/dimagi/rd-toolkit).
+    * Assign a group name to enable all the RDT associated with the group, example: `mal_pf`.
+    * Assign individual RDT names to limit the options available for the user, separated by one space, example: “sd_bioline_mal_pf_pv carestart_mal_pf_pv”.
```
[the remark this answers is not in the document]
latin-panda: Fixed

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+     * Auto filled by the CHT integration.
+* `rdtoolkit_time_resolved` | Text
+     * The date and time when the RDT result is ready to capture.
+     * Auto filled by the CHT integration.
+
+### Sample form for provisioning RDT
+![Provisioning RDT form](rdtoolkit-povision-form.png)
+
+### Capturing RDT result form
+It is required to add the following labels in the “appearance” tag:
```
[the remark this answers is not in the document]
latin-panda: Nice! Thanks!

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: // The form associated with the task creation.

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: It turns out that the folder needs to be called the same as this file and it doesn't need path
Fixed!

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: It turns out that the folder needs to be called the same as this file and it doesn't need path
Fixed!

latin-panda:
Hi @craig-landry, great feedback! can you please take a look again? Thanks

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+---
+
+_Introduced in v3.12.0_
+
+CHT has an integration with [Rapid Diagnostic Toolkit (RDToolkit)](https://github.com/dimagi/rd-toolkit) which is an Android app developed by Dimagi that facilitates the use of Rapid Diagnostic Test (RDT) in global health care. 
+The integration is only available in the CHT Android app and consists of two Enketo widgets (rdtoolkit-provision and rdtoolkit-capture) that are used in XForms to provision RDT and capture results.
+
+## Create the XForms
+It is required to create one form for provisioning RDT and one form for capturing RDT results, see [the documentation](https://docs.communityhealthtoolkit.org/apps/tutorials/app-forms/) for more details about app forms.
+
```
abbyad, line 18: Out of curiosity, what happens if you have the provision and capture fields in the same form?

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+### Provisioning RDT form
+The following labels are required in the “appearance” tag:
+* `rdtoolkit-provision`: To attach the Enketo widget that connects with the RDToolkit app.
+* `rdtoolkit-action-btn`: To attach the action button that provisions RDT, this button will open the RDToolkit app.
+
+This form has the following mandatory fields:
```
abbyad, line 24: It would be helpful to specify where those fields are expected to be. Are they are the top level, or adjacent to the rdtoolkit field?

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,201 @@
...
+CHT has an integration with [Rapid Diagnostic Toolkit (RDToolkit)](https://github.com/dimagi/rd-toolkit) which is an Android app developed by Dimagi that facilitates the use of Rapid Diagnostic Test (RDT) in global health care. 
+The integration is only available in the CHT Android app and consists of two Enketo widgets (rdtoolkit-provision and rdtoolkit-capture) that are used in XForms to provision RDT and capture results.
+
+## Create the XForms
+It is required to create one form for provisioning RDT and one form for capturing RDT results, see [the documentation](https://docs.communityhealthtoolkit.org/apps/tutorials/app-forms/) for more details about app forms.
+
+### Provisioning RDT form
+The following labels are required in the “appearance” tag:
+* `rdtoolkit-provision`: To attach the Enketo widget that connects with the RDToolkit app.
+* `rdtoolkit-action-btn`: To attach the action button that provisions RDT, this button will open the RDToolkit app.
```
abbyad, line 22: I can't think of other examples that require two appearances, so I am curious why two appearance fields are needed here? What happens if only one is set?

craig-landry:
@latin-panda 
> can you please take a look again?

Thanks for the changes. I'm cool with those areas as they are now. I see @abbyad has some requested changes so I'll let him approve after those are resolved.

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: The `rdtoolkit-provision` attaches the widget and sets the scope of the widget, so we know where to display the preview and where to get the rdtoolkit exclusive fields (example: rdtoolkit_session_id field). 

The `rdtoolkit-action-btn` defines where the button to open RDToolkit app is going to display, for example we might want to display the button before a disclaimer label. If this is not provided, then the user won't be able to provide a RDTest.

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: Good question, if we put both widgets next to each other (to simulate pages), the pyxform return errors because the groups and fields have similar name. I guess if we merge the 2 into one group then probably we'll have 2 buttons displaying at the same time (provision and capture), where after provisioning will have to wait until the capture is ready to be able to safe the form..  but keeping both in the same form will require a bit of work 🤔

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: Good catch! So, I've included the image of the xls, but definitely adding the expected level will be better.
This is fixed!

content/en/apps/guides/forms/rdtoolkit-integration.md
abbyad: If omitting the `rdtoolkit-action-btn` would make it not work, it would be better to make the default button included with `rdtoolkit-provision`. If now or in the future we want to separate that out, we could add a second appearance value which makes it explicit, like `rdtoolkit-action no-button`.

latin-panda: Done! I've removed the class and updated documentation and code.

latin-panda:
Hi @abbyad, thank you for the feedback! The PR has been updated, can you please review again? Thanks for your time!

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,219 @@
...
+    * Defined in the `data` group level.
+    * Auto filled by the CHT integration.
+* `rdtoolkit_time_started` | Text
+    * The date and time when the RDT started.
+    * Defined in the `data` group level.
+    * Auto filled by the CHT integration.
+* `rdtoolkit_time_resolved` | Text
+    * The date and time when the RDT result is ready to capture.
+    * Defined in the `data` group level.
+    * Auto filled by the CHT integration.
```
abbyad, line 62: Seems like a lot of fields that need to be manually added to the XLS Form. It would have been nice for the widget to automatically add the required fields. Are they explicitly set so that the inputs can be set, and perhaps so that the output can be used in the form?

Also, if these fields persist they may be better structured as a table with columns: `Field name`, `XLSForm type`  `Description`, `Parameter type` (eg input or output), and `required`.

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,219 @@
...
+* `rdtoolkit_results_description` | Text
+    * Translation of the results
+    * Defined in the `data` group level.
+    * Auto filled by CHT integration.
+* `rdtoolkit_image` | Text
+    * Base64 image encoding.
+    * Defined in the `data` group level.
+    * Auto filled by CHT integration and will render the image.
+    * **appearance**: hidden
+    * **instance::type**: binary
```
abbyad, line 101: Same as above, would be easier to understand as a table, if these are needed to be explicitly set at all.

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,219 @@
...
+* `rdtoolkit_time_resolved` | Text
+    * The date and time when the RDT result is ready to capture.
+    * Defined in the `data` group level.
+    * Auto filled by the CHT integration.
+
+### Sample form for provisioning RDT
+![Provisioning RDT form](rdtoolkit-povision-form.png)
+
+### Capturing RDT result form
+The following label is required in the `appearance` tag of `data` group:
```
abbyad, line 68: Same as above, would be nice for the `data` group name not to be required.

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,219 @@
...
+      start: 1,
+      end: 2,
+      days: 1
+    }
+  ]
+}
+```
+
+## Add translations
+Add translations by following the [CHT guide](https://docs.communityhealthtoolkit.org/apps/reference/translations/). The following is an example of translation labels:
```
abbyad, line 182: Make sure to use the ref or relref function to [avoid broken links](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#avoid-broken-links).

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,219 @@
...
+universal_control_failure = Invalid - control failed
+```
+
+## Upload the assets
+Once the forms, the task and the translation are ready, use [medic-conf](https://github.com/medic/medic-conf) to upload these assets to the CHT Core instance.
+* [Build and upload translations](https://docs.communityhealthtoolkit.org/apps/reference/translations/#build)
+* [Convert and upload forms](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#build)
+* [Build and upload tasks](https://docs.communityhealthtoolkit.org/apps/reference/tasks/#build)
+
+Alternatively the assets can be uploaded by using the [App Manager app](https://docs.communityhealthtoolkit.org/apps/features/admin/).
```
abbyad, line 219: Again, worth fixing these to [avoid broken links](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#avoid-broken-links).

abbyad (changes requested):
I like the changes made as it is making it easier to use. Some comments inline, and the biggest one is about how the widget is structured. Ideally you could have multiples of any widgets on a page, which breaks if the group name is hard coded. Also, if we are hardcoding the widget name, `data` may be too generic and it appears as though it would cause problems if forms already have that field name. Some XLSForm converters actually name the top level group `data`, which make compatibility with broader XForms more difficult, but more practically, [forms in the default config use the `data` group already](https://github.com/medic/cht-core/blob/c0d452c63b9efefa5942696b184c44f3bba06e96/config/default/forms/app/pregnancy.xml#L913). 

Also, it seems as though this widget should be added to the list in to the [CHT XForms Widgets](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#cht-xform-widgets).

content/en/apps/guides/forms/rdtoolkit-integration.md
```diff
@@ -0,0 +1,219 @@
...
+_Introduced in v3.12.0_
+
+CHT has an integration with [Rapid Diagnostic Toolkit (RDToolkit)](https://github.com/dimagi/rd-toolkit) which is an Android app developed by Dimagi that facilitates the use of Rapid Diagnostic Test (RDT) in global health care. 
+The integration is only available in the CHT Android app and consists of two Enketo widgets (rdtoolkit-provision and rdtoolkit-capture) that are used in XForms to provision RDT and capture results.
+
+## Create the XForms
+It is required to create one form for provisioning RDT and one form for capturing RDT results, see [the documentation](https://docs.communityhealthtoolkit.org/apps/tutorials/app-forms/) for more details about app forms.
+
+### Provisioning RDT form
+The following label is required in the `appearance` tag of `data` group:
```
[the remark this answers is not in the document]
latin-panda: Originally I didn't have required groups, but then after some conversations about making specific selectors for the xml and the limited structure for making specific selectors without recurring to ids, though that must data were expected to be in "data" group.  
I'm going to look at the options to fulfil both concerns and comeback with my findings.

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: Right, I'm going to switch this documentation to a table.

> It would have been nice for the widget to automatically add the required fields.

The downside I see about automatically added fields is that it isn't obvious they exits or what is group/level or how are they called when looking at the form definition (fastest way to know the actual form fields), then app devs need to take the extra step of opening the documentation or finding the form in the db, this happens "n" times. While defining the fields is a less frequent task that happens 1 time.

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: Nice! I'll fix it

content/en/apps/guides/forms/rdtoolkit-integration.md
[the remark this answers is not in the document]
abbyad: Thanks, it's tricky, but worth keeping it as easy as possible for app developers and also maintaining compliance and styling with XForms and XLSForms.

content/en/apps/guides/forms/rdtoolkit-integration.md
abbyad: Ya, I see what you are saying, and that makes sense for the return values. It is in fact how ODK XLSForms have [intents that return multiple values](https://docs.getodk.org/launch-apps-from-collect/#external-apps-to-populate-multiple-fields). 

For the input parameters, you can see the way ODK handles that in their [external intent notation](https://docs.getodk.org/launch-apps-from-collect/#specifying-a-uri-as-intent-data). In the `body::intent` column they pass the URI and then the parameters with the XPath notation (eg `uuid=/myform/group/patient_uuid, ...`) or XLSForm variable notation (eg `uuid=${patient_uuid}, ...`). Using this notation makes it clean and more expected to XLSForm developers, and with the input parameters included like this we don't require unnecessary data fields either. There are some tricky aspects of how you can pass data to the XForm from XLSForm (eg like `body::intent` as used in external intents, and `instance::type`). You might find [this documentation on prefixes that add data to XForm attributes](https://xlsform.org/en/#advanced-use-and-extensibility) helpful for that. If you have a proposed format I am happy to review together and discuss before you reimplement it.

latin-panda: I'm learning this form tech on the way 😅 and not sure if I'm understanding the idea 🤔 

If using the ODK: 
As I can read the [external intent in general](https://docs.getodk.org/launch-apps-from-collect/#specifying-a-uri-as-intent-data) are available through their Android App (to be able to trigger intents) that would require users to install this extra app [ODK Collect](https://docs.getodk.org/collect-install) (+ cht-android, + rdtookit app) which will change the current flow too.

If not using that, then:
I think it's possible to create a custom column with the parameters, instead of typing these in rows then they will type them in 1 cell like this: 

[   myColumn  ]
[  rdtoolkit?filter="mal_pv"&api="https://example.com"   ]
Basically just `rdtoolkit_filter` and the `rdtoolkit_api_url` because the rest are autofilled, this solution is probably is new for us 🤔.

Because fields like `patient_uuid` still needs to be there since that's coming from the `contact` source control and managed somewhere else in cht 🤔 gotta test. Regarding the `patient_id`, `patient_name` and `rdtoolkit_session_id`, I think those should stay as fields, because they probably want to display it as read only to cross check with the dashboards or the rdtoolkit app. 

I try to imagine myself filling the form as app dev to feel the difference between solutions, however this is something I would suggest to test with users because of effort/time in development/testing/review, we'll probably get insights about other pieces of this feature too. Otherwise we can go ahead with the second solution (custom column) 🙂

content/en/apps/guides/forms/rdtoolkit-integration.md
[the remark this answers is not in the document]
latin-panda: I think I've found a middle point:

We were searching by just field name prefixed by "rdtoolkit_" but there was a concern and moved it to specific selector:
Before:
 `$widget.find('input[name+="FIELD_NAME]"')`

Now it includes the form id and the group name `data`:
`$widget.find('input[name^="FORM_ID/data/FIELD_NAME]')`

[Recommended] We can relax the specific selector a bit, by looking for something that starts with the form id, omit anything in the middle (groups, or nested groups) and that ends with the field name, like a sandwich:
`$widget.find('input[name^="/FORM_ID"][name$="/FIELD_NAME]');`

Otherwise we might need to have our own group name: "rdtookit" 🙃

Note: we know what is the form id, it's variable provided by Enketo. I mean not hardcoded.

To confirm, is the sandwich solution okay? 🥪

content/en/apps/guides/forms/rdtoolkit-integration.md
abbyad: Sorry for the confusion! No need to use ODK Collect itself, just imitate it's technique. Most forms that work in ODK Collect would work in the CHT, but the reverse is not always true because of CHT-specific widgets (eg the contact-selector).

In terms of other input fields, they might be in the form, but not necessarily in the same group. For consistency with external intents, I'd suggest a `body::intent` column have the following:
`org.rdtoolkit.provision(filter="mal_pv", api="https://example.com", uuid=/form/inputs/patient_uuid, id=${patient_id}, name=${patient_name})`

If you want to use an `appearance` of `rdtoolkit` that would ensure that the rdtoolkit widget is used, which may make it more future proof if a generic external intent widget is added _later_ (which would be a nice feature, and help us be compatible with ODK XForms).

That makes it have the following **inputs** passed in the _parameter list_, all of which we'd generally use but in theory could be optional:
- `id` (required, only to differentiate tests)
- `name` (required, only to differentiate tests)
- `uuid` (how is this used?)
- `filter`
- `api`

And the **return** values as **fields within the same group**, all of which could be optional:
- `rdtoolkit_session_id`
- `rdtoolkit_state`
- `rdtoolkit_time_started`
- `rdtoolkit_time_resolved`

Something similar would also need to be done for the capture side.

content/en/apps/guides/forms/rdtoolkit-integration.md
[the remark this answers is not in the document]
abbyad: I am not certain of the best approach for the selector... but given that we do this in other widgets there should be a way to get this to work without hardcoding the field name. Perhaps this [section of the mrdt widget](https://github.com/medic/cht-core/blob/6e82151d120e57f2332d6cc1ea78da7a5bae69f9/webapp/src/js/enketo/widgets/mrdt.js#L49-L66) can serve as an example? Given @garethbowen's [work on the mrdt widget](https://github.com/medic/cht-core/commit/50c21bf39a17a5c242492c0359db451e46002061#diff-504326969274038d01a4fffb98edf8c4905c1be1f0e6545296a9fefe77e202c9) he may be a good resource for this.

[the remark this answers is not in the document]
latin-panda: Looking at it, mrdt is using appearance tags as field name like: [or-appearance-mrdt-time-taken](https://github.com/medic/cht-core/blob/6e82151d120e57f2332d6cc1ea78da7a5bae69f9/webapp/src/js/enketo/widgets/mrdt.js#L61) to find and assign information into fields, but it's not really different as it has the same issue about not having specific selectors and more than one could match. 

I think the main difference with this integration is that we need to send more than 1 field and we need to receive more than 1 as well. Fields names won't be avoidable in this case as we need to make sure we're assigning the information in the right place.

I'm going to implement this solution that relaxes the selector a bit: `$widget.find('input[name^="/FORM_ID"][name$="/FIELD_NAME]');`

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: Basically, support both form fields and "query string".  I'm going to do some experiments, it might take some time as I research

abbyad: Right, and to clarify _inputs_ would only be via "query string", and _outputs_ would be form fields.

content/en/apps/guides/forms/rdtoolkit-integration.md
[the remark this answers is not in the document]
abbyad: By `[name$="/FIELD_NAME]'`, are you talking about the `or-appearance-*` name, or the XLSForm field name? If it is the `or-appearance-rdtoolkit-provision` then I think that's the more standard way to do it.

content/en/apps/guides/forms/rdtoolkit-integration.md
latin-panda: From the UX/UI perspective, we still need the patient data (inputs) as fields because that's how the CHW know who they are provisioning RDTs: 
Screen Shot 2021-06-14 at 11 12 28 am 

Patient's uuid is needed for the contact source field association the patient, example: if there's a value it won't ask for the contact, if it's empty then the form ask for the contact (that's how it's working already in cht) it saves one step to the CHW:
Screen Shot 2021-06-14 at 10 02 09 pm

What I'm going to do is to leave open the possibility of receiving the contact information by query string, if this information is missing then I'll take it from the fields. I'll explain both in the docs, but the example probably more oriented to having the patient as fields since it's probably the most common scenario.  🤔

content/en/apps/guides/forms/rdtoolkit-integration.md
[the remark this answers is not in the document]
latin-panda: `FIELD_NAME` is the element name in the XML tag and from the "name" column in the XLS, the standard in both.

Screen Shot 2021-06-14 at 10 10 35 pm

I think using appearance or any extra tag to identify fields is overhead, adds to complexity when building the forms.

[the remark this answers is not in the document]
abbyad: Using the appearance is the convention used in XLSForm for widgets like this, so I would follow that convention for the rdtoolkit group. The inner fields which are outputs can have set field names, as done in the external intent example.

content/en/apps/guides/forms/rdtoolkit-integration.md
abbyad: Similar to [above](https://github.com/medic/cht-docs/pull/510#discussion_r651094219), we should be following the XLSForm convention here and having the inputs in the query string as the main way to use this widget. Generally those patient fields are already at the top level, and don't need to be repeated, which causes duplication in the resulting data as well. Being able to have the fields in the group is a nice _fallback_ if fields are missing in the query string though, but may be adding unnecessary complexity to code and documentation. Feel free to follow up offline if further clarification is needed.

latin-panda:
Deprecated by a different approach, please refer to related issue for more details.

PR #517, opened 2021-06-16
add task_id field

njogz:
Update documentation as a result of work on this [ticket](https://github.com/medic/cht-core/issues/6690).


dianabarsan (approved):
nice, thanks for this.

garethbowen:
@njogz I think this can be merged now, right?

PR #520, opened 2021-06-22
CHT Core Starter Documentation

billwambua:
Original https://github.com/medic/app-services-team/blob/master/cht-core-starter/README.md

### DIFF
--- a/content/en/apps/guides/hosting/switching-cht-versions-locally.md
+++ b/content/en/apps/guides/hosting/switching-cht-versions-locally.md
@@ -12,7 +12,7 @@ relatedContent: >
 
 ## TLDR;
 
-When developing configuration locally, especially when working with multiple projects with different versions, it is sometimes convinient to:
+When developing configuration locally, especially when working with multiple projects with different versions, it is sometimes convenient to:
 
 - Persist a database for each project when working across multiple projects.
 - Start the right version of cht-core when working across different projects.


kennsippell:
@billwambua The officially supported method for running CHT services is [via docker](https://docs.communityhealthtoolkit.org/core/guides/docker-setup/) - and I believe docker remains the long-term bet for the CHT community. The "cht-core-starter" script is a bash script that is a useful workaround for problems until medic/cht-core#7061 is ready. We need something - because issues like https://github.com/medic/cht-core/issues/6168 make docker an invalid option for cht app developers... But I'm not sure the script is even available anywhere publicly (https://github.com/medic/app-services-team/blob/master/cht-core-starter/README.md is in a private repo) -- so we are making public docs for a private script here. 

When medic/cht-core#7061 is fixed, I believe the cht-core-starter script will die.  There have been fixes scheduled for improvement since 2019, so most of us have assumed these docker issues would be solved quickly - but that hasn't been the case. Whatever the history, the cht-core-starter scripts have been kept internal to Medic Mobile teams and it doesn't make sense to have public documentation for a private script.  We should either open up the script to the community (if there is a demand for it), or (if there is no demand for it) close this PR and keep it internal as we anticipate the medic/cht-core#7061 fix.

mrjones-plip:
@kennsippell  - I missed [your comment back in July](https://github.com/medic/cht-docs/pull/520#issuecomment-875285602) - sorry! 

I think the fix to #medic/cht-core/issues/6168 isn't tied to docker per se, but instead is fixed by having the ability to trivially add a valid TLS certificate to your local setup.  This is solved for `medic-os` docker containers with the [add-local-ip-certs-to-docker.sh script](https://github.com/medic/cht-core/blob/master/scripts/add-local-ip-certs-to-docker.sh).  Or, if an app developer is not running docker containers, and is running either [Core Starter](https://github.com/medic/app-services-team/tree/master/cht-core-starter) or a [developer set up](https://github.com/medic/cht-core/blob/master/DEVELOPMENT.md), they can use [nginx-local-ip](https://github.com/medic/nginx-local-ip), which is agnostic to what ever back end you're using (even beyond any CHT setups).

At a more holistic level, my apologies for not moving #medic/cht-core/issues/7061 forward.  I'll reach out to @binokaryg and figure next best steps.

billwambua:
@kennsippell true. CHT-Core Starter being a private script makes this documentation unuseful for now. 

I think I made it with the idea that the script would be made open soon.

I'll close it for now. If we do end up opening up the script to the public this can be re-opened.

@mrjones-plip good pointers.

PR #521, opened 2021-06-24
add connected_users.count to the monitoring api documentation

njogz:
Add `connected_users.count` to the monitoring api documentation


latin-panda (approved):
Thank you!!

PR #523, opened 2021-06-28
Documentation for CHT Application tests

pkitutu:


content/en/apps/guides/application-tests/features_to_test.md
```diff
@@ -0,0 +1,62 @@
+---
+title: "Common features of the app that require automated testing"
+linkTitle: "Features to test"
+weight: 2
```
kennsippell, line 4: i'm a bit unclear on the difference between this content vs [this existing content](http://docs.communityhealthtoolkit.org/cht-conf-test-harness/). I'm wondering if they can somehow be effectively combined?  I think the location of this content is great, and there are some additions here that seem very valuable - but having deep links into the harness documentation would be very valuable. For example `getContactSummary` should link to [this](http://docs.communityhealthtoolkit.org/cht-conf-test-harness/Harness.html#getContactSummary).

pkitutu: Thanks for the feedback @kennsippell. I think we can everything incorporated in the new documentation or we could just update the old documentation. Will make the suggested deep links

kennsippell:
@pkitutu  Confirming that this is still on your radar?

pkitutu:
[Issue #32](https://github.com/medic/app-services-team/issues/32) taken on by @binokaryg

binokaryg:
New PR: #698

Issue #525, opened 2021-07-07
No option to run CI for external contributions


dianabarsan:
Relevant PR: https://github.com/medic/cht-docs/pull/485

Observed on CHT-Core, (sometimes?) when PR is submitted from a fork, a user with write permissions (just guessing here) sees a button to run CI for the PR: 
![image](https://user-images.githubusercontent.com/35681649/124700697-d5c8ea00-def5-11eb-9aca-7b3543209cf2.png)
This button wasn't available for me on cht-docs. 

Here is some detail into how this can be set up: https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#configuring-required-approval-for-workflows-from-public-forks

garethbowen:
I checked and it looks like the configuration is set correctly. I think the problem is to do with the actions.yml which is set to build on branch, not on pull_request, eg:

```
on:
  push:
    branches:
      - '*' # branch checking is below so all this is one CI file
```
Whereas the cht-core build.yml is set like..


```
on: [push, pull_request]
```

mrjones-plip:
@dianabarsan  - fix should be in to address this.  To test it is indeed fixed, we need to have a GH account that's not in Medic org fork, do a commit and then create a PR requesting to merge to our `master` branch?

PR #528, opened 2021-07-14
Allow CI runs for external contributions

mrjones-plip:
per #525

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,13 +4,7 @@
 #  - Hugo builds the site twice: once for link checking and once minified for pushing live
 #  - Bash script for muffet is intentionally stand alone so devs/content editors can easily run it locally
 
-on:
-  push:
-    branches:
-      - '*' # branch checking is below so all this is one CI file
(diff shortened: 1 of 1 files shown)


mrjones-plip:
Hmmm - followed [same model as cht-core CI](https://github.com/medic/cht-core/blob/d97ce79b34b729286fbab40c8283261f4994a107/.github/workflows/build.yml#L3), but I'm seeing two builds now where as before we only had one.  Thoughts @garethbowen ?

garethbowen (approved):
Looks good to me.

cht-core also runs two builds for PRs. The first checks that the code in the branch passes, and the second checks that if the code were merged into master it would still pass. This is useful if someone took a branch and linked to a doc, then someone else merges a change to master that moves the doc. In this case one build will pass and one will fail.

I assume it's difficult to test that this solves the problem until this PR is merged?

mrjones-plip:
awesome - thanks @garethbowen.  Yeah - sad path testing can only be done after we merge to master. I'll do that now and the figure how to do an after the fact test.

PR #529, opened 2021-07-14
Update references from medic-conf to cht-conf

jkuester:
Was going through some of the tutorials in these docs and noticed that the references to `medic-conf` no longer matched up with the documentation out here https://github.com/medic/cht-conf/blob/master/README.md

I figured it would be best to just update all the references to `medic-conf` to properly refer to `cht-conf/cht`.


mrjones-plip (changes requested):
@jkuester  - Thanks so much for grabbing this! 

I know you hit up @latin-panda  for the PR, but I took a quick looksee and it looks like there's 41 instances (just using "ctrl + f" in the browser) of `cht ` in this PR where `medic-conf` got some how converted to just `cht` with the `-conf` missing.

Few examples are [here](https://github.com/medic/cht-docs/pull/529/files#diff-85cba429bc2c56379838fefcb4af6fa7ed9d1cad770b68142a6d64df7efa9202R23), [here](https://github.com/medic/cht-docs/pull/529/files#diff-102020168218952621dd22924de794747ba46dc25c539f5a16f0b3129b56141eR54) and [here](https://github.com/medic/cht-docs/pull/529/files#diff-9c35007030714ade173b7a5eca7f439524fdd54c0831e4b4656648ef29515483R134).

latin-panda (changes requested):
Thank you for sending this! I left a comment.

Thanks @mrjones-plip :) however the new name for the cmd line is `cht`: [read me](https://github.com/medic/cht-conf/blob/master/README.md), [warning](https://github.com/medic/cht-conf/blob/35c03c8993de231ef66b4c7d124f4021c2b9e69a/src/lib/main.js#L54).

Not necessarily this PR scope, but for existing projects, the switch from `medic-conf` -> `conf` in cmd line can make people lost at first, even though we announced the change. I'd be good to announce this in the CHT-Docs as well. Can we ping relevant forums post in the CHT-Docs front page? 
Also, it'd be good to have a page in the CHT-Docs where it explains how CHT-Conf works and the available commands, as centralised additional reference. 
@mrjones-plip @jkuester  what do you think? If this makes sense, I can create the relevant tickets.

content/en/apps/tutorials/application-settings.md
```diff
@@ -186,7 +186,7 @@ Configure the CHW role's depth to 2 by adding the following key/value pairs to t
 To upload app settings to your local instance, run the following command:
 
 ```zsh
-medic-conf --url=https://<username>:<password>@localhost --accept-self-signed-certs upload-app-settings
+cht-conf --url=https://<username>:<password>@localhost --accept-self-signed-certs upload-app-settings
```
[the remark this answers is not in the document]
jkuester: Shoot. Missed one.  Good catch!

jkuester:
@latin-panda Yeah, to me it seems like a good idea to announce this change somewhere in the CHT-Docs as well. The "discoverability" of the command changing is not great (if you don't read the release notes or the cht-config docs). Basically I just knew the package name had changed. So I installed the new cht-conf package. Then I tried running `cht-conf ...` and got an error saying the command was not found. Tried `medic-conf ...` and that worked but it gave me the deprecation warning. Then I knew I should just use `cht ...`  I am not sure what the best format would be for communicating the change, but if you all have an idea for that I am happy to include it in this PR.

Also, +1 for a page in the CHT-Docs where it explains how CHT-Conf works. It would not have to be too detailed either if we decided it would be better to just link back to the cht-conf README, but having a landing/overview page in the docs makes sense.  (Maybe add a new cht-conf page under CHT Applications > Reference?)

mrjones-plip (approved):
@latin-panda  - oh!  Thanks so much for setting me straight on `cht` vs `cht-conf` :sweat_smile: 

Approving this as is and deferring @latin-panda to unblock!

mrjones-plip:
> @latin-panda Yeah, to me it seems like a good idea to announce this change somewhere in the CHT-Docs as well. The "discoverability" of the command changing is not great (if you don't read the release notes or the cht-config docs). Basically I just knew the package name had changed. So I installed the new cht-conf package. Then I tried running `cht-conf ...` and got an error saying the command was not found. Tried `medic-conf ...` and that worked but it gave me the deprecation warning. Then I knew I should just use `cht ...` I am not sure what the best format would be for communicating the change, but if you all have an idea for that I am happy to include it in this PR.
> 
> Also, +1 for a page in the CHT-Docs where it explains how CHT-Conf works. It would not have to be too detailed either if we decided it would be better to just link back to the cht-conf README, but having a landing/overview page in the docs makes sense. (Maybe add a new cht-conf page under CHT Applications > Reference?)

+1 to all this!! @smbuthia  - any thoughts on socializing the change in the util name from `medic-conf` -> `cht`?  Forums? Blog post is maybe too much?

Otherwise, I'd say spin up a new `cht-docs` ticket to document usage of the `cht` util.  We can decide where it lives at that time (and we can always move it if one of you busts this out right away).

content/en/apps/tutorials/local-setup.md
```diff
@@ -64,42 +64,42 @@ Once the command is done running, navigate to [https://localhost](https://localh
...
 
 <br clear="all">
 
  *****
 
-### 2. Install medic-conf
+### 2. Install cht-conf
 
-Using npm and python on your terminal, install medic-conf and pyxform globally using the following commands:
+Using npm and python on your terminal, install cht-conf and pyxform globally using the following commands:
```
mrjones-plip, line 75: there's actually [a screen shot](https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/#2-install-medic-conf) on this page that we should update too to match the new name!

jkuester: Oh good call!  Switched it for an updated screenshot.

latin-panda (approved):
LGTM! ✨✨
I've added a [ticket](https://github.com/medic/cht-docs/issues/530) for the CHT-Conf page

PR #531, opened 2021-07-23
document docker compose for app devs

mrjones-plip:
See 7211 #medic/cht-core/issues/7211


mrjones-plip:
Ugh - not sure what's going on with all these extraneous commits in this PR. I guess I forked off of not master?  Starting again.

PR #544, opened 2021-08-26
timeout-minutes set to 15

mrjones-plip:
best practice is to always have an [upper bounds](https://github.com/marketplace/actions/enforce-timeout-minutes) for a GH action run time.  Looking at [existing actions](https://github.com/medic/cht-docs/actions), our average time is about a 1m5s, so 15 min should be plenty of extra room!


garethbowen:
@mrjones-plip I think this is ready to merge?

mrjones-plip:
@garethbowen  - thanks so much for reviewing older PRs!  Nudge much appreciated.

PR #555, opened 2021-09-27
Update impact-metrics.md

helizabetholsen:
Updated language to reflect caring activities metric language rather than CHW actions. Also updated to include "frontline health workers" in our activity metrics and changed from activity to engagement in-line with our M&E work


garethbowen (approved):
Thanks!

PR #557, opened 2021-09-29
Rebranding of app dependencies

garethbowen:


garethbowen:
@mrjones-plip I just noticed these were woefully out of date during the standup this morning...

content/en/core/overview/supported-software.md
```diff
@@ -9,7 +9,7 @@ relevantLinks: >
 
 Medic supports minor versions of the CHT Core Framework for three months after the next minor version is made available, and the latest minor of a major version for twelve months after the next major version is made available.
 
-Once a version is no longer supported it will not receive any further releases and upgrading may be required before Medic can investigate any issues.
+Once a version is no longer supported it will not receive any further patch releases.
```
mrjones-plip, line 12: We don't actually talk about what a "patch release" is and only mention "semver" in [unrelated passing](https://docs.communityhealthtoolkit.org/core/guides/releasing/#android-apps). I think the old language is more both more explicit to non-technical readers so they understand what they have to do to get a fix (MUST UPGRADE!).  Not that big a deal, but maybe there's a way we can merge the "upgrade" language with the "patch release" language?

garethbowen: Mostly I was trying to drop the word "Medic" out of here, because anyone can investigate issues... what about "... and upgrading to a supported version will be required to resolve any issues you have." (or something?

mrjones-plip: This the full new sentence then?

> Once a version is no longer supported it will not receive any further patch releases and upgrading to a supported version will be required to resolve any issues you have.

Yes, this looks good!

garethbowen, line 12:
```suggestion
Once a version is no longer supported it will not receive any further patch releases and upgrading to a supported version will be required to resolve any issues you have.
```

mrjones-plip (approved):
lgtm!

PR #570, opened 2021-11-10
update getMostRecentReport function description

### LINKED ISSUE #569: Update cht-nootil function description
We need to update `getMostRecentReport` in https://github.com/medic/cht-docs/blob/master/content/en/apps/reference/_partial_utils.md due to a change in the function's implementation https://github.com/medic/cht-nootils/pull/23

freddieptf:
Fixes #569


smbuthia (approved):
Looks good!

PR #573, opened 2021-11-12
6345 Fix jquery version documented to have select2 bug

jkuester:
The current "Update dependencies" documentation states that version `5.6.0+` of jquery causes issues when used with select2.  This should actually be version `3.6.0+` since there is no `5.6.0` of jquery (currently `3.6.0` is still the latest).


dianabarsan (approved):
Lol, this was actually funny.

PR #581, opened 2021-11-24
Update local-setup documentation

tatilepizs:
Update how to bypass denial page mac+chrome


content/en/apps/tutorials/local-setup.md
```diff
@@ -37,15 +37,15 @@ To read more about these concepts, see our [Docker Setup guide]({{< relref "core
 
 Before you begin, you need to have some useful software and tools that are required for things to work:
 
-* [nodejs](https://nodejs.org/en/) 8 or later
+* [nodejs](https://nodejs.org/en/) from version 8 to 12
```
jkuester, line 40:
```suggestion
* [nodejs](https://nodejs.org/en/) version 12 or later
```

I believe that now at least Node 12 is required for _building_ cht-core.

jkuester (changes requested):
+1  Thanks for helping to keep these docs up to date!  (Just one minor change suggested)

content/en/apps/tutorials/local-setup.md
tatilepizs: Does it work with a node version higher than 12? I thought that we couldn't install the latest one

jkuester: So, I am pretty sure that it _should_ work with higher versions and I know that folks have tried it before with Node 16 (and I thought they got it working).  But, just now I checked the Node versions that we use in our [CI builds on GitHub actions](https://github.com/medic/cht-core/blob/master/.github/workflows/build.yml) and I see that we only try to build the project with Node 12 and then test running it with Node 8, 10, and 12.  So, you are probably right that we should just suggest version 12 here as the recommended version for development (and not have "or later").

tatilepizs: Done, I changed it to version 12

jkuester (approved):
Looks great!  You can go ahead and merge this PR.

PR #582, opened 2021-11-29
Correcting example code for `defaultResolvedIf`

binokaryg:
The guide text says:
> You can also use `this.definition.defaultResolvedIf`

But the code snippet shows: `this.defaultResolvedIf`, which does not work


garethbowen (approved):
Great catch, thanks!

PR #591, opened 2022-01-05
Add clarity on couchDB port number when replicating data

jonathanbataire:
# Description 
CouchDB listens on port 5984 by default however this is not always the case as i have come to experience. Personally mine runs on 5985. Currently the documentations assumes couch is listening on the default port and offers no information on errors that may result if its not the default
## Type of change
- Clarify on the port number to cater for cases when it's not the default.


content/en/apps/guides/debugging/replicating-production-data-locally.md
```diff
@@ -38,7 +38,7 @@ If the instance is too large to replicate locally (or you are too impatient), yo
...
 1. Wait for the data to replicate. You know this is done once the app lets you interact with it. _(If you want to get the user's data before purging, consider disabling purging. Another exercise for the reader)_
 1. Make sure your local CouchDB has CORS enabled: http://docs.couchdb.org/en/2.1.1/config/http.html?highlight=CORs. Consider using [add-cors-to-couchdb](https://github.com/pouchdb/add-cors-to-couchdb#user-content-what-it-does) or its recommended settings.
-1. Allow your CouchDB to be accessible via `https`. One way is to run `ngrok http 5984` to make your CouchDB accessible via a url like `https://abcd1234.ngrok.io`.
+1. Allow your CouchDB to be accessible via `https`. One way is to run `ngrok http XXXX`. Here `XXXX` represents CouchDB HTTPS port number _(it's `5984` by default)_ which can be found in the [database configuration](https://docs.couchdb.org/en/stable/config/http.html#chttpd/port) under `chttpd`. This will make your CouchDB accessible via a url like `https://abcd1234.ngrok.io`.
```
derickl, line 41: Instead of XXXX maybe use <port>? ngrok http <port>

Also, it's unlikely one would find a non-standard port number documented in the couchdb docs. But, reading through the couchdb config files documented [here](https://docs.couchdb.org/en/stable/config/intro.html) may be of help (are you able to verify if you find anything useful in your local config file?)

derickl:
@jonathanbataire could we wrap this up?

content/en/apps/guides/debugging/replicating-production-data-locally.md
kennsippell: can you just use `x` as couchdb's non https port (5988) and ngrok handles all the https? 
does that work? is that simpler?

jonathanbataire: Derick its easier to access db configuration through fauxton under `config`.

garethbowen:
@jonathanbataire Is this ready for review again?

derickl (approved):
Looks good

PR #594, opened 2022-01-06
Add a link to side-load steps, add clarity on how to enable side-loading

mrjones-plip:


content/en/core/guides/android/publishing.md
```diff
@@ -39,9 +39,9 @@ New apps cannot longer be uploaded with the APK format in the Play Store. Apps c
 
 This method gives an app developer full control over installation. It is also possible to do the installation without an internet connection which makes it ideal for remote installation, or to save bandwidth when performing multiple installs.
 
-1. In the phone settings select the option to "Allow installation of non-Market applications"
-2. Download the correct APK on to the phone. It's important to select the right APK for the instruction set and Android version, as documented [in this table]({{< ref "core/guides/android/development-setup#apks" >}}).
-3. Select to install the APK.
+1. In the phone settings [select the option](https://www.tech-recipes.com/rx/22616/android-allow-installation-of-non-market-apps/) to "Allow installation of non-Market applications"
```
dianabarsan, line 42: Finding the correct guide is pretty challenging. This guide is from 2012, last updated in 2018, and provides instructions for Android 2.2 -> Android 8. Even though instructions for Android 8 might still apply for newer versions, we still might get forum questions about it. The page is also stock full of ads. 
Can we maybe find a more official or professional guide to link in our docs? 

Additionally, since Android 4 to Android 7. the settings is described as "Allow installation of apps from other sources than the Play Store", while in Android 8 and above, it's described as "Install unknown apps". In my opinion, either one of these is better than the antiquated "Allow installation of non-Market applications" (How many of our users know that the Play Store was once called "Android Market"?).

dianabarsan (commented):
Whoa, super quick reaction times! Thanks! 

I left one comment inline.

content/en/core/guides/android/publishing.md
mrjones-plip: Ah - thank you!! My ad blocking set up filtered out the ads, and I should have paid closer attention to how up to date the guide is.  I agree we should get something more current!  I'll have another pass at finding a better guide and ping ya.

mrjones-plip: The new link I just committed is super official (android.com), but not nearly as helpful :shrug:  Very open to input for a better/different link!

dianabarsan (approved):
Excellent. I agree that it's less beginner friendly. 
On the other hand:
1. it's more professional
2. inspires greater confidence than some random tech site with 70% add content (it is instructing users to go into their phone settings and disable security features...)
3. more in line with what we would have as "medic approved instructions" 
4. covers more versions and is more likely to be up to date

Thanks for taking the time to find this!

Issue #596, opened 2022-01-06
Group guides about Tasks together


abbyad:
The last three items in the Guides section don't match the organizational hierarchy of the other guides. They could be moved to existing subsections, but would probably be more futureproof to have them combined under a Tasks section.

<img width="951" alt="image" src="https://user-images.githubusercontent.com/789512/148441793-01a62ec7-f3e3-4417-86c4-1ac5806473b1.png">

abbyad:
Just noting that these were already grouped together in `Tasks` [when introduced](https://github.com/medic/cht-docs/pull/564), but not grouped in the tree because the `Tasks` folder is missing an `_index.md` file.

PR #599, opened 2022-01-07
Add index for Tasks

abbyad:
For #596

Groups the existing task guides that were grouped in a file system folder, but weren't represented in the tree menu.


craig-landry (approved):

mrjones-plip:
thanks for grabbing the review @craig-landry !

PR #605, opened 2022-01-11
603 Remove unnecessary right padding

### LINKED ISSUE #603: Avoid unnecessary wrapping in menu
Multi word menu items are being wrapped even when there is sufficient space to fit on a single line. This screenshot is of [the "features'"landing page](https://docs.communityhealthtoolkit.org/apps/features/) as of f4993b57b165ea6dc9c851773c69c2695d1c5f8c:

image

A right padding is added at each level of the tree, so it is especially bad for lower level subitems. This can be easily corrected in our custom menu partial by removing the right padding styling `pr-md-3`.

image

abbyad:
Fixes #603

The fix looks good in all the widths I tested, and doesn't have any known negative consequences. @mrjones-plip as part of your review can you please confirm that it looks good for you as well?


mrjones-plip (approved):
lgtm! works in the desktop mode and tablet mode - nice tweak!

Issue #601, opened 2022-01-11
Improve 'How to bulk load users'


abbyad:
Page: /core/guides/users-bulk-load/
Version: 2020-06-03 16:12:04 -0700 -0700
---

This page seems as though it would be relevant for app management, rather than core development. I propose we move it to [apps/guides/data/](https://docs.communityhealthtoolkit.org/apps/guides/data/) where we have similar content, like the guide on `csv-to-docs`.

PR #607, opened 2022-01-11
Move bulk load users page

abbyad:
For #601

### DIFF
--- a/content/en/apps/guides/data/users-bulk-load.md
+++ b/content/en/apps/guides/data/users-bulk-load.md
@@ -4,6 +4,8 @@ linkTitle: "Bulk Load Users"
 weight: 15
 description: >
   Notes for how to bulk load users
+aliases:
+    -    /core/guides/users-bulk-load
 ---
 
 First, using the webapp, create the top level places/facilities like
--- a/content/en/apps/concepts/users.md
+++ b/content/en/apps/concepts/users.md
@@ -8,7 +8,7 @@ relatedContent: >
   design/personas
   apps/tutorials/contact-and-users-1
   apps/tutorials/contact-and-users-2
-  core/guides/users-bulk-load
+  apps/guides/data/users-bulk-load
 ---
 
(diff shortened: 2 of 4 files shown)


content/en/apps/guides/data/users-bulk-load.md
```diff
@@ -4,6 +4,8 @@ linkTitle: "Bulk Load Users"
 weight: 15
 description: >
   Notes for how to bulk load users
+aliases:
+    -    /core/guides/users-bulk-load
```
mrjones-plip, line 8: Pedantic indentation parity fix:

```suggestion
  - /core/guides/users-bulk-load
```

content/en/apps/guides/data/users-bulk-load.md
```diff
@@ -4,6 +4,8 @@ linkTitle: "Bulk Load Users"
 weight: 15
 description: >
   Notes for how to bulk load users
+aliases:
```
mrjones-plip, line 7: Aside from adding an alias to catch the redirect, I feel like we can jazz things up with a link/read more/what not to:
* [CSV To Docs](https://docs.communityhealthtoolkit.org/apps/guides/data/csv-to-docs/)
* [`cht-conf`](https://github.com/medic/cht-conf) (which is strange we link to the GH repo in a bunch of docs places, but a fix for another day)

mrjones-plip (approved):
approving with some optional fixes.

content/en/apps/guides/data/users-bulk-load.md
abbyad: Thanks for noticing this! I had copied it from a page that had this formatting. I'll update both to be like the other 8.

content/en/apps/guides/data/users-bulk-load.md
abbyad: Good idea. I have made an attempt in https://github.com/medic/cht-docs/pull/607/commits/c394a35240cb894f510724f989208db596382a80 with a page header.

image

PR #614, opened 2022-01-20
Update `POST /api/v1/users` documentation for bulk user creation

m5r:
This PR documents how to create many new users with the `POST /api/v1/users` and reflects its actual permissions requirements according to [the current state of the code](https://github.com/medic/cht-core/blob/bb825795a52b2e0443b0070fdb8433b215657852/api/src/controllers/users.js#L129)

[Related cht-core PR](https://github.com/medic/cht-core/pull/7454)

medic/cht-core#7490

### DIFF
--- a/content/en/apps/reference/api.md
+++ b/content/en/apps/reference/api.md
@@ -970,10 +970,19 @@ Content-Type: application/json
 <summary class="text-primary mb-3">Response if the username already exists</summary>
 
 ```
-HTTP/1.1 409 Conflict
-Content-Type: text/plain
+HTTP/1.1 400 Bad Request
+Content-Type: application/json
 
-Document update conflict.
+{
+  "code": 400,
+  "error": {
+    "message": "Username \"mary\" already taken.",
+    "translationKey": "username.taken",
+    "translationParams": {
+      "username": "mary"
+    }
+  }
+}
 ```
 </details>
 
--- a/content/en/apps/reference/api.md
+++ b/content/en/apps/reference/api.md
@@ -820,6 +820,17 @@ Content-Type: application/json; charset=utf-8
 
 Create new users with a place and a contact.
 
+ Creating multiple users at once by passing an array of users was introduced in version 3.15.  
+All users need to meet the following requirements before any of them are created:
+- All required fields are filled in
+- The password is at least 8 characters long and difficult to guess
+- The phone number is valid when [`token_login`]({{< ref "apps/reference/app-settings/token_login" >}}) is enabled
+
+Users are created in parallel and the creation is not aborted even if one of the users fails to be created.
+
+Passing a single user in the request's body will return a single object whereas
+passing an array of users will return an array of objects as shown in the examples below.
+
 #### Permissions
 
 `can_create_users`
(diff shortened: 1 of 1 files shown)


dianabarsan (commented):
Nice one! I left a couple of requests inline.

content/en/apps/reference/api.md
```diff
@@ -870,17 +872,161 @@ Content-Type: application/json
...
+  }
+]
+```
+</details>
 
 #### Errors
 
-Response if the username already exists:
+<details>
+<summary class="text-primary mb-3">Response if the username already exists</summary>
```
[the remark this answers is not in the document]
m5r: Yes, good catch! I updated it with the correct error returned

content/en/apps/reference/api.md
```diff
@@ -818,14 +818,27 @@ Content-Type: application/json; charset=utf-8
 
 ### POST /api/v1/users
 
-Create a new user with a place and a contact.
+Create new users with a place and a contact.
+
+ Creating multiple users at once by passing an array of users was introduced in version 3.15.  
+All users need to meet the following requirements before any of them are created:
+- All required fields are filled in
```
dianabarsan, line 825: Are the required fields listed somewhere?

m5r: Yes, they're mentioned just above, at the beginning of the `Users` section  
https://docs.communityhealthtoolkit.org/apps/reference/api/#supported-properties-2

dianabarsan: Awesome!

dianabarsan (approved):
⭐

mrjones-plip:
Let's revert this to not use the `<details>` expandable sections and keep it all inline. Here's why:

The API page is so very long that I kinda feel like the introduction of the new `<details>` stanzas is nice, as it hides the content in a helpful way:

image

But, in the end I think the API page has a good table of contents on the right. Also, while long, you can use your browsers "find" feature (something developers will know) to find content and the collapsible sections break this.  Finally, this inst inline with our [current code formatting styles](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#code-snippet-formatting). 

I know @abbyad is refactoring content based on some app dev feedback, so in a later ticket we might do a site wide improvement to collapse all code snippets/sections on longer pages.

mrjones-plip (changes requested):
Let's revert this to not use the `<details>` expandable sections and keep it all inline. Here's why:

The API page is so very long that I kinda feel like the introduction of the new `<details>` stanzas is nice, as it hides the content in a helpful way:

image

But, in the end I think the API page has a good table of contents on the right. Also, while long, you can use your browsers "find" feature (something developers will know) to find content and the collapsible sections break this.  Finally, this isn't inline with our [current code formatting styles](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#code-snippet-formatting). 

I know @abbyad is refactoring content based on some app dev feedback, so in a later ticket we might do a site wide improvement to collapse all code snippets/sections on longer pages.

m5r:
Thanks for the comments @mrjones-plip, I wanted to have a clear separation for examples but you're right in that it diverges from our current formatting style. I'm reverting this change to keep it consistent, we can always come back to it at a later point to make a site-wide change like you said :+1:

mrjones-plip (approved):
Thanks for the updates!  Looks good to me.

PR #616, opened 2022-01-26
Updating references from "master" to "main"

jonathanvq:
https://github.com/medic/cht-core/issues/6574

Updating references from "master" to "main" in a GitHub Action and the readme file.  The repository's default branch has been renamed to "main" already, and it was not breaking, but I considered it was ideal to have these messages and references up to date.


mrjones-plip (approved):
lgtm!

PR #618, opened 2022-01-28
Specify branch in checkout in ci, match if syntax

mrjones-plip:
per [thread on slack](https://medic.slack.com/archives/C024KTGRW/p1643350473551689)


dianabarsan (approved):
Nice! Thank for the quick reaction!

PR #623, opened 2022-02-01
Add missing full name in the user doc when creating users

jonathanbataire:
# Description
Currently when creating users with cht-conf using the documentation, the user's doc in the medic db doesn't include the user's full name. This is because `full name` is misrepresented as `name` in the tutorial. As a result the user details in the console will not have a `full name`

### DIFF
--- a/content/en/apps/tutorials/contact-and-users-2.md
+++ b/content/en/apps/tutorials/contact-and-users-2.md
@@ -72,7 +72,7 @@ Next you are going to create CHW Areas for the Health Facilities you created in 
 
 Create a CSV file named `users.csv` and add the details of the Users, CHW contacts, and CHW Areas you would like to create. Save this file in the base project directory.
 
-| username | password | roles | name | phone | contact.name | contact.phone | contact.sex | contact.age | place.type | place.name | place.parent |
+| username | password | roles | fullname | phone | contact.name | contact.phone | contact.sex | contact.age | place.type | place.name | place.parent |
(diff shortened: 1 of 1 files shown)


smbuthia:
@jonathanbataire curious if there's a code reference for this in [cht-conf](https://github.com/medic/cht-conf) (I couldn't find it). Could you link it here? Additionally, I think there are other references to column name [here](https://docs.communityhealthtoolkit.org/apps/guides/data/csv-to-docs/) and [here](https://docs.communityhealthtoolkit.org/apps/guides/data/users-bulk-load/). It would be great to update across all docs.

jonathanbataire:
> @jonathanbataire curious if there's a code reference for this in [cht-conf](https://github.com/medic/cht-conf) (I couldn't find it). Could you link it here? Additionally, I think there are other references to column name [here](https://docs.communityhealthtoolkit.org/apps/guides/data/csv-to-docs/) and [here](https://docs.communityhealthtoolkit.org/apps/guides/data/users-bulk-load/). It would be great to update across all docs.

@smbuthia you can find references to this in cht-core I'll tag devs to give some solid feedback to this
The other traces of name you have tagged seem to be in the contact(contact doc) which is okay. Here im looking at the user doc

njogz:
@smbuthia @jonathanbataire `fullname` is defined in the api [here](https://github.com/medic/cht-core/blob/master/api/src/services/users.js#L34). The `name` attribute in the database gets set to what is is provided in the username column as seen [here](https://github.com/medic/cht-core/blob/master/api/src/services/users.js#L322).

smbuthia (approved):

PR #626, opened 2022-02-04
fix order and fix order of "loadash"

mrjones-plip:


content/en/core/guides/update-dependencies.md
```diff
@@ -39,7 +39,7 @@ Then for each folder go through these steps.
 - Don't update `jquery` to 3.6.0+ as the `select2` search input looses focus on click event, this is an [open issue](https://github.com/select2/select2/issues/5993) in their repository.
 - Make sure the version of `webapp/jquery` is the same as `webapp/enketo-core/jquery`.
 - Make sure the version of `api/enketo-xslt` is the same as `webapp/enketo-core/enketo-transformer/enketo-xslt`.
-- Don't update `loadash/core` since 4.17.21 is failing, this is an [open issue](https://github.com/lodash/lodash/issues/4904) in their repository.
+- Don't update `core/lodash` since 4.17.21 is failing, this is an [open issue](https://github.com/lodash/lodash/issues/4904) in their repository.
```
latin-panda, line 42: Hi @mrjones-plip, why the change in order?
Just in case, we import lodash's core module like this:  [`const _ = require('lodash/core');`](https://github.com/medic/cht-core/blob/8f328d40b21dd95a6f24606e19c607c75e8dc41d/admin/src/js/services/contact-muted.js#L1)
I do noticed the typo 😩 `loadash` -> `lodash`

mrjones-plip: I thought we were dealing with paths here - I'm glad you're reviewing my PR! I'm still very much learning how node/npm pull all these together, so thank you double checking my work  :pray: 

So, I thought "core" meant the lodash one in core's `package.json` [here](https://github.com/medic/cht-core/blob/118c7756eec02b08e88433226f1a39293aef8209/package.json#L105). When we say "`webapp/jquery`" do we not mean the one [here](https://github.com/medic/cht-core/blob/118c7756eec02b08e88433226f1a39293aef8209/webapp/package.json#L40) in `./webapp/package.json`? 

When I checked, I didn't see a folder called `REPO-ROOT/webapp/jquery/` nor one called `REPO-ROOT/core/` so I assumed these bullet lists were referring to _directories_ that had `package.json` files with the libraries mentioned, not modules with in the libraries .

Also, while we're in here, I see  the [lodash issue mentioned](https://github.com/lodash/lodash/issues/4904) in this line is deleted?  Do we know what this means? I see [we added this line](https://github.com/medic/cht-docs/commit/60395623eb9b03a5d26c6ff8ed7088dec4439084#diff-931f03b1b67a55349374a0499ba4740948a241689e58cf5262c03c0da4443562) July of last year.

latin-panda: > So, I thought "core" meant the lodash one in core's package.json here. When we say "webapp/jquery" do we not mean the one here in ./webapp/package.json?

Oh I see, the bullets don't have consistent wording, sorry for the confusion. 
In this case we refer to lodash's core module that it's used in more than 1 place (/shared-libs, /admin, /webapp). This is a JavaScript module, these libraries use different structures, like having folders or files per each module or some might define more than one module in 1 file. However they to access them is the same, by using a path like notation `lodash/core`, so it looks like this in the code: `const _ = require('lodash/core');` 
Here `webapp/jquery` refers to the jQuery lib that cht-core's webapp is using in the `/node_module`, webapp has a specific problem because of Enketo that other places don't have (like /admin). Enketo uses jQuery internally and Webapp is using it too for other jobs outside forms, they need to match to avoid issues.
So a better why to say might be: `CHT-Core's webapp is using Enketo and jQuery library, at the same time Enketo internally uses a specific version of jQuery. Make sure webapp installs the same jQuery version than the one Enketo uses internally"  

> while we're in here, I see the [lodash issue mentioned](https://github.com/lodash/lodash/issues/4904) in this line is deleted? Do we know what this means?

😩 Not good they deleted it, they should have kept it and maybe close it with a clarifying comment. I don't know why. Maybe they fixed the issue in npm (?)

mrjones-plip: Thanks so much for the help! Really appreciate you digging into this with me.

> Perhaps a better to say it is: `CHT-Core's webapp is using Enketo and jQuery library, at the same time Enketo internally uses a specific version of jQuery. Make sure webapp installs the same jQuery version than the one Enketo uses internally`

While this is more explicit, for me (the newbie :wave: ), it's not any more actionable.  What are the literal steps I would take to ensure the two versions are installed match? Looking at the [other PR you're helping me on](https://github.com/medic/cht-core/pull/7501/files/7ef7a54c46daaa7dee654f2cd234088d316051fa#diff-006cce2fd88392a0bff4cf16091d480b6238c6bdf52c9042bf7805879c791a8fL40) (thank you!), it looks like maybe it's:

> CHT-Core's webapp is using Enketo and jQuery library, at the same time Enketo internally uses a specific version of jQuery. Make sure webapp installs the same jQuery version than the one Enketo uses internally.
> 
> Do this by checking the [jquery entry](https://github.com/medic/cht-core/blob/e068fe72750ee8f1207cf5dbc71462c3d35051f9/webapp/package.json#L40) in `./webapp/package.json`  matches `./webapp/node_modules/enketo-core/package.json`. 
> ```
> $ grep '"jquery"' ./webapp/package.json
>    "jquery": "^3.6.0",
> $  grep '"jquery"' ./webapp/node_modules/enketo-core/package.json
>    "jquery": "3.3.x",


The `webapp/node_modules/enketo-core` isn't there on one of my sample `cht-core` repos though.  Is this because I haven't run `npm ci` in it?

Also...after typing all this, why don't we just pin it? We're not going to update enketo anytime soon - so why not just change the `package.json` to have `"jquery": "3.3.1" which is what I believe enketo 4.60.6 uses, yeah? Well, ok, we will in  CHT 4.0.0 which actually is soon, so maybe this is a "why didn't we"?

per lodash issue being deleted:

> Not good they deleted it, they should have kept it and maybe close it with a clarifying comment. I don't know why. Maybe they fixed the issue in npm (?)

Maybe! since I don't know what the issue is, do you know how to check if it's resolved? I dug up this in a search cache:

> ### isEqual returns 'stack.get is undefined' errors when use...
> ...stack is a instance of Stack class, but in core version, stack is a Array, it has not a function named 'get'

mrjones-plip: Thanks for the feedback @latin-panda - sorry to not keep this moooovin :cow: 

back to you for review

latin-panda (approved):
Thank you!

PR #627, opened 2022-02-04
Add link to related OpenMRS guide

abbyad:
The guide for OpenMRS integration is the most related content page but was missing.


mrjones-plip (changes requested):
Good catch! 

I think we should actually go further though. When I read the last paragraph:

> Integrating CHT apps with OpenMRS can be achieved using the [OpenMRS REST API](https://rest.openmrs.org/) and following the guidance in the [Custom Integrations](https://docs.communityhealthtoolkit.org/apps/features/integrations/custom/) documentation.

This seems like flat out the wrong link for "Custom Integrations" and that we'd be burying the lede by having your addition as just a related link.  Instead I think we should not add your related link and change the above to read:

> Integrating CHT apps with OpenMRS can be achieved by following the guidance in the [OpenMRS integration](https://docs.communityhealthtoolkit.org/apps/features/integrations/openmrs/) documentation, which leverages the [OpenMRS REST API](https://rest.openmrs.org/).

abbyad:
Thanks, I like that shift for the last sentence. I'll update accordingly.

abbyad:
I updated the content, and left the relevant link to keep it in prominence. If we find that four related content links are too many then I would prefer to remove the RapidPro or DHIS2 links.

mrjones-plip (approved):
lgtm!

PR #631, opened 2022-02-09
Add caveats and trade-offs to hierarchy depths

mrjones-plip:


content/en/apps/tutorials/application-settings.md
```diff
@@ -99,7 +99,13 @@ In this example, the `generate_patient_id_on_people` and `death_reporting` trans
...
+You can configure hierarchies by editing the object corresponding to the `"contact_types"` key in `app_settings.json`. The following code sample represents the default hierarchy configuration. You can modify existing contact types by editing the objects within the array. 
+
+When configuring a new deployment, it's important to plan your hierarchy well.  While there are no limits in the CHT to the hierarchical depth or breadth it will support, there are some trade-offs and caveats:
+* Any time you need a hierarchical change after many documents have been created, this is [non-trivial change]({{< relref "apps/guides/updates/moving-contacts" >}}). Try and avoid this by planning ahead as best as possible.
+* If your hierarchy is too shallow, and you force contacts onto a level causing that level to excessively wide, there will be a performance loss because everyone at this level will have to sync every document. Replication depth, as outlined below, can not improve this.
```
garethbowen, line 106: This captures the trade-offs nicely, but I wonder if we can use simpler language, eg:

```suggestion
* Any time you need a hierarchical change after many documents have been created, this is [non-trivial change]({{< relref "apps/guides/updates/moving-contacts" >}}). Try and avoid this by planning ahead as best as possible.
* If your hierarchy is too shallow, and you force contacts onto a level causing that level to excessively wide, there will be a performance loss because everyone at this level will have to sync every document. Replication depth, as outlined below, can not improve this.
```
```suggestion
* While it it possible to [update the hierarchy configuration]({{< relref "apps/guides/updates/moving-contacts" >}}) after launch it can be difficult and potentially disruptive to users. Try and avoid this by planning ahead as best as possible.
* If your hierarchy is too shallow, users will download more docs than are necessary for their work which will impact the performance of the app. Taking the time to get the hierarchy configuration right makes it easy to give users access to only the docs they need.
```

garethbowen (changes requested):
Awesome! I've added what I think is simpler wording with the same meaning. Happy to iterate on this quite complicated topic...

garethbowen (approved):
Done! Nice one. Thanks for closing the loop on this.

PR #634, opened 2022-02-11
7348 - EOL dates

latin-panda:
https://github.com/medic/cht-core/issues/7348


ngaruko (approved):
Looks good. Thanks for taking on the release quickly.

PR #644, opened 2022-02-23
remove git page that was talking about not git

mrjones-plip:


content/en/contribute/docs/workflow.md
```diff
@@ -12,7 +12,7 @@ Anyone can contribute to CHT documentation by opening an issue in the [`cht-docs
 
 ### Basics
 
-* It is helpful to be comfortable with [git]({{< ref "core/guides/using-git" >}}) and [GitHub](https://lab.github.com/) to contribute to the CHT community.
+* It is helpful to be comfortable with [GitHub](https://lab.github.com/) to contribute to the CHT community.
```
craig-landry, line 15: Instead of dropping the git link, how about just pointing to something more useful?

```suggestion
* It is helpful to be comfortable with [git](https://git-scm.com/doc/ext) and [GitHub](https://lab.github.com/) to contribute to the CHT community.
```

content/en/contribute/code/workflow.md
```diff
@@ -96,7 +96,7 @@ Never force push remote. Prefer rebasing over merging as it makes for a cleaner
 
 Commit reformats and refactors separately from actual code changes to make reviewing easier.
 
-For more help with Git see: [Using Git]({{% ref "core/guides/using-git" %}}).
+For more help with Git see: [Using Git](https://lab.github.com/).
```
craig-landry, line 99: The url used is mostly about the GitHub product/tool but the context of this line is more about the git cool. Suggestion here is to point to the official git site where they provide ways to learn

```suggestion
For more help with Git see: [Using Git](https://git-scm.com/doc/ext).
```

craig-landry (changes requested):
The content being removed is oddly specific and makes me wonder if it handle a special case we have where those advanced tuning concepts are required. I don't think so though, and perhaps we've just moved on since those days.

content/en/contribute/docs/workflow.md
mrjones-plip: ah - good call!

content/en/contribute/code/workflow.md
mrjones-plip: yeah - wanted feature parity w/ the other link - but yeah `git` != GH

craig-landry (approved):

PR #646, opened 2022-02-23
Document regular upgrade advice

garethbowen:
@abbyad Change as discussed in the meeting today.


abbyad (approved):
Nice to have this included here!

PR #651, opened 2022-03-15
Add known issues for two regressions

mrjones-plip:


content/en/core/releases/3.14.0.md
```diff
@@ -9,7 +9,7 @@ toc_hide: true
 
 ## Known issues
 
-None.
+* [#7515](https://github.com/medic/cht-core/issues/7515) - Translations not working when using repeat groups with dynamic repeat count - Fixed in 3.14.1
```
garethbowen, line 12: The "fixed in 3.14.1" bit feels weird, like we're recommending people upgrade, but that might cause them to stumble on the other issue. Can we say "fixed in 3.14.2"?

garethbowen (approved):
Approved pending one suggestion inline.

content/en/core/releases/3.14.0.md
mrjones-plip: Yeah - I wasn't sure which way to go with that - literally, technically accurate or "this is where we ended up".  Fixed in 546c19f - thanks!

PR #654, opened 2022-03-18
add forum post as FAQ for couch2pg use of data in superset

mrjones-plip:


content/en/core/overview/data-flows-for-analytics.md
```diff
@@ -86,3 +86,7 @@ The [cht-core](https://github.com/medic/cht-core) is mostly data collection tool
 
 The machines running each of CouchDB and PostgreSQL instances are backed up daily.
 
+
+## Frequently Asked Questions
+
+- [Visualizing CHT data using Superset, couch2pg and Postgres?](https://forum.communityhealthtoolkit.org/t/visualizing-cht-data-using-superset-couch2pg-and-postgres/1783/2)
```
michaelkohn, line 92: I'm not actually sure what the "question" is :-)  

If we want to add more info about how materialized views work and are refreshed, maybe we should either put more details in the [data transformation](https://docs.communityhealthtoolkit.org/core/overview/data-flows-for-analytics/#2-data-transformation) section or in the couch2pg documentation itself?

mrjones-plip: > I'm not actually sure what the "question" is :-)

Fair point! If we want to go with an FAQ, I guess it'd be:

```suggestion
- [How do you use materialized views in Postgres to visualize CHT data in Superset?](https://forum.communityhealthtoolkit.org/t/visualizing-cht-data-using-superset-couch2pg-and-postgres/1783/2)
```

Otherwise, I think the forum post is a nice, stand alone FAQ - but that actually doesn't scale to generic content for the [data transformation](https://docs.communityhealthtoolkit.org/core/overview/data-flows-for-analytics/#2-data-transformation) section.  I think that would take a lot more work and this felt like a quick win.

The  [couch2pg documentation](https://github.com/medic/cht-couch2pg) is entirely lacking when it comes to how to use the data. I think adding info in the docs makes the most sense.

michaelkohn: To me the data transformation section actually already covers the high level architect-y stuff re: creating views, materialized views, and them being refreshed every time it runs.... and the technical details (requirements for indexes and clarity around when things are refreshed) should live in the couch2pg documentation.  Of course it sounds like couch2pg is going to be retired in the not too distant future and I'm not sure how much of what is in the docs site today will still be relevant for cht-pipeline.

Anyway... adding the link certainly does no harm! Kind of gets me thinking about just putting a link like this: https://forum.communityhealthtoolkit.org/search?q=couch2pg

mrjones-plip: Ah - thanks for the pushback! In my mind, I didn't see any SQL on the docs page, saw some on the forum post, thought, "more SQL, more better!" 

However, the reason I picked you for the review is your intimacy with the topic - it's way more than mine!  Happy to leave things as is.

garethbowen:
@mrjones-plip Can you follow up on whether this PR is needed or not?

mrjones-plip:
been fine for so long w/o - closing for another day.

mrjones-plip:
thx again @michaelkohn ! 

thx for nudging to cull @garethbowen !

PR #658, opened 2022-03-24
Add glossary page

garethbowen:
Review please @smbuthia - glossary page as discussed. Mostly just copied and pasted off Notion. There's plenty more to add but I'm hoping everyone helps out to contribute now there's a place for this content to exist.

@mrjones-plip I tagged you because I'm not 100% happy with how this functions. Because some glossary items have a paragraph explaining them and some don't it's ended up hard to scan over. Any ideas from a Hugo expert would be appreciated!

### DIFF
--- a/content/en/glossary.md
+++ b/content/en/glossary.md
@@ -1,0 +1,216 @@
+---
+title: "Glossary and Definitions"
+linkTitle: "Glossary"
+menu:
+weight: 10
+---
+
+##### Accredited Social Health Activist (ASHA)
+
+##### Antenatal Care (ANC)
+
+##### Antiretrovirals (ARV)
+
+##### Artificial intelligence (AI)
+
+##### Backend
+
+The parts of the software that make the front end happen. These are all hidden from the user.
+
+##### Center for Disease Control and Prevention (US) (CDC)
+
+##### Childhood Immunizations (IMM)
+
+##### Community Based Organization (CBO)
+
+##### Community Based Volunteers (CBV)
+
+##### Community Health Assistants (CHA)
+
+##### Community Health Extension Worker (CHEW)
+
+##### Community Health Promoters (CHP)
+
+##### Community Health Toolkit (CHT)
+
+##### Community Health Volunteer (CHV)
+
+##### Community Health Worker (CHW)
+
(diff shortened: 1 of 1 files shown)


mrjones-plip (approved):
tl;dr - This follows a good KISS approach - ship it!

----
In order of increasing complexity, here's some of "tl" thoughts that you should "dr" ;) Archiving these for later should we want to make it more feature rich or add more formatting:
* Increase the header level from 5 to 6 (`#####` -> `######`). This toned down the "header-ness" but was subjective and a nitpick (btw - [there is no h7](https://www.w3schools.com/tags/tag_hn.asp)!)
* Change the list [to use `dl` `dt` via this syntax](https://discourse.gohugo.io/t/markdown-to-formatted-faq/26799):
   ```
   Backend
   : The parts of the software that make the front end happen. 
   ```
   Results in:
   image
   I love the format and resulting HTML structure, I didn't like that we'd have add awkward empty definitions:
   ```
   Accredited Social Health Activist (ASHA)
   : 
   ```
* Maybe use [footnotes](https://www.markdownguide.org/extended-syntax/#footnotes)? This doesn't seem to scale.
* Use an [A-Z navigation](https://discourse.gohugo.io/t/how-to-make-like-glossary-page/14182)? Maybe at a later time.
* Use [tooltips for inline glossary](https://kubernetes.io/docs/contribute/style/hugo-shortcodes/#glossary)? Time intensive! Not KISS!

garethbowen:
> Increase the header level from 5 to 6.

Done!

> Change the list to use dl dt

Initially that looked really nice and the rendering was tight, but you're right it's not great for the items we don't have descriptions for. Also, using headers means you can link to a specific definition but using dl dt you can't. So I've stuck with the current header system for now.

> Maybe use footnotes

I decided not to do this because I want the content inline rather than having to find the term you want then click a link to find out what it means.

> Use an A-Z navigation

I found the same blog post you did but didn't dig deep enough to find how kube is doing it. However I've added h2s for each letter which I think makes it look better, but also gives you links down the RHS to the letter you're interested in.

> Use tooltips for inline glossary

Yeah that's beautiful. I'm seriously tempted, but I'll leave the for round 2 when I have another couple of hours to spare... New issue raised: https://github.com/medic/cht-docs/issues/659

smbuthia (approved):

Issue #666, opened 2022-04-01
Link not displaying correctly in 'Building A Simple Task'


abbyad:
Page: /apps/tutorials/tasks-1/
Version: 2021-11-29 11:05:14 -0800 -0800
---
The link code is shown instead of the link text itself, likely due to an extra whitespace between the `{{` and `<`.

<img width="477" alt="image" src="https://user-images.githubusercontent.com/789512/161286041-73a087bf-ee7e-4cc5-b842-a96fa24f2267.png">

PR #667, opened 2022-04-01
Fixes link

abbyad:
For https://github.com/medic/cht-docs/issues/666

@mrjones-plip should we add checks for `{{ <` as part of the build?


abbyad:
Link is still broken... will fix now

mrjones-plip (approved):
per this PR, lgtm!

per checking for `{{<` and not `{{ <` ...maybe? I guess a you wouldn't even need a regex for it because  `]({{ <` is a literal string that you could catch with a regex-less `grep` or something.

Maybe if it happens again we should code for it, but don't sweat it until then?  Either way, another ticket and not in this PR ;)

Issue #648, opened 2022-02-28
Document what a PWA is, and when and how to use it


garethbowen:
Not sure exactly where this should go and might have to wait until the [PWA regression](https://github.com/medic/cht-core/issues/7362) is fixed (3.15?).

PR #672, opened 2022-04-07
Document installation of the CHT as a PWA

garethbowen:
#648

### DIFF
--- a/content/en/core/overview/pwa.md
+++ b/content/en/core/overview/pwa.md
@@ -6,4 +6,30 @@ description: >
   What does it mean that the CHT Core web application is a Progressive Web App.
 ---
 
-## Introduction
+## What is a Progressive Web App (PWA)?
+
+A PWA is a web application that can be used like a website in the browser, but the user can choose to "install" it. This means a shortcut is added to the home screen of the device, and when the application is run it doesn't have the usual browser address bar and tabs so it looks like a regular application.
+
+The CHT Core webapp has been developed to be a PWA to give users more choice about how applications are installed.
+
+## Benefits of PWA installation
+
+Using a PWA avoids the need to build an Android application and publish it to the Play Store which can be cumbersome and slow. This also means there's no need for the user to install or update the Android application.
+
+In addition you can install PWAs on a wider range of devices, for example, desktop computers.
+
+Deployments can also use a mixture of PWA and Android applications.
+
+## Benefits of Android applications
+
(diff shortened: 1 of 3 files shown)


content/en/core/overview/pwa.md
```diff
@@ -0,0 +1,35 @@
+---
+title: "Installation as a Progressive Web App"
+linkTitle: "Progressive Web App"
+weight: 2
+description: >
+  What does it mean that the CHT Core web application is a Progressive Web App.
```
jkuester, line 6:
```suggestion
  What does it mean that the CHT Core web application is a Progressive Web App?
```

Or `What it means that the CHT Core web application is a Progressive Web App.`  It just seems to word this as a question without the question mark.

content/en/core/overview/pwa.md
```diff
@@ -0,0 +1,35 @@
...
+
+## What is a Progressive Web App (PWA)?
+
+A PWA is a web application that can be used like a website in the browser, but the user can choose to "install" it. This means a shortcut is added to the home screen of the device, and when the application is run it doesn't have the usual browser address bar and tabs so it looks like a regular application.
+
+The CHT Core webapp has been developed to be a PWA to give users more choice about how applications are installed.
+
+## Benefits of PWA installation
+
+Using a PWA avoids the need to build an Android application and publish it to the Play Store which can be cumbersome and slow. This also means there's no need for the user to install or update the Android application.
```
jkuester, line 17: > This also means there's no need for the user to install or update the Android application.

It might be worth noting how the PWA is "updated".  Something like "The PWA is updated automatically when syncing with the CHT server."

(I know this is not really the same "update" as the Android application, but I think a reader who is not super familiar with PWA's still might wonder how things stay up do date if there is no app to update...)

jkuester (approved):
Good docs! :books: 

This is slightly unrelated, but I was wondering if we should maybe, in the Publishing > F-Droid section we should add a link to the [blog post](https://medic.org/stories/using-f-droid-for-app-distribution-a-product-experiment/) where there is some more information?

content/en/core/overview/pwa.md
garethbowen: Changed to not be a question.

content/en/core/overview/pwa.md
garethbowen: Good call. Added.

garethbowen:
Good point about f-droid! The page says "we're still evaluating" which is no longer true. Updated to link to the blog post.

PR #673, opened 2022-04-07
Adding some notes regarding permissions in mobile

latin-panda:


ngaruko (approved):
Thanks for adding these notes @latin-panda .

PR #675, opened 2022-04-09
Training Instance title and link updates

abbyad:
Fixed to match updated content and better align with other titles in section. Also fixes a broken link.


abbyad:
Pushed through without additional review since it was a new page with broken image link.

PR #677, opened 2022-04-12
Update medic conf to cht-conf

smbuthia:
The [Contact and User Management - Part 1](https://docs.communityhealthtoolkit.org/apps/tutorials/contact-and-users-1/) had a reference to 'medic conf' instead of the updated 'cht-conf'.


abbyad (approved):
Nice catch!

PR #678, opened 2022-04-12
Document training app feature for android flavors

abbyad:
For https://github.com/medic/cht-android/issues/258


ngaruko:
Tested with `unbranded` but switching on and off the `IS_TRAINING_APP` flag
```
unbranded {
      // we will not create project-specific src directories
      // for `unbranded` - it will use the defaults in
      // src/main
      dimension = 'brand'
      buildConfigField "boolean", "IS_TRAINING_APP", 'true'
    }
```
When the flag is  set to `true`, a toast shows at the beginning of the app 
Screen Shot 2022-04-20 at 1 00 11 PM

and an orange contour around all screens
Screen Shot 2022-04-20 at 1 01 15 PM
 And none when the flag is set to 'false'
image

Good to merge @abbyad .

PR #684, opened 2022-04-24
Add paragraph about risk of spamming task docs

garethbowen:
medic/cht-core#6564


garethbowen:
Review please @dianabarsan . This explains the end date change in cht-core. Wasn't quite sure if this was the right place to put it - suggestions welcome!

dianabarsan (approved):
Looks great. I think this is the right place to add this information. 
If task generation proves problematic again, maybe we could consider adding a section [under Performance](https://docs.communityhealthtoolkit.org/apps/guides/performance/).

PR #686, opened 2022-04-27
Add nightly link checks 

mrjones-plip:
per #366


.github/scripts/muffet.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# This script is used in Github Actions to do link checking on commits to branches and main, so
+# This script is used in Github Actions to do link checking on commits to branches and master, so
```
garethbowen, line 3: `main`, right?

mrjones-plip: d'oh! fixed in  367f835 and 39f0f5d

mrjones-plip:
@garethbowen - big bummer is we can't [test it 'til it's in `main` :shrug](https://stackoverflow.com/a/63437012): 

>  Scheduled workflows run on the latest commit on the default or base branch.

For initial launch, we'll just post to Slack.  First couple of runs will spam the channel, but it should quiet down after we tend to the 404ed  links. I thought this was easier than creating a ticket and then checking if the ticket existed and not create dupe etc.

garethbowen (approved):
Awesome work! Go for it.

PR #688, opened 2022-05-02
Update build to use muffet tag name

garethbowen:


garethbowen:
@mrjones-plip The build has been failing with two errors. The first one is failing to find the muffet version. I _think_ it should be the name of the tag (which starts with a `v`). The second error is sending the update to Slack, but we can fix that one later if this change works.

mrjones-plip (approved):
thanks for the PR to fix my PR !

I had a note to myself "after we're done with feature release, look into link checker" and was going to schedule an hour of real time debugging w/ you b/c each test needs to be merged to master.  Though, let's hope this works instead.

PR #695, opened 2022-05-17
Update condition-cards.md

binokaryg:
Replace `>==` by `>=` because `>==` is not available in JS.


garethbowen (approved):
Thanks @binokaryg !

Issue #692, opened 2022-05-05
Broken Android Hosting Guide link in search results


abbyad:
The [Android Hosting Guide](https://docs.communityhealthtoolkit.org/apps/guides/hosting/android-app/) comes up in search for "MDM" but no longer exists.

We should alias to the appropriate location, likely the [Android Publishing Guide](https://docs.communityhealthtoolkit.org/core/guides/android/publishing/#mobile-device-management). This is related to https://github.com/medic/cht-docs/issues/691.

### Search Result
![image](https://user-images.githubusercontent.com/789512/166899175-95224d64-d1a6-4b9c-962e-d7c77c0c3fed.png)

kitsao:
Should this be closed @mrjones-plip?

PR #697, opened 2022-05-18
fix 404 page so it redirects 

mrjones-plip:
per #692


abbyad (approved):
Looks good! 

For #692 not 693, right?

mrjones-plip:
D'oh! Yes, #692 - thanks!

PR #699, opened 2022-05-25
Added default port and creds info.

PrjShrestha:


abbyad (changes requested):
Great to have this specified! In terms of context, what problem is this trying update to solve? That will help figure out where this fits best within the page.

Also, to be consistent in our use of CHT terms, such as `CouchDB`, you can check the [word list](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#community-health-toolkit-word-list) in the style guide. One that should be there but isn't is `cht-core`... which should always show as `cht-core`, not `Cht-core`, even at the beginning of a sentence. It may look weird like that though, so I'd rephrase to not start with `cht-core`, or use the full name, like `CHT Core Framework`.

PrjShrestha:
@abbyad Thank you for your suggestions. When I followed the tutorial and deployed the app I had no idea where it was deployed (which port) and which creds to use. Hence I would like them to be mentioned in the doc. If these are mentioned in another doc I would like them to be tagged here.

PrjShrestha:
@abbyad Please check now.

content/en/core/guides/using-windows.md
```diff
@@ -110,6 +110,10 @@ $: cd ~/medic/cht-core
 $: code .
 ```
 
-## Problems?
+## Default port and credentials for cht-core web
+The default launch port for cht-core is 5988, which can be changed in cht-core/api/server.js.
```
garethbowen, line 114: I think the recommended approach would be to provide an environment variable rather than editing the code - that way the change won't be lost if you upgrade.

```suggestion
The default launch port for cht-core is 5988, which can be changed by providing the environment variable at runtime `API_PORT`, for example: `API_PORT=6000 node server.js`.
```

garethbowen:
@PrjShrestha Is this ready to merge?

PR #704, opened 2022-06-03
Add documentation on input data available to forms

### LINKED ISSUE #297: Documentation on data that can be accessed and used in forms would be helpful.
Forms have access to data that comes from the app. I do not believe we have any docs on what is actually available to forms. A few examples are shown below. I'm not sure of all the data that is passed to a form so this list is not exhaustive. 

string | contact_id 
string | facility_id
string | name

Lots of forms use PLACE_TYPE to determine the contact type.

We have meta data available as well for things like start and end of form

jkuester:
Adds a new page to the Reference Guide that details what input data is available to app/contact forms.

Closes #297

### DIFF
--- a/content/en/apps/guides/forms/form-inputs.md
+++ b/content/en/apps/guides/forms/form-inputs.md
@@ -1,0 +1,141 @@
+---
+title: "Input data available in forms"
+linkTitle: "Form Inputs"
+weight: 
+description: >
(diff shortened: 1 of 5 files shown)


yrimal (approved):
Well explained and put in ! Thanks for drafting this Josh.

PR #715, opened 2022-06-22
fix partners conf filename reference

elvisdorkenoo:
Hello, I think for the partners logos config, the file that has to be created or edited is `partners.json` instead of `branding.json`


latin-panda:
@nsitaula can you please verify this PR? 🙏 Thanks!

nsitaula (approved):

elvisdorkenoo:
cc @jonathanvq

Issue #718, opened 2022-06-28
Improve '.permissions'


abbyad:
Page: /apps/reference/app-settings/user-permissions/
Version: 2022-05-03 05:13:19  0300  0300
---

From @craig-landry:
> There was some confusion on a project with the `can_edit` permission since users can still create reports on devices, but they just don't sync. That is a risk since it means reports are getting created and never synced and may result in data loss.

abbyad:
Although the docs could be improved, it seems as though it would help to resolve what looks like the [underlying defect](https://github.com/medic/cht-core/issues/6215), that the permission controls the sync only and doesn't prevent the user from taking the edit action within the app.

garethbowen:
I propose we fix the underlying defect, and also this related issue https://github.com/medic/cht-core/issues/6505 and update the docs as part of that issue.

abbyad:
Great, let's do that. I'll let you schedule those issues as needed. It will be some time before any changes make it to all deployments, so I'll open a PR with small note in the docs in the meantime.

garethbowen:
Merged.

PR #719, opened 2022-06-29
Clarify use of can_edit

abbyad:
For #718


garethbowen (changes requested):
Thanks. A suggestion inline. The main change is highlighting that it's creating as well as editing docs that's permitted.

abbyad:
Thanks for the comment and suggestion. It's ready for review again.

garethbowen (approved):
Thanks!

PR #717, opened 2022-06-22
Updates to MDM section

abbyad:
For #716

Issue #716, opened 2022-06-22
Improve 'Securing Android Devices'


abbyad:
Page: /apps/guides/security/securing-android/
Version: 2020-06-05 21:05:09 -0400 -0400
---

Revisiting this page after a [recent forum conversation](https://forum.communityhealthtoolkit.org/t/enhancing-security-of-cht-apps-through-two-factor-authentication-2fa/1953?u=marc), the MDM section felt a bit bare and outdated. Creating this issue to connect an upcoming PR.

garethbowen:
Merged.

PR #720, opened 2022-06-29
716 consolidation of MDM changes

abbyad:
For #716, in response to https://github.com/medic/cht-docs/pull/717#discussion_r906870287


garethbowen (approved):
Nice one, thanks!

Do you think it's worth listing out a bunch of MDMs we know about or is that overkill?

PR #724, opened 2022-07-04
Feature development guidelines

smbuthia:
This aims to address requests to make the cht-core contribution process clearer, and outline steps through which developers can build and test new features.

### DIFF
--- a/content/en/contribute/code/workflow.md
+++ b/content/en/contribute/code/workflow.md
@@ -8,13 +8,23 @@ description: >
 
 ## Code
 
-### Writing
+### How to submit changes
+
+#### Step 1: Fork
+
+#### Step 2: Branch
+
+#### Step 3: Write and Test Code
 
 Where possible, follow our [coding style guide]({{< ref "contribute/code/style-guide" >}}).
 
-Aim for self-documenting code. Where code cannot be made self-documenting add commenting. Usually comments are useful when they explain why some code exists, and should not be explaining what some code is doing. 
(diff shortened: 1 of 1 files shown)


garethbowen:
@smbuthia Initially this document was intended to be for Medic teammates. The steps for external contributors are different in a few places (eg: forking). Is this change intended to make it useful fro both, or just external?

Secondly I don't think we need to write instructions for git commands like forking, branching, etc as these are covered by the git documentation.

One thing I think we do need to cover is when is it appropriate to fork. If a partner is forking they need to know about potential downsides of this approach.

smbuthia:
Thanks, Gareth. 

> @smbuthia Initially this document was intended to be for Medic teammates. The steps for external contributors are different in a few places (eg: forking). Is this change intended to make it useful for both, or just external?

This is an attempt at making the process a little clearer for community members since this document is shared with the community. We probably don't want 2 versions of the same doc floating around so perhaps we could specify where (and why) the steps are different for the Medic developers/stewards compared to community contributors.

> Secondly I don't think we need to write instructions for git commands like forking, branching, etc as these are covered by the git documentation.

I agree that we don't need to write instructions for these. The titles are to provide clarity for each step. For example, "once you fork, you can deploy your own version of the Core Framework and will only need to create a PR if you would like to contribute the feature back to the supported Core Framework" or "Medic does not provide support for forked/unmerged version of the Core Framework". 

> One thing I think we do need to cover is when is it appropriate to fork. If a partner is forking they need to know about potential downsides of this approach.

Yes! This could be one of the clarifications under the forking section.

smbuthia:
Closing this to review the approach.

PR #730, opened 2022-07-21
Add cht docker helper content and images. toc to 4 levels vs three 

mrjones-plip:
PR does:
* Migrates content in [from cht-core](https://github.com/medic/cht-core/tree/master/scripts/docker-helper)
* Changes TOC depth from 3  to 4 instead. On long pages, like the this one, it really aids navigation

Once merged and pushed live, we'll deprecate [the source](https://github.com/medic/cht-core/tree/master/scripts/docker-helper) and add a "use docs page!" link instead.

Per #729

PR #731, opened 2022-07-22
Upgrade Hugo and Docsy to latest versions

jkuester:
Upgrades the version of Hugo used in the GitHub actions to deploy the site from `0.71.1` to `0.101.0` (latest).  

Pins the version of Docsy to the latest `0.4.0` and switches from referencing Docsy as a git submodule to pulling in Docsy [as a Hugo Module](https://www.docsy.dev/docs/updating/convert-site-to-module/) which is now the [recommended way](https://www.docsy.dev/docs/get-started/#installation-options) to include Docsy (though the submodule approach is still supported).  The Hugo module approach has the advantaged of eliminating the extra step of pulling down the submodules when building the site locally. Also, it makes it easy to pin to a specific version of Docsy to allow for more controlled upgrades.

These upgrades were largely passive. However, I have updated several templates in `layouts` to reflect the latest changes from Docsy (while maintaining the existing customizations we already had). Additionally, there were several layout templates that were not needed any more (deleting them had no noticeable effect on the site).  This is thanks to newer versions of Hugo having the ability to do [target specific cascading front matter](https://gohugo.io/content-management/front-matter/#target-specific-pages) which enables a more simple implementation of [alternative site structures in Docsy](https://www.docsy.dev/docs/adding-content/content/#alternative-site-structure) (like ours). I have removed these unnecessary files.  

Ultimately, other than some _very_ subtle differences in the the layout/display of the site, there is only two really noticeable differences that I have identified:

1. I have switched to leveraging the `sidebar_menu_foldable` option that is now available in Docsy which provides animated chevrons in the sidebar tree (instead of our existing roll-your-own implementation). The big difference, though, is that the Docsy chevron is placed on the left side instead of the right:

    | Before | After | 
    | --- | --- |
    | image |image |

2. The formatting has changed for the "Community Health Toolkit" title in the navbar (this would be something easy to fix, but it did not look like we had customized that title to begin with, so I was not sure if we wanted to just stick with the Docsy default here...):

    | Before | After | 
    | --- | --- |
    | image | image |


mrjones-plip:
First pass on this looks good!  Love that we're removing almost as much as we're adding \o/  Of course totally awesome to be current!

On the technical front, I think we should call out how to handle one time conversion from `main` today to what it will be tomorrow. Mainly that git can't `rmdir` (see below).  I'll see if there's any other gotchas technically - stay tuned!

```
➜  cht-docs git:(csv-bulk-user-upload) gco uplift_hugo                                                                                                                                             
warning: unable to rmdir 'themes/docsy': Directory not empty                                                                                                                                       
Branch 'uplift_hugo' set up to track remote branch 'uplift_hugo' from 'origin'.                                                                                                                    
Switched to a new branch 'uplift_hugo'                                                                                                                                                             
➜  cht-docs git:(uplift_hugo) ✗ git status                                                                                                                                                         
On branch uplift_hugo                                                                                                                                                                              
Your branch is up to date with 'origin/uplift_hugo'.                                                                                                                                               

Untracked files:                                
  (use "git add <file>..." to include in what will be committed)                                 
        themes/                                 

nothing added to commit but untracked files present (use "git add" to track)  
```

From a design front, let's wait for @n-orlowski to get back to weigh in. The title is an easy fix should we want to `text-transform: uppercase;`. The chevrons in the left nav are not only aligned left instead of right, but they're pretty heavily animated, which is also a change.  

@n-orlowski  - here's a video I took of the animations after the upgrade:

https://user-images.githubusercontent.com/8253488/181162223-46c77867-f95f-4928-acd4-93e2603b8e15.mp4

jkuester:
> I think we should call out how to handle one time conversion from main today to what it will be tomorrow

Ah, good catch @mrjones-plip!  I had tested that a fresh clone of `cht-docs` worked as expected, but I did not think to try the "normal" upgrade path of an existing local clone.  I will also do some thinking here to see if there is a better way we could handle this.

mrjones-plip (approved):
Looking again at the left hand nav, I really like that it succeeds where we fail at today:
* does not show chevrons unless there are sub pages 
* does not reload the page while you browse in and out of the chevrons - makes the site feel much faster! 
* does not reload the page on mobile when exploring hamburger nav menu - even moaarr faster!

I rescind my concern to wait on Nicole's return and think we should just merge it and then we can open up a new ticket if we want to fix title or nav animation/chevron alignment.

that said, we need to re-enable deeper TOC levels on the right side toc. top is production, bottom is this branch:

missing toc

This [was introduced very recently](https://github.com/medic/cht-docs/pull/730/files#diff-28043ff911f28a5cb5742f7638363546311225a63eabc365af5356c70d4deb77R72-R73) and should just be the corresponding `.toml` change in the new hugo.

Approving the PR assuming right side TOC is a quick fix and the need to add a shout out to address `unable to rmdir 'themes/docsy'` on upgrading an existing install.

mrjones-plip:
UPDATE! I see this branch was cut off of `main` before the changes got merged to do the deeper TOC on the right. We can merge `main` to this branch to see it, or else it will just naturally resolve itself when we merge this PR. Still OK to merge!

jkuester:
@mrjones-plip I did some checking and it seems like this issue of the submodule's folder remaining around after the the submodule has been removed [is a known bug](https://git-scm.com/docs/git-rm#_bugs).  I have not found any way of easily solving this.  

Having the `themes` folder around should not actually affect functionality of the side build (since I think Hugo will defer to the `config.toml` settings when resolving the theme). My main concern is that someone will accidentally commit the contents of `themes`.  What do you think about adding `themes/` to the .gitignore?

mrjones-plip:
> What do you think about adding `themes/` to the `.gitignore`?

Yeah - let's give it a whirl! It would be nice to avoid that first error and the not be in this state (although `hugo serve` worked just fine):

```➜  cht-docs git:(uplift_hugo) ✗ git status
On branch uplift_hugo
Your branch is up to date with 'origin/uplift_hugo'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        themes/

nothing added to commit but untracked files present (use "git add" to track)
```

If it's too destructive/wonky we can review recent committers to this repo and just reach out to them directly with steps to migrate to new repo architecture.

mrjones-plip:
gah - it didn't do quite what we wanted - we still get an error when switching to the new branch from the old `main`. However, there's no more files in state of `Untracked files` for `./themes/` and there's no chance of an accidental commit of that dir the future, so it's a net positive. Ship it!! 

```
➜  MedicMobile git clone https://github.com/medic/cht-docs dleteme-cht-docs
Cloning into 'dleteme-cht-docs'...
remote: Enumerating objects: 12112, done.
remote: Counting objects: 100% (561/561), done.
remote: Compressing objects: 100% (327/327), done.
remote: Total 12112 (delta 286), reused 429 (delta 181), pack-reused 11551
Receiving objects: 100% (12112/12112), 76.89 MiB | 21.51 MiB/s, done.
Resolving deltas: 100% (7081/7081), done.
➜  MedicMobile cd dleteme-cht-docs 

➜  dleteme-cht-docs git:(main) git submodule update --init --recursive
Submodule 'themes/docsy' (https://github.com/google/docsy.git) registered for path 'themes/docsy'
Cloning into '/home/mrjones/Documents/MedicMobile/dleteme-cht-docs/themes/docsy'...
Submodule path 'themes/docsy': checked out '053672c24cf51417c300f8ea5d32db545340746d'
Submodule 'assets/vendor/Font-Awesome' (https://github.com/FortAwesome/Font-Awesome.git) registered for path 'themes/docsy/assets/vendor/Font-Awesome'
Submodule 'assets/vendor/bootstrap' (https://github.com/twbs/bootstrap.git) registered for path 'themes/docsy/assets/vendor/bootstrap'
Cloning into '/home/mrjones/Documents/MedicMobile/dleteme-cht-docs/themes/docsy/assets/vendor/Font-Awesome'...
Cloning into '/home/mrjones/Documents/MedicMobile/dleteme-cht-docs/themes/docsy/assets/vendor/bootstrap'...
Submodule path 'themes/docsy/assets/vendor/Font-Awesome': checked out '538a5bc4cf7989821d052e3ff919e2bead07021c'
Submodule path 'themes/docsy/assets/vendor/bootstrap': checked out '3b558734382ce58b51e5fc676453bfd53bba9201'

➜  dleteme-cht-docs git:(main) gco uplift_hugo                            
warning: unable to rmdir 'themes/docsy': Directory not empty
Branch 'uplift_hugo' set up to track remote branch 'uplift_hugo' from 'origin'.
Switched to a new branch 'uplift_hugo'

➜  dleteme-cht-docs git:(uplift_hugo) cat .gitignore 
public/
resources/
node_modules/
# Ignore old theme submodule folder
themes/
tech-doc-hugo
.DS_Store
.idea
.hugo_build.lock
```

jkuester:
> we still get an error when switching to the new branch from the old main.

Right. IMHO this is actually a "feature" in that it serves as a reminder that the folder can be deleted.  I will still send a message out so folks are aware that they can safely just manually remove that folder.

PR #739, opened 2022-07-28
Fix search page

jkuester:
The [Hugo/Docsy uplift](https://github.com/medic/cht-docs/pull/731) accidentally broke the search page completely.  :cry: 

Basically, the cascading front-matter from the base `_index.md` file caused the `search.md` file to not get treated as a search page. I have made the necessary corrections to the front-matter to fix this. 

However, I also noticed that the search page no longer contained the sidebars. After some investigating I was able to conclude that this was caused by the unintentional deletion of a couple of the default layout files during the refactor.  (Thanks to a a massive oversight on my part I did not realize they were necessary for the proper layout of the Search page.)  I have restored these pages and now the search page should behave as expected!

### DIFF
--- a/content/en/_index.md
+++ b/content/en/_index.md
@@ -5,17 +5,13 @@ identifier: "docs"
 weight: 1
 cascade:
 - _target:
-  path: "/**"
+    path: "/**"
   kind: "page"
   type: "docs"
 - _target:
-  path: "/**"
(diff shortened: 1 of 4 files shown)


garethbowen (approved):
I'm not really sure what I'm looking at, but I can't see anything obviously wrong...

PR #745, opened 2022-08-03
Add release notes for 3.16.0

jkuester:
There are no breaking changes and I do not think there is any special upgrade steps that are needed.


garethbowen (approved):
Love it!

mrjones-plip (approved):

PR #746, opened 2022-08-08
7696 - reports filter redesign

latin-panda:
## Description

This PR updates documentation related to the filter redesign:
- Details about new permission to enable old filter design
- Replaced old screenshots 

[Ticket](https://github.com/medic/cht-core/issues/7653)
[CHT Core PR](https://github.com/medic/cht-core/pull/7696)


content/en/apps/features/reports/_index.md
```diff
@@ -30,17 +30,27 @@ Apps built with the Core Framework have a “review” feature that allows manag
...
+
+
 Using the search box, you can search for reports by patient name, phone number, ID number and more. To reset the filters or the search and view  the full list of forms, click on the reset icon on the right side of the toolbar.
 
+{{< figure src="reports-search-reset.png" link="reports-search-reset.png" class="left col-lg-12" >}}
+
+
+#### How to display the old filter and search?
+
+To display the old filter and search enable the [permission]({{< ref "apps/reference/app-settings/user-permissions" >}}) `can_view_old_reports_filter` to the user's role.
```
michaelkohn, line 50: I think instead of this section, we can add a "Note" (like in [aggregate targets](https://docs.communityhealthtoolkit.org/apps/features/supervision/#chw-aggregate-targets) ) that says something like:

> NOTE
A new user experience for Filter and Search was introduced in v3.XX.  The previous version can be re-enabled for users by adding the [permission]({{< ref "apps/reference/app-settings/user-permissions" >}}) `can_view_old_reports_filter` to the user's role; however, the old version should be considered deprecated and will be completely removed in a future release.

content/en/apps/reference/app-settings/user-permissions.md
```diff
@@ -64,3 +64,4 @@ Each permission is defined as an array of user role identifiers that have the pe
 | `can_view_unallocated_data_records` | Allows viewing reports that have no associated contact |
 | `can_view_users` | Allows viewing all user accounts |
 | `can_write_wealth_quintiles` | Allows updating contacts with wealth quintile information |
+| `can_view_old_reports_filter` | Allows users to see the old filter in Reports Tab. However the Admin user will always see the new redesigned filter. |
```
michaelkohn, line 67: Perhaps leave another note here that it is deprecated and will be removed in a future release?

michaelkohn (commented):
Thanks!  I made a few comments on the text.  For the images, I like the look of the mobile device framing, but it is inconsistent with the rest of the documentation site (@n-orlowski curious your opinion on this). Also, the size of the new images are quite different from the old... perhaps Hugo just deals with it correctly and resizes them, but I can't really tell from this GH view.

content/en/apps/features/reports/_index.md
latin-panda: I like it! This is done and I've removed the image about old filters.

content/en/apps/reference/app-settings/user-permissions.md
latin-panda: Good catch! Done

latin-panda:
Thanks for the feedback @michaelkohn! It's ready for review again!

> Also, the size of the new images are quite different from the old... perhaps Hugo just deals with it correctly and resizes them, but I can't really tell from this GH view

Yes, Hugo handles it. To see the changes it's better [to run the project](https://github.com/medic/cht-docs#installing-hugo). 
Basically install hugo based on your machine, usually it's 1 command thing. Then install the project `npm install` and finally start the doc site `hugo server`. After that you should be able to see the docs site, it automatically refreshes after every change.
https://user-images.githubusercontent.com/66472237/183847338-a54c6f30-8aa4-434d-a24e-82e25ddfaea1.mov

n-orlowski:
[No framing on screens pls!](https://docs.communityhealthtoolkit.org/apps/features/contacts/) Helps for general consistency (folks not having to find the same phone asset) and makes the screen content a little larger. Thanks for checking 🙂

latin-panda:
@michaelkohn @n-orlowski I've updated the images, please have a look again!
Video:
https://user-images.githubusercontent.com/66472237/184068452-7a2d140c-20e1-4775-bffd-a5750e4e2999.mov

michaelkohn (approved):

michaelkohn:
Thanks for the video, it was really helpful! But I didn't notice it until after I approved :-/ and with the video, I can see how the images show up.    One small tweak would be in the Filter & Search section.  I'd recommend we format it like how it looks on the top of this page where it has the text and then below it has the mobile view and desktop view in the same "row".

latin-panda:
@michaelkohn Thanks I've pushed the change, please have a look:
https://user-images.githubusercontent.com/66472237/184295848-cf31c403-255c-4bb0-8380-c5a52e555704.mov

Is it okay?

michaelkohn:
Yep, LGTM thanks @latin-panda !

Issue #747, opened 2022-08-11
After hugo upgrade, certain content isn't rendering correctly


mrjones-plip:
We recently [upgrade docsy and hugo](https://github.com/medic/cht-docs/pull/731) which is causing items like the "Maternal and Newborn Health Reference App"  page in the [pregnancy workflow](https://docs.communityhealthtoolkit.org/apps/examples/anc/#pregnancy-workflow) to show raw HTML instead of rendered HTML:

![image](https://user-images.githubusercontent.com/8253488/184226117-a5a6686e-6454-4b64-981f-7f301cc1d992.png)

mrjones-plip:
It looks like our  [/layouts/shortcodes/workflow-table.html](https://github.com/medic/cht-docs/tree/main/layouts/shortcodes/workflow-table.html) file isn't rendering correctly.  This was added back in [2020 here](https://github.com/medic/cht-docs/pull/239/files).

PR #749, opened 2022-08-12
Update ANC Reference App example to fix workflow tables

jkuester:
Okay, so this turned out to be a simple, but weird, problem. Ultimately I think this is an order of operations problem and it seems like somewhere along the way Hugo changed which order it was formatting things in and this broke stuff.

I was able to get everything to render properly simply by _un-indenting_ the `workflow-table` shortcode lines in `anc.md`.  For some reason, with everything just not indented, it seemed to get the order of ops correct.  Ultimately, however, this did not seem like the proper fix (since Hugo actually has [an example of nested indented shortcodes](https://gohugo.io/templates/shortcode-templates/#nested-shortcode-image-gallery)), so I kept digging.

Eventually I found Basically, Hugo has two kinds of delimiters:

- `{{% ... %}}` is the [normal delimiter to use](https://gohugo.io/content-management/shortcodes/#use-shortcodes) if your shortcode contains markdown content that you want formatted.
- `{{< ... >}}` is the delimiter to use if the contents of the shortcode [do not have markdown formatting](https://gohugo.io/content-management/shortcodes/#shortcodes-without-markdown) (e.g. if they are already raw HTML).

In this case I think we needed to use the second type of delimiter since, once the inner short-codes are formatted, they are just raw HTML.

Closes https://github.com/medic/cht-docs/issues/747

### DIFF
--- a/content/en/apps/examples/anc.md
+++ b/content/en/apps/examples/anc.md
@@ -43,11 +43,11 @@ Both maternal and newborn care workflows are defined to connect form actions and
 
 ### Pregnancy Workflow
 
-{{% workflow-table
+{{< workflow-table
   condition-image="pregnancy-1.png"
   task-image="pregnancy-2.png"
   resolution-image="pregnancy-3.png"
- %}}
+ >}}
   {{% workflow
         condition="Upon discovering a pregnancy, a CHW submits a [Pregnancy Form](https://docs.google.com/drawings/d/1u4OQIgTyzUysFv9Cop-C54nrbPiuPmq7GTpBTFmwZn0/edit) confirming a new pregnancy with the estimated gestational age."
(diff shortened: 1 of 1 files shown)


mrjones-plip (approved):
Simple, yet thorough fix!

Testing the branch looks good:

image

checking for other instances of the `workflow-table` looks good:

```sh
➜  cht-docs git:(747_fix_workflow_tables) grep -r workflow-table .   
./content/en/apps/examples/anc.md:{{< workflow-table
./content/en/apps/examples/anc.md:{{< /workflow-table >}}
./content/en/apps/examples/anc.md:{{< workflow-table
./content/en/apps/examples/anc.md:{{< /workflow-table >}}
```

LGTM!

PR #750, opened 2022-08-12
add troubleshooting section to bulk upload

mrjones-plip:
Covers troubleshooting for:
* Access denied
* Wrong type, this is not a person


mrjones-plip:
cc @tatilepizs

content/en/apps/guides/data/users-bulk-load.md
```diff
@@ -126,3 +126,72 @@ You will need to manually add these new places to the spreadsheet so that you ca
...
+## Trouble shooting
+
+### "Wrong type, this is not a person."
+
+If you have miss-matched contact types, you will get an error upon import:
+
+> **Wrong type, this is not a person.**
+
+
+As of CHT 3.7.0, you're [allowed to declare different contact types]({{< relref "apps/reference/app-settings/hierarchy#app_settingsjson-contact_types" >}}) in your `app_settings.json`. If you have added a `.contact_types[]` stanza to your JSON, you will need to update the automatic value of the `contact.contact_type` column.  The default value is:
```
m5r, line 139:
```suggestion
As of CHT 3.7.0, you're [allowed to declare different contact types]({{< relref "apps/reference/app-settings/hierarchy#app_settingsjson-contact_types" >}}) in your `app_settings.json`. If you have populated the `.contact_types[]` property in your JSON, you will need to update the automatic value of the `contact.contact_type` column.  The default value is:
```

I had never seen the word `stanza` before but it's pretty satisfying to say out loud :joy:  
Since the audience of these docs are usually not native English speakers, I think it would be safer to fall back to more common words

m5r (approved):
Great addition, thanks for your work!  
I left one suggestion inline but I'm already approving this PR so you don't have to wait for my approval when I come back on monday

content/en/apps/guides/data/users-bulk-load.md
mrjones-plip: Ah - it is a great word!  I'll go with the more common suggestion - thanks!

PR #751, opened 2022-08-16
Add documentation for the `add-date` XPath function

jkuester:
Adds documentation for the new `add-date` XPath function being added in https://github.com/medic/cht-core/pull/7729.

### DIFF
--- a/content/en/apps/reference/forms/app.md
+++ b/content/en/apps/reference/forms/app.md
@@ -199,6 +199,18 @@ To instruct the widget to process an array of strings or numbers, create a new `
 
 ## CHT XPath Functions
 
+### `add-date`
+
+_Available in +4.0.0._
+
+Adds the provided number of years/months/days/hours/minutes to a date value. 
+
+```
+add-date(date, years, months, days, hours, minutes)
+```
+
+This function is useful things like for calculating a person's birthdate date based on how many years/months old they are: `add-date(today(), 0-${age_years}, 0-${age_months})`. 
+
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/forms/app.md
```diff
@@ -199,6 +199,18 @@ To instruct the widget to process an array of strings or numbers, create a new `
...
+
+_Available in +4.0.0._
+
+Adds the provided number of years/months/days/hours/minutes to a date value. 
+
+```
+add-date(date, years, months, days, hours, minutes)
+```
+
+This function is useful things like for calculating a person's birthdate date based on how many years/months old they are: `add-date(today(), 0-${age_years}, 0-${age_months})`.
```
njogz, line 212: I think this can be made clearer by stating that you can add negative numbers to get a date in the past. I'm thinking something like

```This function is useful for calculating dates for example getting a date two weeks from now add-date(today(), null, null, 14)```


```You can also add negative numbers to get dates in the past which is useful for things like calculating a person's birthdate date based on how many years/months old they are: add-date(today(), 0-${age_years}, 0-${age_months})```

njogz (changes requested):
Left comment about being more explicit about being able to add negative numbers.

content/en/apps/reference/forms/app.md
jkuester: Good call!  Done.

njogz (approved):
Looks good

PR #754, opened 2022-08-23
add steps on how to use horti for first time install

mrjones-plip:


mrjones-plip:
cc @michaelkohn @elvisdorkenoo and @latin-panda - Care teams will use these same steps for your FRs too

PR #756, opened 2022-08-25
add 3.17.0 release notes

mrjones-plip:
per medic/cht-core#7725


mrjones-plip:
closing in favor of #767 - will port filter content to that PR

Issue #758, opened 2022-08-29
relatedContent on CHT Applications ->Features -> * not showing


mrjones-plip:
In this [targets page](https://docs.communityhealthtoolkit.org/apps/features/targets/) the code has some related content defined in the markdown (see below). However, the related content does not show at the bottom of the page. This is true for all the pages in 
`content/en/apps/features/*` but not for the sub-directory `/apps/features/integrations/*`.

Affected pages:

 * [Contacts](https://docs.communityhealthtoolkit.org/apps/features/contacts/)
 * [Messaging](https://docs.communityhealthtoolkit.org/apps/features/messaging/)
 * [Reports](https://docs.communityhealthtoolkit.org/apps/features/reports/)
 * [Supervision](https://docs.communityhealthtoolkit.org/apps/features/supervision/)
 * [Tasks](https://docs.communityhealthtoolkit.org/apps/features/tasks/)
 * [Targets](https://docs.communityhealthtoolkit.org/apps/features/targets/)
 * [UHC Mode](https://docs.communityhealthtoolkit.org/apps/features/uhc-mode/)
 * [App Management](https://docs.communityhealthtoolkit.org/apps/features/admin/)
 
```
---
title: Targets
weight: 8
description: >
  Dashboards to track metrics for an individual CHW or for an entire health facility
keyword: targets
relatedContent: >
  apps/reference/targets
  core/overview/db-schema#targets
  apps/features/integrations/dhis2
  apps/features/supervision/#chw-aggregate-targets
---
<!-- ## Targets: Performance Dashboards -->

*Targets* is the user dashboard or analytics tab. The widgets on this tab provide a summary or analysis of the data in submitted reports. These widgets can be configured to track metrics for an individual CHW, for a Supervisor overseeing a group of CHWs, or for an entire health facility. 
```

mrjones-plip:
I tried replacing the code for code for the targets page with known good content that did show the related content - this didn't fix it :thinking:

mrjones-plip:
@jkuester  - pinging you in case some of the [hugo uplift work](https://github.com/medic/cht-docs/pull/731) you corrected this already on some other pages and know an easy fix!  (but **_not_** because you are now the "fixes all the docs bugs" person :laughing: )

jkuester:
[...]

PR #760, opened 2022-08-30
Fix list pages so that Related Content section is displayed

jkuester:
Updates the location of the `list.html` page template so that it is actually used.  Hugo will try to use templates from `layouts/docs` before falling back to `layouts/_default`.  So, the [Docsy list template](https://github.com/google/docsy/blob/main/layouts/docs/list.html) was getting resolved first and was not overridden by our custom `list.html`.  

Understandably, this would break the "Related Content" section for any list pages (e.g. [`apps/features/integrations`](https://docs.communityhealthtoolkit.org/apps/features/integrations/)).  However, I do not understand exactly why this broke the related content for pages such as [`apps/features/contacts`](https://docs.communityhealthtoolkit.org/apps/features/contacts/) since that does not fit my understanding of a "list" page....  (This is further complicated by the fact that similar pages like [`apps/features/integrations/android`](https://docs.communityhealthtoolkit.org/apps/features/integrations/android/) are not broken (since they use the `content.html` template as expected...)  My best guess is this is a bug (feature?) in Hugo that is caused by the fact that [`apps/features/contacts`](https://docs.communityhealthtoolkit.org/apps/features/contacts/) sits at the _same level_ in the hierarchy as a list page: [`apps/features/integrations`](https://docs.communityhealthtoolkit.org/apps/features/integrations/).  Perhaps that triggers Hugo to treat all of the pages at that level as lists...

Either way, it is enough to say that fixing the `list.html` was enough to cause the "Related Content" to be displayed everywhere that it was missing!

Closes https://github.com/medic/cht-docs/issues/758

### DIFF
--- a/layouts/docs/list.html
+++ b/layouts/docs/list.html
@@ -1,5 +1,5 @@
-<!-- Modified from docsy theme to show related content -->
-<!-- https://github.com/google/docsy/blob/c65715725ee992cd4e59aeefe0f66df80b2eb4ea/layouts/docs/list.html -->
+{{/* Modified from docsy theme to show related content */}}
+{{/* https://github.com/google/docsy/blob/c65715725ee992cd4e59aeefe0f66df80b2eb4ea/layouts/docs/list.html */}}
 {{ define "main" }}
 <div class="td-content">
 	<h1>{{ .Title }}</h1>


layouts/docs/list.html
```diff
@@ -1,5 +1,5 @@
-<!-- Modified from docsy theme to show related content -->
-<!-- https://github.com/google/docsy/blob/c65715725ee992cd4e59aeefe0f66df80b2eb4ea/layouts/docs/list.html -->
+{{/* Modified from docsy theme to show related content */}}
+{{/* https://github.com/google/docsy/blob/c65715725ee992cd4e59aeefe0f66df80b2eb4ea/layouts/docs/list.html */}}
```
jkuester, line 2: This is more evidence that this template was never actually getting used by Hugo at all.  Hugo can be picky about where you can use normal HTML comments and where you have to use these special template comments.  Once I moved this file and it started getting used, Hugo failed to load the page because of these bad comments...

mrjones-plip (approved):
lgtm - thanks for the speedy turn around!

PR #766, opened 2022-08-31
core developer install: add make, g++ and ubuntu 18 notes

mrjones-plip:


garethbowen (approved):
Looks fine to me. Hopefully we can clean this up a lot when we drop python 2 and one day ubuntu 18.

PR #770, opened 2022-09-02
privacy policy now public as of 3.17

mrjones-plip:


content/en/apps/guides/privacy/privacy-policy.md
```diff
@@ -9,6 +9,11 @@ description: >
 ---
 
 As of `3.10.0`, privacy policies can be customized for every language, by adding desired content into HTML files. 
+
+As of `3.17.0`, privacy policies are publicly accessible, no longer only accessible by logged in users:
```
garethbowen, line 13: Thanks for this! Let's add a little bit about how to use the page.

```suggestion
As of `3.17.0`, privacy policies are publicly accessible, no longer only accessible by logged in users. This means it can be shared with third parties, for example, app store compliance. The URL to share is `{url}/{db}/privacy-policy`. _Added in 3.17.0_.
```

garethbowen (commented):
Great addition, thanks!

content/en/apps/guides/privacy/privacy-policy.md
latin-panda: Because we mentioned before `As of 3.17.0...` then we can probably skip the part `... Added in 3.17.0`.
You can add an example:

> If your instance url is `https://my-health-facility.org` then the privacy policy is available in `https://my-health-facility.org/medic/privacy-policy.`

Suggesting this because during training I noticed that our new teammates struggled to understand how the url should look like when reading the documentation.

garethbowen: @latin-panda Thanks for catching that! I've edited my suggestion accordingly.

latin-panda:
@mrjones-plip I think it'd be nice to have this merged before 3.17 release :)

mrjones-plip:
Yikes! Sorry I let this fallow and thanks so much for the nudge. I've put it out for review after committing the suggestion.

latin-panda (approved):
Thank you!!

PR #775, opened 2022-09-07
Create incremental-updates

michaelkohn:
Adding a page to illustrate how to implement incremental updates using permissions


content/en/apps/guides/updates/incremental-updates.md
```diff
@@ -0,0 +1,34 @@
...
+
+
+## Roll-out Scenarios
+
+The table below illustrates some potential rollout scenarios and recommended Actions.
+
+Scenario | Actions
+-- | -- 
+[Default] All users see the "new" version once you upgrade | None, just upgrade.
+All users see the "old" version once you upgrade. All users see the "new" version at the same time in the future. | Add the permission to all Roles before yo upgrade. When you are ready to transition everyone to the the new version, just remove the permission form all Roles.
```
garethbowen, line 31:
```suggestion
All users see the "old" version once you upgrade. All users see the "new" version at the same time in the future. | Add the permission to all Roles before you upgrade. When you are ready to transition everyone to the the new version, just remove the permission form all Roles.
```

content/en/apps/guides/updates/incremental-updates.md
```diff
@@ -0,0 +1,34 @@
...
+
+{{% alert title="Note" %}} This functionality is intended as a way to phase in changes, the old version will be completely removed in a future release. {{% /alert %}}
+
+## Configuration
+
+If the specific CHT Core update supports incremental rollouts, users can be configured to see the "old" version by means of a [permission]({{< ref "apps/reference/app-settings/user-permissions" >}}). The permission can be added to existing Roles, or for more granular control and rolling out by cohorts, you can create a new Role with the one permission and add that Role to desired users. It is important to note that currently the user interface of the Admin app does not support assignment of multiple Roles to a single user, but it is possible to do this through the API or in CouchDB directly. Also, the permission will not be included in `app_settings` automatically so if you want it to be selectable from the Admin app user interface, you will need to add it to `app_settings`.
```
garethbowen, line 21:
```suggestion
If the specific CHT Core feature supports incremental rollouts, users can be configured to see the "old" version by means of a [permission]({{< ref "apps/reference/app-settings/user-permissions" >}}). The permission can be added to existing Roles, or for more granular control and rolling out by cohorts, you can create a new Role with the one permission and add that Role to desired users. It is important to note that currently the user interface of the Admin app does not support assignment of multiple Roles to a single user, but it is possible to do this through the API or in CouchDB directly. Also, the permission will not be included in `app_settings` automatically so if you want it to be selectable from the Admin app user interface, you will need to add it to `app_settings`.
```

content/en/apps/guides/updates/incremental-updates.md
```diff
@@ -0,0 +1,34 @@
+---
+title: "Rolling out CHT Core updates incrementally"
+linkTitle: "Incremental updates"
```
garethbowen, line 3: "incremental updates" has been used to mean delivering a new cht-core version to a subset of users so it may be confusing. Can we rebrand this to "Feature Flags" which is a more industry standard name for this concept? If you agree, this will need updates throughout the doc...

garethbowen (changes requested):
Wow! Fast turnaround, thanks!

content/en/apps/guides/updates/incremental-updates.md
michaelkohn: I thought there was a conversation at some point arguing against using "feature flag".... but I agree and have updated it!

garethbowen (approved):

PR #777, opened 2022-09-08
Adding monitoring page to hosting guides

mrjones-plip:


mrjones-plip:
Hey all - while I don't need all three of the reviewers to do a deep dive on this PR, I would like a quick sanity check that his is on target enough to publish out.

Thanks!

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+Alerting, must be a push mechanism, forcibly notifying users who can act on the alert. These can go over SMS, email, Slack, WhatsApp or any other mechanism that will find it's way on to the mobile or desktop device of the right users. 
+
+The process of setting up monitoring and alerting should be done together. Monitoring sets the baseline and then alerting tells admins when the metric has gone beyond the baseline to a critical state. Certain metrics, like uptime for example, likely do not need to have a monitoring visualization on a dashboard, but the monitoring system should still be the authority to send an alert to denote when the service has restarted unexpectedly.
+
+## Outside the CHT
+
+Be sure to monitor important items that the CHT depends on in order to be healthy. You should alert when any of these are close to their maximum(disk space)  or minimum (days left of valid TLS certificate):
```
garethbowen, line 25:
```suggestion
Be sure to monitor important items that the CHT depends on in order to be healthy. You should alert when any of these are close to their maximum (disk space) or minimum (days left of valid TLS certificate):
```

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -0,0 +1,72 @@
...
+* Disk & swap space
+* CPU utilization
+* Memory utilization
+* Network utilization
+* Process count
+* OS Uptime
+
+## Inside the CHT
+
+As of [CHT 3.9.0]({{% ref "core/releases/3.9.0.md#instance-monitoring" %}}), the [Monitoring v1 API]({{< relref "apps/reference/api#get-apiv1monitoring" >}}) has been available. In  [CHT 3.12.0]({{% ref "core/releases/3.12.0.md#instance-monitoring" %}}) this API was depricated in [favor of v2]({{< relref "apps/reference/api#get-apiv3monitoring" >}}). These APIs do not require any authentication and so can easily be used with third party tools as they do not need a CHT user account.
```
garethbowen, line 38: Let's just simplify this - the API docs cover the version and deprecation.

```suggestion
The [monitoring API](({{< relref "apps/reference/api#get-apiv2monitoring" >}})) was added in 3.9.0 and does not require any authentication and so can easily be used with third party tools as they do not need a CHT user account.
```

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+All metrics need to be monitored over time so that you can easily see longitudinal patterns when debugging an outage or slow down. The names below are extrapolated from the paths in the JSON returned by the API and should be easy to find when viewing the `/api/v1/monitoring` or `/api/v2/monitoring` URLs on your CHT instance.
```
garethbowen, line 40:
```suggestion
All metrics need to be monitored over time so that you can easily see longitudinal patterns when debugging an outage or slow down. The names below are extrapolated from the paths in the JSON returned by the API and should be easy to find when viewing the Monitoring API URL on your CHT instance.
```

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -0,0 +1,72 @@
...
+
+* Messaging Outgoing State Failed
+* Outbound Push Backlog
+* Sentinel Backlog
+
+### Zero or Near Zero Values
+
+Finally, these values should always be _not_ zero, and you should alert when are zero or very close to it.  You may opt to alert only when they are zero for more than 24 hours:
+
+* Date Uptime
```
garethbowen, line 72: I think this section should be restructured into a single list, potentially a table with a column for "expected value" or similar.

garethbowen (commented):
Really useful and I think targeting the right audience.

Consider also linking from the API page to here, because right now it's quite dry.

henokgetachew (approved):
Looks good to me overall. If the target audience is not expected to be experienced however, we should consider suggesting example tools for monitoring.

content/en/apps/guides/hosting/monitoring.md
mrjones-plip: great thinking! fixed in 09f309d

mrjones-plip:
@garethbowen  - thanks so much for the review - much appreciated!  back to you for re-review after I've made changes.

garethbowen (approved):
Awesome, thanks!

PR #782, opened 2022-09-15
Document the new "roles" property

garethbowen:
medic/cht-core#7592

### DIFF
--- a/content/en/apps/reference/api.md
+++ b/content/en/apps/reference/api.md
@@ -770,6 +770,8 @@ If `app_settings.app_url` is not defined, the generated token-login URL will use
 
 Returns a list of users and their profile data in JSON format.
 
+NB: The `roles` property was added in cht-core v4.1.0.
+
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/api.md
```diff
@@ -813,6 +817,7 @@ Content-Type: application/json; charset=utf-8
     "contact": {
       "_id": "eeb17d6d-5dde-c2c0-62c4a1a0ca17fd17",
       "type": "person",
+      "roles": [ "chw" ]
```
dianabarsan, line 820: I'm unsure if the "roles" property needs to be added to the contact.

dianabarsan (approved):
I think I spotted a miss, but otherwise great. Approving so it won't block you.

content/en/apps/reference/api.md
garethbowen: Definitely not! Fixed, thanks.

PR #783, opened 2022-09-16
Update direct-to-patient.md

antonykhaemba:
Updated


nsitaula (approved):
Hi Antony, I did not see changes except header for:
3. Ways to improve response rate from patients or household members?

PR #784, opened 2022-09-16
Tiny telemetry update from filter redesign work.

latin-panda:


latin-panda:
@michaelkohn this is the tiny thing I forgot to add about the filter work. Please have a look :)

michaelkohn (commented):
Thanks for adding this! It might be too late to easily change this and if so, definitely fine leaving it as is. But if it's still easy, what do you think about some other naming options? Just thinking ahead and also how we might organize telemetry fields for monitoring UX interaction (most of our other telemetry stuff is really about performance).

`usage:reports:sidebar_filter:open`
category:page:feature:action

`ux:reports:sidebar_filter:open`
Same as above but a _different_ "category"

`reports:sidebar_filter:open`
Same as above but _without_ the "category"

`search:reports:sidebar_filter:open` 
Because sidebar filter is related to many of the other `search:reports:xxxx` items (though by being under "search" implies that search was actual performed, whereas what we are measuring here is how many times the person _thought_ about doing a search :-)

latin-panda:
@michaelkohn looking at the other telemetry entries, it seems that the name is like this: `what:where:detail` being `what` the "subject" which the entry is about.

For example:
`boot_time:purging:<boolean>`
`rules-engine:tasks:all-contacts`
`replication:<database>:<direction>:failure:reason:error`
`enketo:contacts:<form>:add:<component>`
`enketo:reports:<form>:<action>:<component>`

Thinking about it... if I want to look for an entry, my initial reaction is to look for the `what` first then the `where`. If we want to adopt a formula, my preference would be something closer to natural language:
`search:contacts:<filter[:filter]>` --> `The search in contacts tab applied the date filter`

I think it's good idea to write a formula in the docs: "If you're adding a new entry please use this formula ......" and ask the research team(?) to see if that makes sense 🤔

As for the new `sidebar_filter:reports:open`, I can update the name when working in the bulk delete work, because the filter work is ready to ship.

latin-panda:
@michaelkohn what do you think? :)

michaelkohn (approved):

Issue #786, opened 2022-09-19
Link to the Core Developer Setup Guide from the Contributing Code page


jkuester:
Page: /contribute/code/
Version: 2021-11-24 11:32:41  1300  1300
---

I have had a lot of trouble remembering where to find the [Core Developer Setup Guide](https://docs.communityhealthtoolkit.org/apps/guides/hosting/core-developer/).  I think it makes sense to have it where it is (given the overall structure of these docs), but it is not very discover-able for someone just digging through the docs looking at it.  

It would be very helpful to just link it from the Contributing Code page since that is likely to be the first place a lot of developers will look....

PR #787, opened 2022-09-19
Update Contributing Code page to link to Dev Setup Guide

jkuester:
Closes https://github.com/medic/cht-docs/issues/786


mrjones-plip (approved):
more cross linking more better - thanks @jkuester !

PR #805, opened 2022-09-29
Add documentation for create_users_for_contacts transition

jkuester:
Adds documentation for the new `create_users_for_contacts transition.

Issue: https://github.com/medic/cht-core/issues/7703
cht-core changes: https://github.com/medic/cht-core/pull/7709

### DIFF
--- a/content/en/apps/reference/app-settings/transitions.md
+++ b/content/en/apps/reference/app-settings/transitions.md
@@ -684,7 +684,13 @@ Imagine a CHW is leaving the program, and the CHW's device is returned to their 
 
 To do this, when the `create_user_for_contacts` transition is enabled, the supervisor would submit a configured user replacement form for the original user's contact on the device. This form can create a new contact for the new CHW and will trigger a client-side transition to mark the original contact as replaced. After that, the supervisor can give the device to the new CHW, and they can begin using it.
 
-Subsequent reports submitted on the device by the new CHW will be associated with the new contact. When the device is eventually able to synchronize with the server, it will be automatically logged out (as it was actually still logged in as the original user). A server-side transition will be triggered to inactivate the original user and to create a new user for the new CHW. An SMS message containing a token login link will be sent to the new CHW allowing them to login as their proper user. 
+Subsequent reports submitted on the device by the new CHW will be associated with the new contact. When the device is eventually able to synchronize with the server, it will be automatically logged out so the transition to the new user can be completed. A server-side transition will be triggered to initialize the new user for the new CHW. An SMS message containing a token login link will be sent to the new CHW allowing them to login as the initialized user. The password for the original user will be automatically reset by the server-side transition causing any remaining sessions for the original user (e.g. on other devices) to be logged out. 
+
+#### Details
+
+This process does not actually delete the original user, but just resets the password to a random value. To recover the original user, a server administrator should update the user's password to a known value or re-issue a token login link for the user (if enabled).
+
+Because the server-side transition immediately invalidates any remaining sessions for the original user, it is not recommended to use this process for replacing users that are logged in on multiple devices simultaneously. The data on the device used to replace the user will always be completely synchronized before the user is replaced. However, un-synced data from other devices can be left on those devices when the user is replaced using a separate device.
 
 #### Configuration
 
@@ -704,7 +710,7 @@ These forms should only be submitted for the _original user's contact_. You can 
 
 These forms should only be accessible to offline users (replacing online users is not currently supported). This is the default in the example app form properties file ([see `replace_user.properties.json`](https://github.com/medic/cht-core/blob/master/config/default/forms/app/replace_user.properties.json)).
 
-You can prevent a user from being replaced multiple times by including `!contact.user_for_contact || !contact.user_for_contact.replace` in the form properties expression.
+You can prevent a user from being replaced multiple times by including `!contact.user_for_contact || !contact.user_for_contact.replace` in the form properties expression.  _(This expression is not recommended for situations where multiple users can be associated with the same contact since replacing one of the users would prevent any of the other users for that contact from accessing the form.)_
 
 See the `replace_user` app form provided in the [Default config](https://github.com/medic/cht-core/tree/master/config/default) as an example.
 
--- a/content/en/apps/reference/app-settings/transitions.md
+++ b/content/en/apps/reference/app-settings/transitions.md
@@ -54,7 +54,7 @@ The following transitions are available and executed in order.
 | [muting](#muting) | Implements muting/unmuting actions of people and places. Available since 3.2.x. Is partially applied on the client, as of 3.12.0. |
 | [mark_for_outbound]({{% ref "apps/reference/app-settings/outbound" %}}) | Enables outbound pushes. Available since 3.5.x |
 | [self_report](#self_report) | Maps patient to sender. Available since 3.9.x |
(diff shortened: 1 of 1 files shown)


mrjones-plip (changes requested):
Are we planning on having another docs PR for information about [the XLSX](https://github.com/medic/cht-core/blob/cf9f1be6855429a301e64c105b3fad35f28cf614/config/default/forms/app/replace_user.xlsx) form itself?  I guess this falls more into reference app land than reference documentation...

I think let's just KISS for now and add a one liner along the lines of "To test using this feature per the example scenario, update your app settings as specified here, and then upload the `replace_user` app form. "

jkuester:
@mrjones-plip regarding the XLSX file, I did include:

> See the `replace_user` app form provided in the [Default config](https://github.com/medic/cht-core/tree/master/config/default) as an example.

in the doc section for setting the replace forms.  Is there more you would like to see then that?  

I guess I am thinking along the lines of `reference app land than reference documentation`.  The `replace_user` form is intended as an example, but it should not be expected to work out-of-the-box for anyone who uploads it to their server.

mrjones-plip (approved):

jkuester:
@mrjones-plip technically, any changes needed for the `replace_user` form would be only tangential to the actual `create_users_for_contacts` functionality.  E.g.:

- The form properties expression would need to be updated to include the desired offline user roles (if you needed something besides `chw` and `chw_supervisor`), but that is just basic form visibility control (and not specific to this functionality).  
- Also, if you are using a custom hierarchy and contact types, you would need to update the `replace_user` form config to properly create the new contact.  However, the same can be said for the other forms that create contacts (e.g. `delivery`). Deployments with custom hierarchies have to edit any of the example forms that handle contacts.

I am happy to add documentation around this if we think it is necessary (either now or in a future PR), but I think it probably should go under [Examples](https://docs.communityhealthtoolkit.org/apps/examples/) or [Tutorials](https://docs.communityhealthtoolkit.org/apps/tutorials/) (not sure about the difference between those two...).

mrjones-plip (approved):
lgtm for reals this time.

thanks for the quick turn around!!

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -671,4 +672,63 @@ If this configuration is not set then the message defaults to what is set in the
...
+
+See the `replace_user` app form provided in the [Default config](https://github.com/medic/cht-core/tree/master/config/default) as an example.
+
+##### Example
+```json
+"app_url": "https://demo.app.medicmobile.org",
+"token_login": {
+  "enabled": true,
+  "translation_key": "sms.token.login.help"
+}
```
mrjones-plip, line 717: just found this when testing!

```suggestion
},
```

mrjones-plip (changes requested):
omg, i'm back

mrjones-plip (approved):
28th time is the charm?

lgtm

mrjones-plip:
@jkuester - the reason the `,` isn't showing up in that preview doc we were just looking is you need to mash that ~~like and subscribe~~ squash and merge :stuck_out_tongue:

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -671,4 +672,63 @@ If this configuration is not set then the message defaults to what is set in the
...
+
+##### `replace_forms`
+
+Forms that trigger the `create_user_for_contacts` transition must set the `replacement_contact_id` property to the id of the contact that should be associated with the new user. 
+
+These forms should only be submitted for the _original user's contact_. You can control the form visibility by including `user._id === contact._id` in the form properties expression. 
+
+These forms should only be accessible to offline users (replacing online users is not currently supported). This is the default in the example app form properties file ([see `replace_user.properties.json`](https://github.com/medic/cht-core/blob/master/config/default/forms/app/replace_user.properties.json)).
+
+You can prevent a user from being replaced multiple times by including `!contact.user_for_contact.replaced` in the form properties expression.
```
jkuester, line 707:
```suggestion
You can prevent a user from being replaced multiple times by including `!contact.user_for_contact || !contact.user_for_contact.replaced` in the form properties expression.
```

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -671,4 +672,63 @@ If this configuration is not set then the message defaults to what is set in the
...
+
+#### Example scenario
+
+Imagine a CHW is leaving the program, and the CHW's device is returned to their supervisor. The supervisor wants to transfer the device to a new CHW immediately without having the device online to sync with the server.
+
+To do this, when the `create_user_for_contacts` transition is enabled, the supervisor would submit a configured user replacement form for the original user's contact on the device. This form can create a new contact for the new CHW and will trigger a client-side transition to mark the original contact as replaced. After that, the supervisor can give the device to the new CHW, and they can begin using it.
+
+Subsequent reports submitted on the device by the new CHW will be associated with the new contact. When the device is eventually able to synchronize with the server, it will be automatically logged out (as it was actually still logged in as the original user). A server-side transition will be triggered to inactivate the original user and to create a new user for the new CHW. An SMS message containing a token login link will be sent to the new CHW allowing them to login as their proper user.
```
dianabarsan, line 687: Can the previous account still be used to log in? I think we should cover this in the documentation.

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -54,6 +54,7 @@ The following transitions are available and executed in order.
 | [muting](#muting) | Implements muting/unmuting actions of people and places. Available since 3.2.x. Is partially applied on the client, as of 3.12.0. |
 | [mark_for_outbound]({{% ref "apps/reference/app-settings/outbound" %}}) | Enables outbound pushes. Available since 3.5.x |
 | [self_report](#self_report) | Maps patient to sender. Available since 3.9.x |
+| [create_user_for_contacts](#create_user_for_contacts) | Triggers an existing offline user to be replaced with a new user on the same device (without needing to immediately sync with the server). Available since 3.17.x  | |
```
jkuester, line 57:
```suggestion
| [create_user_for_contacts](#create_user_for_contacts) | Triggers an existing offline user to be replaced with a new user on the same device (without needing to immediately sync with the server). Available since 4.1.x  | |
```

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -671,4 +672,63 @@ If this configuration is not set then the message defaults to what is set in the
     ]
   }
 ]
-```
\ No newline at end of file
+```
+
+### create_user_for_contacts
+
+An existing offline user can be replaced on a device so that a new user can use that device without needing to immediately sync with the server. _Available since 3.17.x._
```
jkuester, line 679:
```suggestion
An existing offline user can be replaced on a device so that a new user can use that device without needing to immediately sync with the server. _Available since 4.1.x._
```

content/en/apps/reference/app-settings/transitions.md
jkuester: Might need to update this more based on the the conversation here: https://github.com/medic/cht-core/pull/7709/files#r1002087972

jkuester, line 707: Need to include a note that this will not work for situations where multiple users are associated with the same contact.

m5r (approved):
Nicely explained!

dianabarsan (approved):
Very nicely written!

Issue #807, opened 2022-09-30
Improve 'API to interact with CHT Applications'


yrimal:
Page: /apps/reference/api/
Version: 2022-09-13 16:13:33 -0700 -0700
---

This page documents `reported_date` in form parameters for and in example. However, `reported_date` doesn't set the date of sms sent, instead `sent_timestamp` works. Update the document to reflect this.

PR #808, opened 2022-09-30
Correct field refrence in API reference

yrimal:
- Correct field reference as mentioned in #807


dianabarsan (approved):
Thanks for the quick fix!

PR #801, opened 2022-09-28
Update App Hosting guide to work with 3.x, call out it won't work with 4.x

mrjones-plip:
see also [cht-core PR](https://github.com/medic/cht-core/pull/7828)

PR #809, opened 2022-10-03
convert all compose files to use `docker-compose-developer-3.x-only.yml` until 4.0 is out

mrjones-plip:
@dianabarsan  - thanks for fielding forum posts about this!  while a temporary fix that we'll have revert fix soon, I think worthy while.

Mirrors works in #801 for docs and already merged  [cht-core PR](https://github.com/medic/cht-core/pull/7828)


dianabarsan (approved):
Nice, thanks for getting the documentation up to date!

PR #817, opened 2022-11-02
Add release notes for 4.0.0

garethbowen:
medic/cht-core#7804


mrjones-plip:
OMG!?!?! The release notes must mean we're REALLY close! \o/

i'll grab this in the next 24hrs - tops

mrjones-plip (approved):
Great work!  FOUR OH FOUR THE WIN-OH!

Approving, though 4.0 is not quite through AT/RT, so we might add a ticket or two more.

PR #841, opened 2022-11-13
Remove sensitive words

garethbowen:
medic/cht-core#6574


jonathanvq (approved):
Thanks for the changes

PR #859, opened 2022-11-17
Make upgrade instructions harder to miss

garethbowen:


mrjones-plip (approved):
Great! thanks for the quick turn around and for being receptive to feedback.

Issue #861, opened 2022-11-18
Improve 'app': Alert breaks flow


abbyad:
Page: /apps/reference/forms/app/
Version: 2022-11-18 12:40:24 -0500 -0500
---

The alert note breaks the flow of the page because the indenting is lost:
<img width="561" alt="image" src="https://user-images.githubusercontent.com/789512/202784809-4a086582-c452-4b95-a95a-cfff895f56a1.png">

Proposing that we move the text to within the bullet point.

abbyad:
Indenting the note itself does not work well

abbyad:
On second attempt, it works much better

<img width="522" alt="image" src="https://user-images.githubusercontent.com/789512/202785738-45aabad6-c91e-4789-a33d-545b4f243430.png">

This was done using two space indent in markdown, as opposed to four spaces in previous comment.

PR #862, opened 2022-11-18
861 - Refactor XForm widgets

abbyad:
The widgets were previously (mostly) in a bulleted list, which made the flow difficult to read, and impossible to link to individual widgets in documentation. This version moves each widget to its own subsection so that they can be referenced properly. This structure also allows for further details to be included, such as screenshots and other details for their use.

For https://github.com/medic/cht-docs/issues/861

### DIFF
--- a/content/en/apps/reference/forms/app.md
+++ b/content/en/apps/reference/forms/app.md
@@ -86,24 +86,32 @@ The `+` operator for string concatenation is deprecated and will be removed in a
 
 ## CHT XForm Widgets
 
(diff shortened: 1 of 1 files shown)


abbyad:
This is more heavy handed approach to dealing with the original issue, yet has additional benefits which I think are worth including at this stage.

abbyad:
Reassigned reviewer due to availability

jkuester (approved):
:rocket: Much better than a bulleted list!

PR #864, opened 2022-11-23
Update architecture docs for 4.0.0

garethbowen:


content/en/core/overview/architecture.md
```diff
@@ -8,31 +8,59 @@ description: >
...
 
 Another NodeJS service running on the server, sentinel performs actions called transitions every time a document in CouchDB is added or modified. Some examples are validations, generating scheduled messages, automatic responses, creating patients, and sending alerts. See more at the [CHT Core Sentinel repo](https://github.com/medic/cht-core/tree/master/sentinel) on Github. 
 
-### PostgreSQL
+#### CouchDB
 
-A free and open source SQL database that we use for analytics queries for display in tools like klipfolio. We created a library called [couch2pg](https://github.com/medic/couch2pg) to replicate data from CouchDB into PostgreSQL. See more at the [PostgreSQL](https://www.postgresql.org) site.
+A free and open source NoSQL database used to store all data and configuration. This can be multiple instances clustered together for additional scalability. CouchDB is really good at replication which is the process of sending the data to another database and back again, which makes it ideal for replicating data to the phone for offline access. See more at the [CouchDB](http://couchdb.apache.org) site.
+
+#### nginx
```
abbyad, line 31: What's your take on nginx, Nginx, and NGINX (ignoring NGiИX)? Would be nice to capitalize title for consistency, but I realize it's a bit tricky here since the given name is actually nginx.

content/en/core/overview/architecture.md
```diff
@@ -8,31 +8,59 @@ description: >
...
+
+Provides SSL termination and routes requests to API.
+
+#### haproxy
```
abbyad, line 35:
```suggestion
#### HAProxy
```

content/en/core/overview/architecture.md
```diff
@@ -8,31 +8,59 @@ description: >
...
+
+Provides audit logging for any request that makes it to CouchDB so any data access or modification can be validated at a later date.
+
+### CHT Sync
+
+The cht-sync product is a suite of tools for extracting and normalizing data from the Core Framework's CouchDB and rendering the data in analytics dashboards to measure the performance of the program. Read more detail in [cht-sync GitHub repository](https://github.com/medic/cht-sync).
```
abbyad, line 41: I find it awkward to start with the repo name "cht-sync", suggesting to avoid using it right away in the sentence.

```suggestion
A suite of tools for extracting and normalizing data from the Core Framework's CouchDB and rendering the data in analytics dashboards to visualize key data for a CHT deployment. Read more detail in [`cht-sync` GitHub repository](https://github.com/medic/cht-sync).
```

content/en/core/overview/architecture.md
```diff
@@ -8,31 +8,59 @@ description: >
...
+
+[DBT](https://www.getdbt.com/) is used to ingest raw JSON data from the postgres database and normalize it into a relational schema to make it much easier to query.
+
+#### Superset
+
+[Apache Superset](https://superset.apache.org/) is a free an open source platform for creating data dashboards.
 
 ## Client side
 
 ### cht-core
```
abbyad, line 61: Having the "given name" would be helpful here too.

```suggestion
### CHT Core Framework
```

content/en/core/overview/architecture.md
```diff
@@ -8,31 +8,59 @@ description: >
...
 
-[cht-core](https://github.com/medic/cht-core) is composed of the following web applications:
+[cht-core](https://github.com/medic/cht-core) is composed of two web applications:
```
abbyad, line 63: The `:` is hard to follow since you can't see the two web applications in a list. I'd suggest explain the two apps immediately to provide context, and linking down (actual links missing in the suggestion)

```suggestion
The CHT Core Framework provides two web applications, the [CHT app](#) for care teams and program staff, and an [app management interface](#) for program administrators. 
```

abbyad (changes requested):
Great to have this additional info, and the CHT Sync info is really helpful!

I'll start with the recommendations for the diagram since they may have knock-on effects in the text. Also, I am happy to wrangle the diagram if you prefer!
- The text differentiates Server vs Client sides, it would help to see this represented in the diagram too.
- Related to the client side, it may help to have additional context for the "integrations" and "browser" boxes. For instance, having "Integrations with External Tools" or "Integrations with Digital Health Ecosystem", and "Browser-based Dashboards". The later could work into the "Client side" point above.
- Should be consistent with capitalization (we have `CouchDB` and `haproxy`). I would suggest we use given name, like HAProxy, not service names since the diagram will help less technical audiences orient themselves.
- Similarly, we should use full given name for PostgreSQL vs Postgres (which is what is there now) vs postgres (which we don't use)
- I think it will be clearer to differentiate CHT vs external components if we specify `CHT` when the tool is not a standalone third party tool, as in `CHT Sentinel` and `CHT API`
- `cht-upgrade` is never mentioned in the text, should it be removed? And if not, for consistency should it be `CHT Upgrade`?

For the doc page, I made specific comments inline, and they generally fall into these categories:
- The capitalization of headers should be consistent, and I think we should use the "given names" in the headers rather than the service/tool names.
- Some of the given names include CHT, others don't. We should be more consistent, especially in the headers.
- We haven't formalized this in the style guide, but I think we should use the given name in most text ("CHT Sync" rather than "cht-sync") unless there is a reason to refer to the tool specifically. Also, when we do use the tool/repo name it helps to clarify it as such by styling it as code, as in "`cht-sync`".

content/en/core/overview/architecture.md
garethbowen: I went with all caps which they use in their documentation.

content/en/core/overview/architecture.md
garethbowen: Good call!

garethbowen:
@abbyad I think I caught all your feedback and some additional improvements. Can you please re-review?

abbyad (approved):
Looks good! We can further explain some of the concepts, visually through the diagram or more clarity on what CHT Core Framework actually is in the text, but this is already a significant improvement and no reason to hold back this PR. Nice work!

Issue #868, opened 2022-11-24
Broken link in z-score documentation


nsitaula:
The link to the [This example XForm form](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#example-use) in [z-score documentation](https://docs.communityhealthtoolkit.org/apps/reference/forms/app/#z-score) is broken. Is it moved in some other location?

PR #869, opened 2022-11-24
Fix broken link in demo form

garethbowen:
#868


garethbowen:
@nsitaula We moved this script to a private repo to make it clear it's not supported but for a demo it's fine. I've updated the link to tie it to a point in time where the file still existed...

nsitaula (approved):

Issue #870, opened 2022-11-25
Build status image is broken


abbyad:
The build status image is broken, likely because the [build name changed](https://github.com/medic/cht-docs/pull/266).

# Broken image
> <img width="645" alt="image" src="https://user-images.githubusercontent.com/789512/204042975-b60dbc8f-0f92-4231-bf34-a4ab1532cd1c.png">

# Live view

The image below should resolve when this issue is fix, assuming that we restore the same name for the GH Action.

> [![CHT Documentation Site Build](https://github.com/medic/cht-docs/workflows/CHT%20Documentation%20Site%20Build/badge.svg)](https://github.com/medic/cht-docs/actions)

abbyad:
_Seems_ to work with #871!

<img width="695" alt="image" src="https://user-images.githubusercontent.com/789512/204869782-f2ded0bf-afae-475a-ae77-dee51c69155a.png">

It is possible that it represents all builds, not just `main` though. Will test to find out.

abbyad:
Tested again, and with an error in a branch the status still shows as passing, which is correct. I haven't tested that an error in main shows as an error, but if we notice that we can open a new issue.

<img width="889" alt="image" src="https://user-images.githubusercontent.com/789512/226106686-61058f39-11d4-4cd7-bf5a-37e82898aa0c.png">

PR #871, opened 2022-11-25
Revert build name to fix status image

abbyad:
This change will restore the working build status image in the readme, but I am not certain that this is actually how we want to do it. Will **non-**main builds also be represented by this status? If so we need to reconsider this approach.

An alternative is to remove the status image, which seems to have been broken for several years without complaint.
For https://github.com/medic/cht-docs/issues/870

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,9 @@
 
 on: [push, pull_request]
 
-name: Test site and push live if we're on Main
+# This action tests site and pushes live if we're on Main
+# Changes to the `name` will break the build status image in the readme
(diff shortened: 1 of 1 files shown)


mrjones-plip (approved):
Huh - good catch! I suspect that was me some years ago.  Total bummer - sorry!  (However, I take pride now that there's things I did years ago at Medic \o/)

Let's push this out and see how it works in practice.  Low stakes if we get it wrong!

garethbowen:
FYI when changing the build name you also have to update the branch protection rules otherwise PRs get blocked waiting for the old action to complete.

PR #883, opened 2022-11-29
Update language and headers

abbyad:


n-orlowski (approved):

n-orlowski (commented):
Is the national scale on here twice?

PR #897, opened 2022-12-02
Update of CHT4 local setup documentation

esthermmoturi:


content/en/apps/tutorials/local-setup.md
```diff
@@ -67,12 +67,15 @@ COUCHDB_PASSWORD=password
...
 ```
 
-Run the following command to start your CHT instance using Docker Compose:
+Run the following command to start your CHT instance using Docker Compose. Ensure that Docker is installed and running:
 
 ```shell
 cd ~/cht-local-setup/upgrade/
 docker-compose up
 ```
+Note that the first time you run your CHT instance it may take a while. Incase you run into issues running your docker file, ensure that the following setting in Docker is checked.
```
andrablaj, line 76: "In case" rather than "Incase"

andrablaj, line 76: I am not sure if this step is mandatory. I know that Mac M1 users ran into an error (do you remember what the error was @jonathanvq ?) and had to check the box, but I am not sure if it the case for every user.

content/en/apps/tutorials/local-setup.md
```diff
@@ -94,12 +97,23 @@ This CHT instance is empty and has no data in it. While you're free to explore a
 
 ### 2. Install cht-conf
 
-Using npm and python on your terminal, install cht-conf and pyxform globally using the following commands:
-
+Using npm on your terminal, install cht-conf globally using the command below. Confirm if you have previously installed cht-conf.
+If you have, skip this step.
```
andrablaj, line 101: I would say "If you have previously installed cht-conf, skip this step."

content/en/apps/tutorials/local-setup.md
```diff
@@ -94,12 +97,23 @@ This CHT instance is empty and has no data in it. While you're free to explore a
...
 ```shell
 npm install -g cht-conf
+```
+Using python on your terminal, install pyxform globally using the command below. 
+```shell
 sudo python -m pip install git+https://github.com/medic/pyxform.git@medic-conf-1.17#egg=pyxform-medic
 ```
+Incase you encounter issues while installing pyxform using the command above. Paste the commands below in your terminal:
```
andrablaj, line 109: "In case"

andrablaj, line 109: I believe that the issue is related to `python` rather than `pyxform`. I would explain that one needs to install python 2 to have the command above running correctly.

andrablaj (changes requested):

content/en/apps/tutorials/local-setup.md
```diff
@@ -67,12 +67,15 @@ COUCHDB_PASSWORD=password
 EOF
 ```
 
-Run the following command to start your CHT instance using Docker Compose:
+Run the following command to start your CHT instance using Docker Compose. Ensure that Docker is installed and running:
```
mrjones-plip, line 70: The prior section is title "Required Resources" and states:

> Before you begin, you need to have some useful software and tools that are required for things to work:

And then lists docker and docker composer.  This seems redundant and I think we should remove it.

content/en/apps/tutorials/local-setup.md
mrjones-plip: Should be fine for everyone to follow!  macOS and Windows folks should check it, and not that many LInux folks are running Docker Desktop for LInux, but it should be OK for them too.

content/en/apps/tutorials/local-setup.md
mrjones-plip: I would remove this addition.  This whole guide is has many things that need to be installed and it'll end up being quite laborious to say "Install docker (skip if you've installed docker)" and "Install nodejs (skip if you've installed nodejs)". 

These steps are both harmless to run a again and are not time consuming for a beginner to re-run if they're unsure.

Brevity is the soul of wit! ;)

content/en/apps/tutorials/local-setup.md
mrjones-plip: We have a "Troubleshooting" section on this already, I recommend:

```suggestion
If you encounter the error `npm ERR! gyp ERR verb find Python Python is not set` while installing pyxform and are running macOS, see [this troubleshooting section]({{< relref "contribute/code/core/dev-environment#macos--123" >}}).
```

And then we should remove the code below.

content/en/apps/tutorials/local-setup.md
esthermmoturi: I agree

content/en/apps/tutorials/local-setup.md
esthermmoturi: This has been added

mrjones-plip (approved):
lgtm!

andrablaj (approved):

PR #904, opened 2022-12-07
Move from master to 4.0.1 as  stop gap

mrjones-plip:
Until we can figure an easy way to dynamically resolve the latest version, possibly like is [being considered over here](https://github.com/medic/cht-core/pull/7956#issuecomment-1336775724)


dianabarsan (approved):
Cool beans

PR #914, opened 2022-12-15
Action bar - feature flag

latin-panda:
This adds details about the feature flag to enable the old action bar. This permission will hide More Options menu and FAB.

https://github.com/medic/cht-core/issues/7872


michaelkohn (approved):
LGTM!

Issue #831, opened 2022-11-09
Document 4.x self hosting


garethbowen:
How do we recommend self hosters run 4.x today.

Needs to cover:

- single, multi, and distributed couchdb nodes.
- backups
- logs
- etc

henokgetachew:
PRs submitted.

mrjones-plip:
Docs are live!
* https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/self-hosting-multiple-nodes/
* https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/self-hosting-single-node/
* https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/adding-tls-certificates/

mrjones-plip:
we didn't hit this, so reopening:
backups
logs

mrjones-plip:
with logs and backups added, closing this ticket out:
* https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/logs/
* https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/backups/

PR #915, opened 2022-12-15
Single node self hosting docs

henokgetachew:
First PR for [this](https://github.com/medic/cht-docs/issues/831) ticket.

### DIFF
--- a/content/en/core/overview/self-hosting-single-node.md
+++ b/content/en/core/overview/self-hosting-single-node.md
@@ -60,7 +60,7 @@ SSL_KEY_FILE_PATH=/etc/nginx/private/key.pem
 
 *Note:*
 * Make sure that all the directory paths are absolute paths and not relative paths.
-* COUCHDB_UUID can be generated [here](https://www.uuidgenerator.net/)
+* COUCHDB_UUID can be generated at [uuidgenerator.net](https://www.uuidgenerator.net/)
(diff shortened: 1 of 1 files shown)


content/en/core/overview/self-hosting-single-node.md
```diff
@@ -0,0 +1,86 @@
+---
```
garethbowen, line 1: Please move this to beside the other hosting docs here: https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/

content/en/core/overview/self-hosting-single-node.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+*Note:*
+* Make sure that all the directory paths are absolute paths and not relative paths.
+* COUCHDB_UUID can be generated [here](https://www.uuidgenerator.net/)
+* Make sure certificate is named `cert.pem` and key is named `key.pem` - this is right now a requirement because of a known [bug](https://github.com/medic/cht-core/issues/7949)
+
+### Get the certificate loaded into a cht volume
+
+Follow the set of steps outlined [here](https://github.com/medic/cht-core/pull/7834#issuecomment-1268710481).
```
garethbowen, line 68: Let's convert this issue comment to a docs page and then link from here and the multi-node docs. If it's not trivial then feel free to raise an issue to do that one later.

garethbowen (changes requested):
Awesome work, thanks! A few suggestions inline.

mrjones-plip:
closing  this one in favor of a single PR in #916 - will move this content there

PR #917, opened 2022-12-15
fix spelling of "infrastracture"

mrjones-plip:


michaelkohn (approved):
You nailed it, that's definitely a better way to spell it.

PR #919, opened 2022-12-19
Adding guidance for quality assistance

craig-landry:
This creates the testing page as well, which is empty aside from the quality assistance link for now. Ideally there would be more here, but I'd rather keep moving than wait for everything else to exist first.

### DIFF
--- a/content/en/product-handbook/testing/quality-assistance.md
+++ b/content/en/product-handbook/testing/quality-assistance.md
@@ -1,0 +1,55 @@
+---
+title: "Quality Assistance"
+linkTitle: "Quality Assistance"
+weight: 1
+description: >
+  How the Quality Assistance process works on the team
+---
+
+## Goals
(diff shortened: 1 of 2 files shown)


content/en/product-handbook/testing/quality-assistance.md
```diff
@@ -0,0 +1,55 @@
+---
+title: "Quality Assistance"
+linkTitle: "Quality Assistance"
+weight: 1
+description: >
+  How the Quality Assistance process works on the team
```
andrablaj, line 6: I would use "at Medic", rather than "on the team", so that people who are not from Medic and land on this page would understand right away to whom this process applies.

andrablaj (approved):

content/en/product-handbook/testing/quality-assistance.md
craig-landry: Good call! I'll correct it.

Issue #900, opened 2022-12-03
Change favicon to match new CHT logo


abbyad:
Given the release of the [new CHT logo](https://forum.communityhealthtoolkit.org/t/introducing-the-cht-logo/2322), we should update the favicon for the docs site.

Add favicon images for dark and light themes.

abbyad:
The https://github.com/medic/cht-docs/commit/1641f05cb2ea222529ee4be0a91b7830e0d44fba commit was accidentally pushed to main instead of a PR. From my testing though it looks good, so I haven't reverted it to go through the PR process. @craig-landry, when you get a chance, can you confirm that it is looking alright on your side too?

<img width="635" alt="image" src="https://user-images.githubusercontent.com/789512/209021563-92b878a0-ca0d-4d45-872f-31538b749aa3.png">

craig-landry:
@abbyad the new favicon shows as expected for me on Firefox 108.0.1 on Ubuntu Linux.
![image](https://user-images.githubusercontent.com/1485206/209144708-97d16646-a619-4c86-8972-72b8b7a99085.png)

2 items worth noting:
1. I had to hard refresh to get the logo to show up as it was cached as the old logo.
1. As you probably noticed, my Firefox dark theme (the default, I think) makes the logo less visible/attractive. 

For that light/dark thing, my understanding is that browsers can choose to respect the media attribute for this like:
```
<link rel="icon" media="(prefers-color-scheme: dark)"  href="/icons/favicon-dark.png">
<link rel="icon" media="(prefers-color-scheme: light)"  href="/icons/favicon-light.png">
```
I went to edit the partial to give it a shot, but see we don't have such a partial where I think it comes from the base theme we inherit from. I don't know how to do the override on that. I've created https://github.com/medic/cht-docs/issues/924 so we can get to it in due time.

Cheesy dark mode joke:
> Why do programmers prefer dark mode?
> Because light attracts bugs.

😆

craig-landry:
Update: I got the theme and color thing worked out and opened https://github.com/medic/cht-docs/pull/925 @abbyad please have a look and review the PR.

abbyad:
Thanks for the review. I agree it doesn't look great with all tab colors, so will look at other options.

[...]

PR #925, opened 2022-12-22
Adding dark-mode favicon capability and images

### LINKED ISSUE #924: Set up light and dark favicons
Currently we just have one option, which is good on light/regular windows, but not so good on "dark" themes:
image

The following format seems to be the recommended approach for letter the browser show the desired image for the current UI theme:

```
<link rel="icon" media="(prefers-color-scheme: dark)"  href="/icons/favicon-dark.png">
<link rel="icon" media="(prefers-color-scheme: light)"  href="/icons/favicon-light.png">
```

craig-landry:
Resolves #924

Our current theme inherits from docsy and the favicons are generates from the following template: https://github.com/google/docsy/blob/1d126ab1bbf186fa0ea52a6d6a3f58229b2b7942/layouts/partials/favicons.html

This changes adds a new favicon layout to override the base. The new layout is the same as the base, but adds the media attribute to supply browsers with light/dark options should they choose to respect it.

White color images are included for all sizes for the dark option.

Here's how this looks on Firefox 180 on Ubuntu Linux with the dark themed UI in Gnome:
image


abbyad (changes requested):
I had a feeling dark mode was going to elicit follow-up 😁

As mentioned in https://github.com/medic/cht-docs/issues/900#issuecomment-1363401069 I think we should have a single favicon that better well with a variety of tab colors.

craig-landry:
@abbyad From our video conversation about this, did you end up finding a single icon that works for dark/light? I'd like to close this if we don't intend on using it.

abbyad:
I have a better one but haven't committed it yet. Do you want that new icon as part of this PR, or a wholly new one? Either way it still relates to https://github.com/medic/cht-docs/issues/924.

craig-landry:
Let's just do it as a separate PR. I'm going to close this one and will take a look at your new one when you open it.

PR #926, opened 2022-12-23
Update 4.0 Upgrade Guide to note `+` operator does not work with strings

jkuester:
This usage of the `+` operator was previously unknown to me and is out-of-spec.  However, it did work with older versions of Enketo. So, I have added a brief note to the upgrade guide to call out the behavior change.

_(This change was first noted by @binokaryg (and team) when upgrading a project's config to work with CHT `4.x`.)_


mrjones-plip (approved):
🏭 (thanks rando [rando](https://randomoutputs.com/random-emoji-generator) site!)

PR #927, opened 2023-01-04
Make favicon suitable for dark and light modes

abbyad:
Uses the negative logo with a gradient similar to the original CHT logo so that it stands out on different colored tabs.

For #924 and #900


abbyad:
From my quick test, the new favicon on the right seems clear enough on dark and light tabs:

image

I'm sure we could improve on it further, but figure this is better than the current option, and we can revisit later as needed.

craig-landry:
Looks good. While doing screenshots and opening the current prod site for comparison I also realized another state (current tab vs not). The new icon looks better in that case as well.

New logo on left and old on right in each screenshot:

image

image

craig-landry (approved):

abbyad:
Thanks for confirming, and with the additional states!

PR #940, opened 2023-01-13
Initial simple GitHub codeowners setup

craig-landry:
The goal here is to align parts of the handbook with people on the team who should be paying special attention. Different changes need different levels of attention.

This first file sets up a small group at the top level of the docs and the has a level set up where access rights are expanded to product management.

As we add more content to the handbook it is expected that new GitHub groups would be set up and those groups referenced here and aligned with specific areas of documentation.


craig-landry:
@andrablaj I don't know a lot about the syntax for CODEOWNERS. Their [docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners) give some nice examples, but I'm still unclear on some cases where we might want to use a glob-link syntax or even to know exactly how the nested permissions works. If you know or have a good guess, I'm happy to make any changes here. Otherwise it seems easier for now to just guess and work it out as we use the features. Nobody's trying to do anything nefarious with all these docs.

craig-landry:
Oh and I should note that I did create the groups referenced here (one already existed actually) but I have not gone to set up any corresponding [branch protection](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-and-branch-protection). We can try that out together once we have a version of this that looks like it'll work.

.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# Default ownership to "Product Direction"
+/content/en/contribute/medic @medic/product-direction
```
andrablaj, line 2: Could we call the group `product-leadership`? It might be me, but I find the `direction` word confusing.

andrablaj (approved):
Thanks Craig! I don't have in-depth knowledge about `CODEOWNERS`, and this seems like a good start. We can try it out and iterate with improvements later.

.github/CODEOWNERS
craig-landry: I had that originally, but considered other options for a hierarchy with more or less people in the group. Given your feedback though, that sounds more confusing than helpful. I've updated this and the corresponding group in GitHub.

andrablaj (approved):

PR #941, opened 2023-01-13
Some cleanup of Baserow bits on handbook

craig-landry:
* Updated a few areas to make things a bit more generic and not reference specific projects or names.
* Fixed a few links.
* Moved a video demo to Google Drive to avoid any issues around content privacy in what was shown.
* Removed baserow feedback page as we'll just do that process elsewhere.


andrablaj (approved):

craig-landry:
I had Michael on here, but I think the changes I've made are pretty small. If his feedback requires changes I'll just go make another PR real quick. For now I'd rather merge this and keep moving.

PR #948, opened 2023-01-23
Add files via upload

nsitaula:
To update the CHT logo in the documentation [Building CHT Android Flavors](https://docs.communityhealthtoolkit.org/apps/guides/android/branding/#1-required-resources)


mrjones-plip (approved):
looks good - thanks for taking the initiative to change this!

approving, but likely would be better if cropped it closer so the image and text took up more of the 512x512 logo:

image

nsitaula:
> looks good - thanks for taking the initiative to change this!
> 
> approving, but likely would be better if cropped it closer so the image and text took up more of the 512x512 logo:
> 
> image



> looks good - thanks for taking the initiative to change this!
> 
> approving, but likely would be better if cropped it closer so the image and text took up more of the 512x512 logo:
> 
> image

thanks, @mrjones-plip

PR #949, opened 2023-01-25
Add docs for creating user with create_user_for_contacts trans

jkuester:
~~Note that the weird `{{% tab ... %}}` stuff if just [Docsy's Tabbed Panes](https://www.docsy.dev/docs/adding-content/shortcodes/#tabbed-panes).  We don't have to use it if folks don't like it here, but seemed worth a try....~~

image

https://github.com/medic/cht-core/issues/7753

[cht-core PR](https://github.com/medic/cht-core/pull/7988)

### DIFF
--- a/content/en/apps/reference/app-settings/transitions.md
+++ b/content/en/apps/reference/app-settings/transitions.md
@@ -54,8 +54,7 @@ The following transitions are available and executed in order.
 | [muting](#muting) | Implements muting/unmuting actions of people and places. Available since 3.2.x. Is partially applied on the client, as of 3.12.0. |
 | [mark_for_outbound]({{% ref "apps/reference/app-settings/outbound" %}}) | Enables outbound pushes. Available since 3.5.x |
 | [self_report](#self_report) | Maps patient to sender. Available since 3.9.x |
-| [create_user_for_contacts](#create_user_for_contacts) | Triggers an existing offline user to be replaced with a new user on the same device (without needing to immediately sync with the server). Available since 4.1.x  | |
-
+| [create_user_for_contacts](#create_user_for_contacts) | Allows for automatically creating or replacing users based on data from their associated contact. Available since 4.1.x  | |
 ## Transition Configuration Guide
 
 Guides for how to setup specific transitions.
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/app-settings/transitions.md
```diff
@@ -676,61 +675,106 @@ If this configuration is not set then the message defaults to what is set in the
...
+
+#### Configuration
+
+Several configurations are required in `app_settings` to enable the `create_user_for_contacts` transition.
+
+[Login by SMS]({{< ref "apps/reference/api#login-by-sms" >}}) must be enabled by setting the `token_login` configuration.
+
+The [`app_url` property]({{< ref "apps/reference/app-settings#app_settingsjson" >}}) must be set to the URL of the application. This is used to generate the token login link for the new user.
+
+##### Example
```
mrjones-plip, line 688: love this example here to compliment the deep links to the docs - thank you!

mrjones-plip (changes requested):
Oh - neat - tabs!  I kinda like it, but think we should not do it for three reasons, in order of importance:
1. it hides the two tabbed sections from the table of contents on the right, and this disables deep linking to a specific tab
2. it's a new UI design pattern that we should vet first before using 
3. it makes the page require JS to be enabled (ok, you can ignore this one, that's just a mrjones thang!)

I don't feel THAT strongly about item 2 though, so could maybe swayed on number 1?  Lemme know!

Yours as is in PR (oh - ooops I shouldn't have changed header level of example, I see that's under config correctly - sorry!):

image

Mine, edited to not have tabs, change header levels (again - I shouldn't have changed header level of example, I see that's under config correctly - sorry!):

image

Oh wait!  There are link inline here:

image

But when I open that URL in a new page, the anchor isn't rendered in the DOM as it's hidden in the unselected "Replace User" tab , so better, but still sub optimal.  I still say skip it for now!

jkuester:
@mrjones-plip I agree 100% with your points from `#1`.  Looking at this again, I think it is a bad idea to nest _a lot_ of content inside tabs for precisely what you mentioned.  I think tabs can be very useful for displaying a smaller amount of content within a particular section (but probably once you start adding new sub-headers inside the tabs you have too much in the tab...)

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -676,58 +675,96 @@ If this configuration is not set then the message defaults to what is set in the
...
+  "translation_key": "sms.token.login.help"
+},
+"transitions": {
+  "create_user_for_contacts": true
+}
+```
+
+#### Create User
+
+When adding a new contact, the `create_user_for_contacts` transition can be triggered to create a new user associated with that contact. _Available since 4.2.x._
```
mrjones-plip, line 702: I think this is missing a noun, and I think that noun is "form", but I'm not sure!

```suggestion
When adding a new contact form, the `create_user_for_contacts` transition can be triggered to create a new user associated with that contact. _Available since 4.2.x._
```

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -676,58 +675,96 @@ If this configuration is not set then the message defaults to what is set in the
...
+
+##### Example scenario
+
+This functionality would allow a supervisor to add a new contact for a new CHW and have a user for the new CHW automatically be created with the data from the new contact.
```
mrjones-plip, line 706: Re-wording for clarity:

```suggestion
Allow a supervisor to onboard a CHW by creating both a new contact and a new user in an existing place with the data from a single "create contact" form.
```

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -676,58 +675,96 @@ If this configuration is not set then the message defaults to what is set in the
...
+
+Once the new contact is synced with the server and has been processed by Sentinel, the new CHW will receive an SMS message (at the phone number specified in the contact) containing a token login link. This link will allow them to login as the newly created user.
```
mrjones-plip, line 708: Remind folks of limitations of login links:

```suggestion
Once the new contact is synced with the server and has been processed by Sentinel, the new CHW will receive an SMS message (at the phone number specified in the contact) containing a token login link. This link will allow them to login as the newly created user. For security reasons, the token login link is valid for only one use and can only be used within 24 hours.
```

content/en/apps/reference/app-settings/transitions.md
```diff
@@ -676,58 +675,96 @@ If this configuration is not set then the message defaults to what is set in the
...
+
+##### Form Configuration
+
+When the `create_user_for_contacts` transition is enabled, Sentinel will attempt to create a user for any _newly created_ `person` contacts with the `user_for_contact.create` field set to `'true'`. So, `contact` forms and `app` forms for adding persons that should trigger new user creation need to include a `user_for_contact` group that contains a `create` field. The calculation for the value of the `create` field should evaluate to `'true'` when a new user is desired.  Any other value for that field will not trigger user creation.
```
mrjones-plip, line 712: Is this a truthy  `'true'` string or a boolean `true`?  I don't think it much matters, but I was reading it closely and figured I'd ask!

mrjones-plip (approved):
This is great work - thanks @jkuester ! 

I did a deeper dive on content and found some minor points to revise, but nothing that should block going live.  

As such, approving so you can amend as you see fit.

content/en/apps/reference/app-settings/transitions.md
jkuester: Yeah, good question!  It is a `'true'` string because that is how Enketo (ODK?) serializes boolean values.  Not sure exactly why Enketo outputs string values instead of real booleans (which are supported in JSON), but this follows existing patterns we have in other places and keeps the form implementation as simple as possible, so seems okay.

content/en/apps/reference/app-settings/transitions.md
jkuester: I think this should actually read `person contact` since it is the person that triggers the transition (and not the form).  But let me know if you think this just needs to be completely re-worded for more clarity...

content/en/apps/reference/app-settings/transitions.md
jkuester: I have made some additional tweaks and edits based on your suggestion here.  Let me know if it still sounds confusing!

mrjones-plip: new contents looks great - thanks!

mrjones-plip (approved):
:1st_place_medal:

PR #952, opened 2023-02-01
Update local-setup.md

Phil-Mwago:
For more context, we had this issue here on the forum: https://forum.communityhealthtoolkit.org/t/unable-to-push-and-upload-a-blank-project-on-my-local-environment/2427/8

My local set up has issues deploying a blank project on the cht-core location. We replicated this in other local deployments.

### DIFF
--- a/content/en/apps/tutorials/local-setup.md
+++ b/content/en/apps/tutorials/local-setup.md
@@ -162,7 +162,6 @@ cht --url=https://medic:password@localhost --accept-self-signed-certs
 ```
 
 If the above command shows an error similar to this one `ERROR Error: Webpack warnings when building contact-summary` you will need to install all the dependencies and libraries it needs, then you need to restart the docker-compose and try again. Run the following to install the dependancies.
-If this error is persistent, please create a new directory and initialise a new project in the new location (outside of your `cht-core` directory tree).
 
 ```shell
 npm ci
--- a/content/en/apps/tutorials/local-setup.md
+++ b/content/en/apps/tutorials/local-setup.md
@@ -160,7 +160,8 @@ Then deploy the blank project onto your local test environment with the command:
 cht --url=https://medic:password@localhost --accept-self-signed-certs
 ```
 
-If the above command shows an error similar to this one `ERROR Error: Webpack warnings when building contact-summary` you will need to install all the dependencies and libraries it needs, then you need to restart the docker-compose and try again.
+If the above command shows an error similar to this one `ERROR Error: Webpack warnings when building contact-summary` you will need to install all the dependencies and libraries it needs, then you need to restart the docker-compose and try again. Run the following to install the dependancies.
+If this error is persistent, please create a new directory and initialise a new project in the new location (this location should not be in your cht-core folder).
 
 ```shell
 npm ci
--- a/content/en/apps/tutorials/local-setup.md
+++ b/content/en/apps/tutorials/local-setup.md
@@ -162,7 +162,7 @@ cht --url=https://medic:password@localhost --accept-self-signed-certs
 ```
 
 If the above command shows an error similar to this one `ERROR Error: Webpack warnings when building contact-summary` you will need to install all the dependencies and libraries it needs, then you need to restart the docker-compose and try again. Run the following to install the dependancies.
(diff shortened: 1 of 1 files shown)


content/en/apps/tutorials/local-setup.md
```diff
@@ -160,7 +160,8 @@ Then deploy the blank project onto your local test environment with the command:
 cht --url=https://medic:password@localhost --accept-self-signed-certs
 ```
 
-If the above command shows an error similar to this one `ERROR Error: Webpack warnings when building contact-summary` you will need to install all the dependencies and libraries it needs, then you need to restart the docker-compose and try again.
+If the above command shows an error similar to this one `ERROR Error: Webpack warnings when building contact-summary` you will need to install all the dependencies and libraries it needs, then you need to restart the docker-compose and try again. Run the following to install the dependancies.
+If this error is persistent, please create a new directory and initialise a new project in the new location (this location should not be in your cht-core folder).
```
craig-landry, line 164: I'm tempted to say that we should make this a code check in cht-conf, but this change really just adds detail to the above line that's already here. Given that, it seems fine for now.

Slight wording suggestion:
```suggestion
If this error is persistent, please create a new directory and initialise a new project in the new location (outside of your `cht-core` directory tree).
```

I'll approve this now to unblock. If you like your original wording more, it's ok to ignore my suggestion and go with whatever you prefer.

craig-landry (approved):

mrjones-plip (changes requested):
I think this PR is going in the wrong direction.  We already [have explicit code](https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/#4-create-and-upload-a-blank-project) to create a new directory:

```
mkdir cht-app-tutorials
cd cht-app-tutorials
cht initialise-project-layout
```

It's confusing to add content at the very bottom about creating a new directory which already exists. Instead, I think we should more strongly call out the importance to create the directory in the current instructions .  

Two approaches to this might be:
1. change the `mkdir` call to be `mkdir ~/cht-app-tutorials` and then at the top of every sample code add `cd ~/cht-app-tutorials`. This way when it's copied and pasted folks will always be in the right directory. I did this quite a bit in the [multi-node steps](https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/self-hosting-multiple-nodes/#starting-services). While repetitive, it's nice to do to ensure a happy path when something seems possibly overwhelmingly technical to new CHT users.
2. literally add instructions at the top of step 4 that call this out.  Maybe something like: 

   > With `cht-conf` you can also create a blank project. This provides you a template from which you can begin working on CHT. To do so, run the following commands which will create a new directory and initialize it.  Be sure to use this directory for all subsequent calls:

esthermmoturi:
@Phil-Mwago I think the best approach would be to just add the comment below as @mrjones-plip has suggested.
>> With cht-conf you can also create a blank project. This provides you a template from which you can begin working on CHT. To do so, run the following commands which will create a new directory and initialize it. Be sure to use this directory for all subsequent calls:

esthermmoturi (approved):
@Phil-Mwago This looks okay. We will keep updating the documentation as need arises.

esthermmoturi (approved):

content/en/apps/tutorials/local-setup.md
```diff
@@ -160,7 +161,8 @@ Then deploy the blank project onto your local test environment with the command:
...
```
mrjones-plip, line 165: This isn't needed now and is still confusing as it feels like a different directory than in step 4. let's remove it:

```suggestion
```

mrjones-plip (changes requested):
Nice work! I think the updates to step 4 are enough and we should remove the sentence below.

content/en/apps/tutorials/local-setup.md
mrjones-plip: I still think this entire sentence should be removed.

mrjones-plip (changes requested):
Please remove the sentence I cited in my review.

mrjones-plip (approved):
per our call - you can remove sentence so I'll approve now - nice work!

@mrjones-plip This is done.

PR #954, opened 2023-02-14
Multi-node install docs: Fix couchdb node names, use compose overrides 

mrjones-plip:
This PR:
* removes duplicate network declaration of both `cht-net` and `cht-overlay` in compose files.  Keeps it to just one (`cht-net`) and just makes that an overlay network.  This reliably fixes  #medic/cht-core/issues/8050
* instead of having the user manually editing compose files, uses the [concept of compose override files](https://docs.docker.com/compose/extends/). this means users no longer need to manually edit specific parts of files and set up is MUCH less error prone
* ensures that users can just copy and paste commands and it should Just Work :tm:  for a 3 node setup

per #medic/cht-core/issues/8050

### DIFF
--- a/content/en/apps/guides/hosting/4.x/self-hosting-multiple-nodes.md
+++ b/content/en/apps/guides/hosting/4.x/self-hosting-multiple-nodes.md
@@ -65,7 +65,7 @@ To add a manager to this swarm, run 'docker swarm join-token manager' and follow
 Then create overlay network by calling:
 
  ```
- docker network create --driver=overlay --attachable cht-overlay
+ docker network create --driver=overlay --attachable cht-net
  ```
 
 ### CouchDB nodes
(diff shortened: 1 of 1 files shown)


mrjones-plip:
@garethbowen  - with SRE down to one person for the coming weeks, maybe you're free to review this?

content/en/apps/guides/hosting/4.x/self-hosting-multiple-nodes.md
```diff
@@ -191,85 +184,74 @@ Note that secure passwords and UUIDs were generated and saved in the resulting `
...
-{{< highlight yaml "linenos=table" >}}
+```shell
+cd /home/ubuntu/cht/
+curl -s -o ./docker-compose.yml https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:4.0.1/docker-compose/cht-couchdb.yml
+```
+
+Now create the override file to have Node 1 join the `cht-net` overlay network we created above. As well, we'll set some `services:` specific overrides:
+
+```shell
+cat > /home/ubuntu/cht/network-overrides.yml << EOF
```
garethbowen, line 197: The file has more than just network overrides so maybe we should call it `custom-overrides.yml`?

garethbowen (approved):
Very nice. Thanks! One minor suggestion inline.

mrjones-plip:
ah - good point! I went with  `cluster-overrides.yml` as `custom` and `overrides` seemed reduntant, see 72591ea

mrjones-plip:
ah - I see you approved already - thanks! merge away!!

PR #956, opened 2023-02-14
add cht4 docker logs support script 

mrjones-plip:
This PR ads instructions for how to use the recent [compress_and_archive_docker_logs.sh](https://github.com/medic/cht-core/blob/master/scripts/compress_and_archive_docker_logs.sh) script. 

While tempting to put it with the existing [Obtaining Logs](https://docs.communityhealthtoolkit.org/apps/guides/debugging/obtaining-logs/) page, creating a stand alone page with the intent to share logs seemed more appropriate. 

per #medic/cht-core/pull/8061


esthermmoturi (approved):
LGTM!

PR #958, opened 2023-02-15
Add 4.x logs and backup info, small tweaks elsewhere 

mrjones-plip:
This PR adds two new docs:
* 4.x Backups
* 4.x Logs

As well, it makes some minor tweaks to other related docs to order the TOC on the left, delineate between phone and browser logs and server logs

per #831

### DIFF
--- a/content/en/apps/guides/hosting/4.x/backups.md
+++ b/content/en/apps/guides/hosting/4.x/backups.md
@@ -17,7 +17,7 @@ As CHT 4.x uses a container per service, the only data that needs to be backed u
 * Docker Compose and `.env` files 
 * TLS certificates
 
-This is because Docker containers are inherently stateless so all the important binaries are already stored in [CHT's Docker images](https://gallery.ecr.aws/s5s3h4s7/).  Docker Compose files, including the `.env` file, store all of your deployment's configuration.  Finally, the TLS certificates should be backed up to ease recover speed.
+This is because Docker containers are inherently stateless so all the important binaries are already stored in [CHT's Docker images](https://gallery.ecr.aws/s5s3h4s7/).  Docker Compose files, including the `.env` file, store all of your deployment's configuration.  Finally, the TLS certificates should be backed up to reduce recovery time.
 
 How to backup each of these three pieces of data is covered below.
 
--- a/content/en/apps/guides/hosting/4.x/backups.md
+++ b/content/en/apps/guides/hosting/4.x/backups.md
@@ -59,30 +59,24 @@ You should have SSH access to the server with `root` access.
 
 ### Backup software
 
-It's assumed you are using which ever tool you're familiar with which might include [rsync](https://rsync.samba.org/examples.html), [borg](https://borgbackup.readthedocs.io/en/stable/) or other solution.  The locations of the backups should follow the 3-2-1 rule:
+It's assumed you are using which ever tool you're familiar with which might include [rsync](https://rsync.samba.org/examples.html), [borg](https://borgbackup.readthedocs.io/en/stable/), [duplicity](https://duplicity.gitlab.io/) or other solution.  The locations of the backups should follow the 3-2-1 rule:
 
 > There should be at least 3 copies of the data, stored on 2 different types of storage media, and one copy should be kept offsite, in a remote location. _- [Wikipedia](https://en.wikipedia.org/wiki/Backup)_
+
+Duplicity has the handy benefit of offering built in encryption using [GPG](https://gnupg.org/). Consider using this if you don't have an existing solution for encrypted backups. 
 
 ## CouchDB
 
 {{% alert title="Note" %}}
-CouchDB backups, by necessity, will have PII and PHI.  They should be safely stored to prevent unauthorized access.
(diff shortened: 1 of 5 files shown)


content/en/apps/guides/hosting/4.x/backups.md
```diff
@@ -0,0 +1,123 @@
+---
+title: "Backups in CHT 4.x"
+linkTitle: "Backups"
+weight: 330
```
garethbowen, line 4: Very specific!

content/en/apps/guides/hosting/4.x/backups.md
```diff
@@ -0,0 +1,123 @@
...
+
+## Introduction 
+
+As CHT 4.x uses a container per service, the only data that needs to be backed up is:
+
+* CouchDB database
+* Docker Compose and `.env` files 
+* TLS certificates
+
+This is because Docker containers are inherently stateless so all the important binaries are already stored in [CHT's Docker images](https://gallery.ecr.aws/s5s3h4s7/).  Docker Compose files, including the `.env` file, store all of your deployment's configuration.  Finally, the TLS certificates should be backed up to ease recover speed.
```
garethbowen, line 20: Slight grammatical improvement?

```suggestion
This is because Docker containers are inherently stateless so all the important binaries are already stored in [CHT's Docker images](https://gallery.ecr.aws/s5s3h4s7/).  Docker Compose files, including the `.env` file, store all of your deployment's configuration.  Finally, the TLS certificates should be backed up to reduce recovery time.
```

content/en/apps/guides/hosting/4.x/backups.md
```diff
@@ -0,0 +1,123 @@
...
+### Backup software
+
+It's assumed you are using which ever tool you're familiar with which might include [rsync](https://rsync.samba.org/examples.html), [borg](https://borgbackup.readthedocs.io/en/stable/) or other solution.  The locations of the backups should follow the 3-2-1 rule:
+
+> There should be at least 3 copies of the data, stored on 2 different types of storage media, and one copy should be kept offsite, in a remote location. _- [Wikipedia](https://en.wikipedia.org/wiki/Backup)_
+
+## CouchDB
+
+{{% alert title="Note" %}}
+CouchDB backups, by necessity, will have PII and PHI.  They should be safely stored to prevent unauthorized access.
```
garethbowen, line 69: I think we can expand on this and suggest that backups are encrypted. Even if we don't go into details on _how_ to do it we should be advocating for best practice.

content/en/apps/guides/hosting/4.x/backups.md
```diff
@@ -0,0 +1,123 @@
...
+{{% /alert %}}
+
+Assuming your CouchDB is stored in `/home/ubuntu/cht/couchdb`, you should use these steps to back it up:
+
+1. Stop CouchDB.  This ensures when you back up the data files, you have a complete copy of all data in a known good state:
```
garethbowen, line 74: I don't think this is necessary, and we want people to take backups regularly, so will be highly disruptive. Official documentation here: https://docs.couchdb.org/en/stable/maintenance/backups.html

content/en/apps/guides/hosting/4.x/backups.md
```diff
@@ -0,0 +1,123 @@
...
+As all of these files are only read when Docker first loads a service, you can simply copy them whenever you want without stopping any of the CHT services.  They should be copied with the same frequency and put in the same location as the CouchDB data using the [backup software specified above]({{< relref "#backup-software" >}}).
+
+
+## TLS certificates
+
+{{% alert title="Note" %}}
+The `.key` file is the private key for TLS certificate.  It should be safely stored to prevent unauthorized access.
+{{% /alert %}}
+
+Like the compose files, the TLS certificate files can easily be regenerated or re-downloaded from you Certificate Authority, like Let's Encrypt for example. However, you want to have a backup of the at the ready to ease the restore process.
```
garethbowen, line 110: Typo.

```suggestion
Like the compose files, the TLS certificate files can easily be regenerated or re-downloaded from your Certificate Authority, like Let's Encrypt for example. However, you want to have a backup of the at the ready to ease the restore process.
```

garethbowen (changes requested):
This is great! Thank you.

I've left a few suggestions inline.

content/en/apps/guides/hosting/4.x/backups.md
mrjones-plip: Hah! I just wanted to put the self hosted guides up top and the logs and backup on the bottom.  I hit with the `+10` which didn't work, so I hit it with the `+100` which finally did the trick ; )

content/en/apps/guides/hosting/4.x/backups.md
mrjones-plip: good thinking! fixed in 3f1fd4f

content/en/apps/guides/hosting/4.x/backups.md
mrjones-plip: good thinking! fixed in 3f1fd4f

garethbowen (approved):
Awesome, thanks!

PR #960, opened 2023-02-18
fix: small grammatical error

sookwalinga:


garethbowen (approved):
Thanks @sookwalinga !

PR #962, opened 2023-02-21
update related and intro on 4.x index

mrjones-plip:


mrjones-plip:
@abbyad  - if you're interested and have more time, open to having this be the invite to do a more holistic review of [the 4.x docs](https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/) too!  Either in this PR or to open separate fix it tickets against me.

abbyad (changes requested):
Apologies for the late review!

Adding the related info is useful, and I feel we can do more to introduce this page/section and improve readability. For instance, all the sentences currently link out to other pages, rather than explain what this section is about, so I would add an intro paragraph. Also, I would suggest avoiding starting with a note, unless it is absolutely a critical pre-read. Using a similar page as an example, having a note to start makes this page hard to read, not knowing where to look. 
image

While on the topics of readability, it might be worth reusing existing patterns from the docs site, like the "Related Content" section, and the "See Also" tag, since that may make it easier to read through and identify the content.

mrjones-plip:
Thanks @abbyad !  

Rethinking my invitation to you for general input, I propose approving & merging this PR. 

After that, @nsitaula,  @esthermmoturi,  you and I meet and plan on more holistic reviews of the new bifurcated 4x vs 3.x content in search of consistency and contextual improvements like you mentioned. 

I'll assume you agree and schedule a short brainstorm meeting for next week! But lemme know if that doesn't work.

abbyad (approved):
After discussion with @mrjones-plip, @esthermmoturi, and @nsitaula, we decided to make this page match it's 3.x partner, and then work on the larger improvements separately.

PR #968, opened 2023-02-27
rename couchdb node name, fix yml name typo, bump cht 4.0.1 -> 4.1.0 in curl

mrjones-plip:
This PR:
* bumps all `curl` requests fetching docker compose files to use `4.1.0` instead of `4.0.1`
* fixes a misnamed `yml` file for the override in the 2nd couchdb node
* renames `couchdb.x` -> `couchdb-x.local` per #medic/cht-core/issues/7962

### DIFF
--- a/content/en/apps/guides/hosting/4.x/self-hosting-multiple-nodes.md
+++ b/content/en/apps/guides/hosting/4.x/self-hosting-multiple-nodes.md
@@ -216,7 +216,7 @@ Like we did for Node 1, create `/home/ubuntu/cht/docker-compose.yml` and the `cl
 ```shell
 cd /home/ubuntu/cht/
 curl -s -o ./docker-compose.yml https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:4.1.0/docker-compose/cht-couchdb.yml
-cat > /home/ubuntu/cht/network-overrides.yml << EOF
+cat > /home/ubuntu/cht/cluster-overrides.yml << EOF
 version: '3.9'
 services:
   couchdb:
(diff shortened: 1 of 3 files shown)


garethbowen (approved):
Great, thanks!

We might need to find a more sustainable way to update the versions...

mrjones-plip:
thanks mate!

> We might need to find a more sustainable way to update the versions...

We could use the [logic you pushed for](https://github.com/medic/cht-core/blob/5d0d4bc47a55cc28821a8e0b8e033c9fdd001cb7/scripts/docker-helper-4.x/cht-docker-compose.sh#L50) in the docker helper! we'd go from this:

```bash
curl -s -o ./docker-compose.yml https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:4.1.0/docker-compose/cht-couchdb.yml
``` 

to this:

```bash
latest=$(curl -s https://staging.dev.medicmobile.org/_couch/builds_4/_design/builds/_view/releases\?limit\=1\&descending\=true |  tr -d \\n | grep -o 'medic\:medic\:[0-9\.]*')
curl -s -o ./docker-compose.yml  https://staging.dev.medicmobile.org/_couch/builds_4/${latest}/docker-compose/cht-couchdb.yml
```

we already use `curl`, so maybe `tr` and `grep` aren't that much more to ask given it's already successfully been used in docker helper?

garethbowen:
Because this is a manual process we don't necessarily need to script it. One way would be to just explain how to find the latest (either with curl, or linking another page), and then update all the commands with a variable that we expect humans to replace. The obvious downside is it's no longer a copy and paste command. One benefit is for people who have a known version they need to go to (eg: testing something that will one day go to prod on a known release, or feature releases, etc) can follow the same instructions.

Aside... we should add a page to staging.dev which shows available versions in a human readable way.

Issue #967, opened 2023-02-27
Document preferred programming language


garethbowen:
We don't currently have any guidance about what programming languages to use. The current front running is for JS (including Node) everywhere, with the exception for TypeScript in the webapp.

Clearly document the best practice, provide justification, and outline exceptions.

garethbowen:
Merged.

PR #972, opened 2023-03-02
feat(#967): be opinionated about languages 

garethbowen:


garethbowen:
Review please @craig-landry 

This one was more controversial but seemed to get consensus in the end with the update to include TS as a first class language.

content/en/contribute/code/style-guide.md
```diff
@@ -6,25 +6,38 @@ description: >
...
+## Language
+
+Our strong preference is to use JavaScript or TypeScript wherever possible, including in webapps, on the server, and for scripting. This is because:
+
+- every developer on the team knows it already, so we can all maintain it together,
+- it makes it easy to write cross-platform code and we have developers on all major operating systems,
+- it has a vast number of libraries that are easy to include, and
+- it's easy to unit test
+
+Exceptions to this can be made on a case by case basis, but the decision must be made collectively before coding has begun to avoid having to rewrite.
```
craig-landry, line 18: Just a little grammatical suggestion:

```suggestion
Exceptions to this can be made on a case-by-case basis, but the decision must be made collectively before coding has begun to avoid having to rewrite.
```

content/en/contribute/code/style-guide.md
```diff
@@ -6,25 +6,38 @@ description: >
   Guidelines for writing code
 ---
 
+## Language
+
+Our strong preference is to use JavaScript or TypeScript wherever possible, including in webapps, on the server, and for scripting. This is because:
```
craig-landry, line 11: Small wording change. My thinking here is that rather than state "this is how we do it", which someone could read and think, "ok, whatever, that's nice, you have a nice time with that" the suggested wording is more instructive on what we are asking you to do.

```suggestion
Prefer JavaScript or TypeScript wherever possible, including in webapps, on the server, and for scripting. This is because:
```

craig-landry (changes requested):
Aside from my wording suggestions, this looks good to me.

Also, I take this is a first change to be clear on the subject in general. Let's talk about what we can follow with to give more specific guidance on when TypeScript is appropriate or not and how engineers can feel informed when choosing.

garethbowen:
@craig-landry Thanks for the suggestions!

> Also, I take this is a first change to be clear on the subject in general.

Definitely. To be continued...

craig-landry (approved):

PR #973, opened 2023-03-02
make docker-compose.yml default , remove all but first ref of  docker-compose-developer-3.x-only.yml

mrjones-plip:
This PR fixes the repetitive user of `docker-compose-developer-3.x-only.yml` and standardizes on `docker-compose.yml`

The  `docker-compose-developer-3.x-only.yml`  file was created when we released CHT 4.0.0 and we didn't want to confuse people.  However, the instructions as is don't work as the first `curl` uses the URL of the `3.x-only` file but saves it  `docker-compose.yml`.  By removing the long name all the commands are simplified.


garethbowen (approved):
Nice simplification. Thanks!

garethbowen:
Not sure what's going on with the build but maybe related to: https://github.com/medic/cht-docs/issues/975 ?

mrjones-plip:
> Not sure what's going on with the build but maybe related to: #975 ?

While #975 definitely needs some love, the CI break on this ticket was just a hiccup. I kicked it and it ran fine

PR #974, opened 2023-03-03
Table format for core competencies

craig-landry:
The basic heading+description format was fine, but the Hugo template adds an H2 "Feedback" item to the end which at first is a bit confusing as it looks like it's a 6th core competency of "feedback". This format change makes it much more clear what on the page is actually for what.

Here are screenshots showing the formatting confusion and the proposed visual change to avoid it:
image
image


latin-panda (approved):
The content is clear, thanks!

PR #983, opened 2023-03-13
feat: add documentation for extension libs

garethbowen:
medic/cht-core#7824

### DIFF
--- a/content/en/apps/reference/extension-libs.md
+++ b/content/en/apps/reference/extension-libs.md
@@ -25,7 +25,7 @@ return function(/* parameters */) {
 }
 ```
 
-Now populate the function as needed. For complex functions it is recommended to use third party libraries and use a bundler (eg: [webpack](https://webpack.js.org/)) to make it easy to build a single file. For anything non-trivial, it's recommended to use development best practices such as linting, unit tests, and minification.
+Now populate the function as needed. For complex functions third party libraries (such as momentjs, lodash, etc) and use a bundler (eg: [webpack](https://webpack.js.org/)) to make it easy to build a single file. It's recommended to use development best practices such as linting, unit tests, and minification to ensure quality and small download size.
 
 #### xpath functions
 
--- a/content/en/apps/reference/extension-libs.md
+++ b/content/en/apps/reference/extension-libs.md
@@ -25,7 +25,7 @@ return function(/* parameters */) {
 }
 ```
 
-Now populate the function as needed. For complex functions third party libraries (such as momentjs, lodash, etc) and use a bundler (eg: [webpack](https://webpack.js.org/)) to make it easy to build a single file. It's recommended to use development best practices such as linting, unit tests, and minification to ensure quality and small download size.
+Now populate the function as needed. For complex functions, it is recommended to use third party libraries (such as momentjs, lodash, etc) and use a bundler (eg: [webpack](https://webpack.js.org/)) to make it easy to build a single file. It's recommended to use development best practices such as linting, unit tests, and minification to ensure quality and small download size.
 
 #### xpath functions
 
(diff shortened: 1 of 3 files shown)


content/en/apps/reference/extension-libs.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+### Uploading to the CHT
+
+Create a folder within the project configuration to contain the libraries, for example:
+
+```
+./extension-libs
+    average.js
+    calculate-risk-score.js
```
jkuester, line 74: Random question, but as I looked at this, the first thing that popped into my head is what code-reuse would look like in this process.  Say I have a `utils.js` file that contains code that is common between `average.js` and `calculate-risk-score.js`.  Even if I am using webpack to create average/calculate-risk-score so that the `utils.js` code is included in those files, does that mean that the code is going to be duplicated in the extension attachments (and ultimately duplicated in webapp)?  With separate files for each function are we forcing all re-used code to be duplicated?

If so, I wonder if it would be better to just require that all extension functions come from a single file.  (Alternatively, would it make more sense to have app devs include all their code in the `/extension-libs` folder and then cht-conf can handle webpacking everything before attaching it?)

content/en/apps/reference/extension-libs.md
```diff
@@ -0,0 +1,89 @@
...
+
+Now run the command: `cht upload-extension-libs`
+
+This will create or update a document to CouchDB with an ID of `extension-libs` with each of the configured scripts attached. Once this is created the webapp service worker will be updated so the libraries are cached on the phone ready for use offline.
+
+### Invoking the function
+
+#### CHT API
+
+The function will now be available via the CHT API for [tasks]({{< ref "tasks" >}}), [tasks]({{< ref "targets" >}}), and [contact summary]({{< ref "contact-page" >}}) configurations.
```
jkuester, line 85: > CHT API

Can you include a link to the `_partial_cht_api` page?  (Would be helpful clarification in case someone is thinking of the REST api.)

jkuester (approved):
Couple minor suggestions and then one question about code-reuse between functions.

content/en/apps/reference/extension-libs.md
```diff
@@ -0,0 +1,89 @@
...
+
+The first step is to create the js file to return the function that will be called by the web application. Create a new file using this template:
+
+```js
+return function(/* parameters */) {
+  return result;
+}
+```
+
+Now populate the function as needed. For complex functions it is recommended to use third party libraries and use a bundler (eg: [webpack](https://webpack.js.org/)) to make it easy to build a single file. For anything non-trivial, it's recommended to use development best practices such as linting, unit tests, and minification.
```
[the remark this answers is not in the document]
garethbowen: Updated.

content/en/apps/reference/extension-libs.md
garethbowen: Part of the idea is to keep functions separate so if you change one then users only need to download the new one rather than the whole bundle again. An example for this is including the momentjs library so you can use it in the tasks config.

You're right though, if you also want to call momentjs from your risk score calculator then you'd introduce duplication. How do you feel about making the script api available to the extension libs, so you can get a lib fn from inside a lib fn?

content/en/apps/reference/extension-libs.md
garethbowen: It looks a bit weird to link to a partial. Originally I included the partial in the page but it was cluttered and mostly irrelevant.

It might make more sense to create a new CHT API page which has a brief intro, then includes the partial. Then I could link to that from here. What do you think?

garethbowen:
> (Also, if this is the direction we want to go, it seems perfectly feasible and passive to add this functionality in later. We could push what we have now if anyone is waiting on it and then come back later to add the script api to the extension-libs....)

That was my thinking too. I'll raise an issue for this addition.

PR #995, opened 2023-03-16
Add test feedback template

tatilepizs:
Just adding a template to use once the test is complete to provide feedback to the developers.
cc: @lorerod @ngaruko

### DIFF
--- a/content/en/contribute/code/workflow.md
+++ b/content/en/contribute/code/workflow.md
@@ -187,6 +187,80 @@ Issues in this column are in the process of being acceptance tested by a Quality
    1. How was the issue verified? Console logs, page source, errors anywhere, execution steps.  LGTM should begin to be rarely used.
    1. Move the issue to "Ready to merge" the notify the original developer to merge the PR.
 
+Once the AT process is done, use the following templates to provide feedback in the ticket.
+
+{{< tabpane persistLang=false >}}
+{{< tab header="Test passed" >}}
+
+### Test details
+
+**Config:** <Default/standard>
+**Environment:** <Local>
+**Platform:** <WebApp>
+**Browser:** <Chrome>
+
+---
+
+### Test scenario:
+Description of the scenario - This is not required for all the tests
+
+### Reproducible on `master`
+A small description of how it was reproduced, and images or videos that support the comment.
+
+<details>
+<summary>Image/video attached</sumamry>
+</details>
+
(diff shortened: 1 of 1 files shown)


content/en/contribute/code/workflow.md
```diff
@@ -187,6 +187,80 @@ Issues in this column are in the process of being acceptance tested by a Quality
    1. How was the issue verified? Console logs, page source, errors anywhere, execution steps.  LGTM should begin to be rarely used.
    1. Move the issue to "Ready to merge" the notify the original developer to merge the PR.
 
+Once the AT process is done, use the following templates to provide feedback in the ticket.
+
+{{< tabpane persistLang=false >}}
```
andrablaj, line 192:
```suggestion
{{< tabpane persistLang=false lang="markdown">}}
```
The `lang="markdown"` option will render the code block with colors according to Markdown syntax.

content/en/contribute/code/workflow.md
```diff
@@ -187,6 +187,80 @@ Issues in this column are in the process of being acceptance tested by a Quality
    1. How was the issue verified? Console logs, page source, errors anywhere, execution steps.  LGTM should begin to be rarely used.
    1. Move the issue to "Ready to merge" the notify the original developer to merge the PR.
 
+Once the AT process is done, use the following templates to provide feedback in the ticket.
```
andrablaj, line 190: My suggestion is to add this as step number 4. in the list above. The existing step 4 would become step 5.

andrablaj (commented):
Love it; thanks for improving our AT process AND documenting it!

My suggestions are more on the style than the content itself.

content/en/contribute/code/workflow.md
tatilepizs: That was my plan at the beginning, I didn't add it as step 4 because there is one template for the failed scenario (that is the current step 3) and one for the passed scenario (that is the current step 4), so I was not sure if that might be confusing

andrablaj: OK, I understand. Let's keep it this way then, unless someone else has another suggestion.

andrablaj (approved):
LGTM, thanks!  💯

garethbowen (approved):
Looks great, thanks!

tatilepizs:
Merged to `main`

PR #998, opened 2023-03-23
Update app dev setup docs to include Dev Container instructions

jkuester:
Adds instructions to the App Developer Environment setup tutorial.  Things kinda felt all over the place in that tutorial, so I tried to streamline it a bit more.  

Highlights:

- The `local-setup` docs page seems really important since it will be the first place a dev will need to go to start building a CHT app.  To help highlight this fact, I updated the way the page shows up in the sidebar tree so that it is named `Getting started` and always shows up at the top of the `Tutorials` section.
- Our recommended Node/NPM version for app development is going to continue to change with the recommended versions for developing CHT Core.  So, I extracted those values to be global in the `config.toml`.
- The steps for setting up Docker locally are the same between cht-core development and app development, so I extracted those into a _partial that can be included in both pages (less repeated docs)


layouts/shortcodes/read-content.html
```diff
@@ -1,4 +1,4 @@
 {{ $path := .Get "file" }}
 {{ with $.Site.GetPage $path }}
-{{ .Content | markdownify }}
+{{ .Content }}
```
jkuester, line 3: Based on my tests, `markdownify` is not necessary for properly rendering file content using this shortcode.  However, having it in here was breaking the ability for pages with the tabpane to be rendered properly (since it was somehow confusing Hugo when it tried to calculate everything....).

mrjones-plip (approved):
thorough, well thought out and easy to review - thank you!

esthermmoturi (approved):
lgtm! Thanks @jkuester

PR #1005, opened 2023-03-31
feat: improve code health policy

garethbowen:


content/en/contribute/medic/product-development-process/code-health.md
```diff
@@ -6,15 +6,42 @@ description: >
...
+A good way to differentiate code health improvements from other changes is if it's something that a user won't notice, then it's code health. Because no user will ever ask for the change to be made they need to be managed outside of the usual prioritisation process.
 
+A good guideline is to spend 25% of your time improving code health, which can be split up into three distinct categories: Tidying, Cards, and Projects.
 
-### An example of how this works in practice
+### Tidying
+
+These are the smallest improvements and should be part of your work every week. Engineers can decide what tidying they want to prioritize and don't need any approval to do the work, but it will need to go through [Code Review]({{% ref "/contribute/code/workflow#code-reviews" %}}) to ensure a teammate agrees it's an improvement. Tidying can be done as part of an issue you're working on, or you can create a new issue and make the change as a separate PR, depending on which you prefer.
```
andrablaj, line 17: > Tidying can be done as part of an issue you're working on, or you can create a new issue and make the change as a separate PR, depending on which you prefer.

My only concern here is when the tidying work is outside of the scope of the current branch/issue, it might make the code review more difficult to review, as the reviewer will have to review each commit separately to understand which change is tidying and which change is the actual feature. 

Long story short, I recommend creating a new issue for tidying changes to reduce PR complexity and the risk of breaking things (e.g. via a dependency upgrade).

andrablaj (approved):
Thanks for pushing this forward @garethbowen.

I only have a comment, but otherwise all good!

craig-landry (approved):
Works for me. I agree with @andrablaj for cases when tidying is unrelated to what is being worked on, doing it separately is preferred. To be clear though, if tidying is to make the current task better, I think it should be done together. For example, if you're changing code and need a newer version of a library, I would expect a 1 PR that has a commit for the lib update and more commits for whatever code change is being made. I suppose an alternative could be 2 PRs/branches, where the first branch is the lib update and the 2nd branch is made off the first. My preference is still the 1 PR. With that I can see the lib update and why it's necessary together, in context. One thing that could help it be easier to review would be if a PR only had 2 commits in such a case: commit 1 is the lib update, commit 2 is the code change, and followup commits would be squashed into commit 2. Then the PR can be merged (not squashed) as the 2 commits.

Overall I'm still happy with whatever works for the team and has a result of tidying happening frequently.

content/en/contribute/medic/product-development-process/code-health.md
garethbowen: I've updated the text to cover this in more depth, thanks!

andrablaj (approved):
Looks good, thanks!

Issue #1008, opened 2023-04-05
Fix 'CHT API' & 'Partials'


abbyad:
The _partial files are coming up in ways they shouldn't:
- Two partial files are being shown as standalone pages in [reference](https://docs.communityhealthtoolkit.org/apps/reference/) instead of being pulled into their appropriate pages. 
- We are also cross referencing directly to the `_partial_*` files instead of the main pages. 

In this particular case, the missing context makes the appearance of these partials confusing. For instance, it's not clear the difference between [CHT API](https://docs.communityhealthtoolkit.org/apps/reference/_partial_cht_api/) and [API to interact with CHT Applications](https://docs.communityhealthtoolkit.org/apps/reference/api/).

Aside from fixing these, we should consider updating the docs style guide to provide guidance on the `_partial_` files.

---
Page: /apps/reference/_partial_cht_api/
Page: /apps/reference/_partial_utils/
Version: 2023-04-04 14:00:43  1200  1200
---

### Utils correctly pulled into other pages
<img width="1743" alt="image" src="https://user-images.githubusercontent.com/789512/230089272-f5937b36-08bb-4a27-b23e-2876cf2d781f.png">

### Utils showing up when it shouldn't
<img width="658" alt="image" src="https://user-images.githubusercontent.com/789512/230087637-0948b314-4290-419f-82d5-a8e6e3025e70.png">
<img width="553" alt="image" src="https://user-images.githubusercontent.com/789512/230087536-814f8c61-3fe3-40b2-a635-49ac61b92b89.png">
<img width="567" alt="image" src="https://user-images.githubusercontent.com/789512/230087964-991791f0-e471-47e8-a2ca-9993ff8b5850.png">
<img width="604" alt="image" src="https://user-images.githubusercontent.com/789512/230090115-c2704e1b-42db-42bf-ab41-37307635801f.png">
<img width="401" alt="image" src="https://user-images.githubusercontent.com/789512/230089773-2a2bb5ce-01f7-49c3-a9e2-e7ef5b5edbee.png">

abbyad:
Local docker setup too!

<img width="423" alt="image" src="https://user-images.githubusercontent.com/789512/230094704-160aa0b4-53e5-427b-848c-44b2b03c64d2.png">

abbyad:
[...]

PR #1009, opened 2023-04-05
1008 - Fix Partials

abbyad:
Removes partials from view in index summaries and related content links.

For #1008

### DIFF
--- a/content/en/apps/guides/hosting/4.x/_partial_docker_directories.md
+++ b/content/en/apps/guides/hosting/4.x/_partial_docker_directories.md
@@ -1,3 +1,8 @@
+---
+title: "Docker Directory Setup"
+toc_hide: true
+hide_summary: true
+---
 Create the following directory structure:
 
 ```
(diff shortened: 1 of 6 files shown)


jkuester (approved):
Very nice! Had not thought to check the summary pages.  The only other vector I wonder if we care about is when _searching_.  As it is, if I [search the docs](https://docs.communityhealthtoolkit.org/search/?q=api) for `api`, the second result is a link to `_partial_cht_api`.  Not ideal, but probably not a huge issue either....

abbyad:
Hmm, to deal with the search perhaps we try without the frontmatter altogether. Based on the [Hugo docs](https://gohugo.io/content-management/front-matter/) it seems as though it is optional, and it's not providing us with any functionality as it is now.

image

Will test a bit today and update this PR, or open a new one if we want longer to evaluate.

abbyad:
Removing the frontmatter works for some files, but breaks in others.

For instance, removing the frontmatter from this file results in errors

```
---
title: "Local Docker Setup"
toc_hide: true
hide_summary: true
---
{{< tabpane code=false >}}
{{% tab header="Linux (Ubuntu)" %}}
`shell
curl -fsSL get.docker.com -o get-docker.sh && sh get-docker.sh`
```

> ERROR 2023/04/06 14:31:42 Rebuild failed: "cht-docs\content\en\apps\tutorials\_partial_docker_setup.md:2:1": unmarshal failed: invalid character '{' looking for beginning of object key string

We can explore further to see how to begin files without frontmatter, but will leave that outside the scope of this PR.

PR #1013, opened 2023-04-07
fix(#975): Fix daily build

### LINKED ISSUE #975: Daily build - reporting failure to Slack fails
The "Test site and push live if we're on main branch " build appears to have been failing for months and we haven't noticed: https://github.com/medic/cht-docs/actions/workflows/daily.yml

The first failure is to do with the link checking:

```
Run ./.github/scripts/muffet.sh
  ./.github/scripts/muffet.sh
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    GOROOT: /opt/hostedtoolcache/go/1.18.10/x64
dial tcp4 127.0.0.1:1313: connect: connection refused
Error: Process completed with exit code 1.
```

The second failure is to do with reporting the failure to Slack:

```
Error sending message: Error on message: 403 Forbidden
```

Furthermore it should be renamed as it doesn't "push live" any more.

andrablaj:
Fix the daily build.

Closes #975

### DIFF
--- a/.github/scripts/muffet.sh
+++ b/.github/scripts/muffet.sh
@@ -17,7 +17,6 @@
 muffet http://localhost:1313 \
   --buffer-size 50000 \
   --timeout 10 \
-  --concurrency 2 \
   --ignore-fragments \
   --exclude ".*demo\.app\.medicmobile\.org.*" \
   --exclude ".*download\.docker\.com.*" \
@@ -26,7 +25,8 @@ muffet http://localhost:1313 \
   --exclude "https://github\.com/medic/cht-core/issues/new.*" \
   --exclude "https://github\.com/medic/cht-docs/commit.*" \
   --exclude "https://github\.com/medic/cht-core/commit.*" \
-  --exclude "https://github\.com/medic/cht-docs/edit/master/.*" \
+  --exclude "https://github\.com/medic/cht-docs/edit/main/.*" \
+  --exclude "https://communityhealthtoolkit.org/*" \
   --exclude "http[s]*://localhost[8443|5984]*" \
   --exclude "http[s]*://cht\.domain\.com.*" \
   --exclude "http[s]*://127\.0\.0*" \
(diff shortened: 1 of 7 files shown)


andrablaj:
@mrjones-plip @garethbowen  In order to get the build fixed, this is what I've done so far:

- Changed the Slack channel name from `cht-squad-docs` to `cht-doc-site`. Not sure this would fix the `Report errors to Slack, if any` step, as the error message  `Error sending message: Error on message: 403 Forbidden` seems more like an authorization problem.
- To mitigate the `dial tcp4 127.0.0.1:1313: connect: connection refused` error when running the `muffet` script, I increased the sleep time to 30s to ensure `hugo` is ready. 
- To test the script, I added `on: [push, pull_request]` at the beginning of the file to trigger the actions flow upon `push` (reverted the change).

After running the script, I created #1014 with a list of broken links. 

If we want to ignore `x509: certificate signed by unknown authority` [errors](https://github.com/medic/cht-docs/actions/runs/4639317886/jobs/8210031854?pr=1013), we need to add an extra flag to the `muffet` command: `--skip-tls-verification`.

mrjones-plip (approved):
lgtm!

minor question that shouldn't block going live, is why we exclude checks against `https://communityhealthtoolkit.org` ?  There's 5 links in docs to the site and they all work - seems maybe worth keeping?  No biggie!

mrjones-plip:
oh yeah - I forgot to answer your `--skip-tls-verification` question.  Hmmm - i don't feel too strongly, but if I had to choose I would think we should not add this skip - the site is effectively not accessible and we can always manually check a day or so later to see if it's been fixed.

andrablaj:
Thanks @mrjones-plip!

1.  I excluded the checks against `https://communityhealthtoolkit.org` because the requests failed with `400` when running `muffet.sh` locally. We can keep and remove the checks later if they fail in the build.

     Screenshot 2023-04-12 at 08 57 35

2.  

> `--skip-tls-verification` question. Hmmm - i don't feel too strongly, but if I had to choose I would think we should not add this skip - the site is effectively not accessible and we can always manually check a day or so later to see if it's been fixed.

OK, let's not add the flag for now and monitor the build for a few days.

PR #1020, opened 2023-04-14
Add docs for CHT Monitoring setup

mrjones-plip:
This PR:
* extends the existing [monitoring page](https://docs.communityhealthtoolkit.org/apps/guides/hosting/monitoring/) in the hosting section to cover the new [CHT Monitoring and Alerting](https://github.com/medic/cht-monitoring/) setup
* Refactors [existing readme](https://github.com/medic/cht-monitoring/https://github.com/medic/cht-monitoring/blob/9b7c2c43c21698416cb32f78c29f112196caaf38/README.md) into these sections:
   * Prerequisites
   * Setup
   * Upgrading
   * Additional Configuration
   * Configuration Reference
* maybe all these changes  make the page too long? open to input

per [16](https://github.com/medic/cht-monitoring/issues/16)


content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+
+When you see a new version in the [GitHub repository](https://github.com/medic/cht-monitoring), first review the release notes and upgrade instructions. Then, run the following commands to deploy the new configuration:
+
+1. Note the version number to derive the branch to pull.  For example,  `1.1.0` would be `1.1.x`
+2. Run this command to update your instance and restart it. Be sure to replace `BRANCH_NAME` with the value from the first step:
+   ```shell
+   cd ~/cht-monitoring
+   git fetch
+   git checkout BRANCH_NAME
+   ```shell
```
jkuester, line 171: Since new versions of the config could include additions to the docker-compose file, I think this should include:

```shell
docker compose pull
docker compose up -d --remove-orphans
```

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+
+#### Prometheus Retention and Storage
+
+By default, historical monitoring data will be stored in Prometheus (`PROMETHEUS_DATA` directory) for 60 days (configurable by `PROMETHEUS_RETENTION_TIME`). A longer retention time can be configured to allow for longer-term analysis of the data.  However, this will increase the size of the Prometheus data volume.  See the [Prometheus documentation](https://prometheus.io/docs/prometheus/latest/storage/) for more information.
+
+Local storage is not suitable for storing large amounts of monitoring data. If you intend to store multiple years worth of metrics, you should consider integrating Prometheus with a [Remote Storage](https://prometheus.io/docs/operating/integrations/#remote-endpoints-and-storage/).
+
+#### Alerts
+
+This configuration includes number of pre-provisioned alerts.  Specific alerting rules (and other contact points) can be set in the Grafana UI.
```
jkuester, line 183:
```suggestion
This configuration includes number of pre-provisioned alerts.  Additional alerting rules (and other contact points) can be set in the Grafana UI.  
```

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+
+Local storage is not suitable for storing large amounts of monitoring data. If you intend to store multiple years worth of metrics, you should consider integrating Prometheus with a [Remote Storage](https://prometheus.io/docs/operating/integrations/#remote-endpoints-and-storage/).
+
+#### Alerts
+
+This configuration includes number of pre-provisioned alerts.  Specific alerting rules (and other contact points) can be set in the Grafana UI.  
+
+See the both Grafana [high level alert Documentation](https://grafana.com/docs/grafana/latest/alerting/) and  [provisioning alerts in the UI](https://grafana.com/docs/grafana/latest/alerting/set-up/provision-alerting-resources/file-provisioning/#provision-alert-rules) for more information on how to edit or remove these provisioned alerts.
```
jkuester, line 185:
```suggestion
See both the Grafana [high level alert Documentation](https://grafana.com/docs/grafana/latest/alerting/) and [provisioning alerts in the UI](https://grafana.com/docs/grafana/latest/alerting/set-up/provision-alerting-resources/file-provisioning/#provision-alert-rules) for more information on how to edit or remove these provisioned alerts.
```

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+
+Additionally, you can configure where these alerts are sent.  Two likely options are Email and Slack.
+
+##### Email
+
+To support sending email alerts from Grafana, you must update the `smtp` section of your `grafana/grafana.ini` file with your SMTP server configuration.  Ten, in the web interface, add the desired recipient email addresses in the `grafana-default-email` contact point settings.
```
jkuester, line 191:
```suggestion
To support sending email alerts from Grafana, you must update the `smtp` section of your `grafana/grafana.ini` file with your SMTP server configuration.  Then, in the web interface, add the desired recipient email addresses in the `grafana-default-email` contact point settings.
```

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+
+##### Slack
+
+Slack alerts can be configured within the Grafana web GUI for the specific panel you would like to alert on.  Find the panel you want to alert on and edit it.  In the bottom half of the screen, choose the "Alert" tab. Then click "Create alert fule from this panel".
```
jkuester, line 195:
```suggestion
Slack alerts can be configured within the Grafana web GUI for the specific rules you would like to alert on.
```

>  Find the panel you want to alert on and edit it.  In the bottom half of the screen, choose the "Alert" tab. Then click "Create alert fule from this panel".

This feels like odd information to include in the Slack section since it is simply general info on how to create an alert from a panel. They should be able to trigger Slack notifications for existing alerts (and email notifs for alerts created from panels...).

jkuester (changes requested):
Added some thoughts/suggestions inline! 

For what its worth, I think going forward we are going to want to consider a whole subsection of docs pages for cht-monitoring. As we add additional data exporters, the doc size and complexity is going to continue to rise!

content/en/apps/guides/hosting/monitoring.md
mrjones-plip: Agreed!

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+### Prerequisites
+
+- [Docker](https://docs.docker.com/install/)
+- [Docker Compose](https://docs.docker.com/compose/install/)
+- [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git)
+- URL(s) of the CHT instance(s) 
+
+### Setup
+
+Pre-provisioned dashboards are included in the `CHT` folder in the Grafana UI should not be updated directly.
```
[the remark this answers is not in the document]
mrjones-plip: ah - agreed! fixed in 3b8865e

 I tried to find a home for all the original content, but we can remove for now.

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+1. Run the following commands to clone this repository, initialize your `.env` file, create a secure password and create your data directories:
+
+    ```sh
+    cd ~
+    git clone https://github.com/medic/cht-monitoring.git
+    cd cht-monitoring
+    cp cht-instances.example.yml cht-instances.yml
+    cp grafana/grafana.example.ini grafana/grafana.ini
+    mkdir -p grafana/data && mkdir  -p prometheus/data 
+    apt install -y wamerican
```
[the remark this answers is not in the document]
mrjones-plip: ah - good call!  I just learned you can install this file the other month when I did the single/multi node cht stuffs.  Hope my "always use a secure passphrase" preferences aren't too dogmatic!

fixed in c6334d2

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+    cp .env.example .env
+    password=$(shuf -n7 /usr/share/dict/words --random-source=/dev/random | tr '\n' '-' | tr -d "'" | cut -d'-' -f1,2,3,4,5,6,7)
+    sed -i -e "s/password/$password/g" .env
+    echo;echo "Setup complete! To log into Grafana in the browser:";echo
```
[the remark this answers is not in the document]
mrjones-plip: hah!  agreed. fixed in a513d60

content/en/apps/guides/hosting/monitoring.md
```diff
@@ -79,3 +79,161 @@ Messaging Outgoing State Failed | `non-zero` | 0
...
+    cp .env.example .env
+    password=$(shuf -n7 /usr/share/dict/words --random-source=/dev/random | tr '\n' '-' | tr -d "'" | cut -d'-' -f1,2,3,4,5,6,7)
+    sed -i -e "s/password/$password/g" .env
+    echo;echo "Setup complete! To log into Grafana in the browser:";echo 
+    echo "    username: medic"
+    echo "    password: ${password}";echo
+    ```
+   Note that in step 4 below you'll need the username and password which is printed after you run the above command.
+
+2. Edit the `cht-instances.yml` file to have the URLs of your CHT instances. You may include as many URLs of CHT instances as you like. To preserve data consistency parameters in this file should NOT be adjusted after it is initially set. Be sure to not include a trailing slash (`/`).
```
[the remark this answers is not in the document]
mrjones-plip: fixed in 5776efd

content/en/apps/guides/hosting/monitoring.md
mrjones-plip: ah - ok - I had removed that thinking we might not want to upgrade grafana/prometheus when we're just trying to update the cht-monitoring part.  I guess that should come in a separate commit to pin to a specific release vs `latest` like we have no.  That works!

I added a restart in 659ace5 and then I added `pull` just now in 1b34f0a.  Lemme know if we require the `compose up -d --remove-orphans` , assuming not though!

mrjones-plip:
@jkuester - I split up the docs into multiple pages like we discussed (see 2ee3455).  

Back to you for review!

content/en/apps/guides/hosting/monitoring.md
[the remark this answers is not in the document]
jkuester: :laughing: So, I did not actually mean to post that comment as-is!  I created it as I went through the review and meant to come back after reading the rest of it to see if there was a better place to put this. But, I forgot to come back to this. Oh well....

content/en/apps/guides/hosting/monitoring.md
jkuester: So, I might be wrong, but I don't think `restart` will actually update the image used (even if you have done a `pull` and gotten a new version) if you have existing containers (it just keeps the same containers).

The case I am trying to cover is if we ever update our base docker-compose config to add or remove a service (e.g. removing json-exporter since we don't need it any more).  In that case we also want the `--remove-orphans` since the json-exporter service will not be in the new version of the docker-compose file.

mrjones-plip: gotcha!  Added `--remove-orphans` to both calls and added a `docker compose down` before the `up` to not have the `restart` any more in e37e1be4.  Thanks for seeing this through!

mrjones-plip:
back to you for re-review - thanks @jkuester !

content/en/apps/guides/hosting/monitoring/setup.md
```diff
@@ -0,0 +1,168 @@
...
+    cp .env.example .env
+    password=$(shuf -n7 /usr/share/dict/words --random-source=/dev/random | tr '\n' '-' | tr -d "'" | cut -d'-' -f1,2,3,4,5,6,7)
+    sed -i -e "s/password/$password/g" .env
+    echo;echo "Initial project structure created! To log into Grafana in the browser:";echo 
+    echo "    username: medic"
+    echo "    password: ${password}";echo
+    ```
+   Note that in step 4 below you'll need the username and password which is printed after you run the above command.
+
+2. Edit the `cht-instances.yml` file to have the URLs of your CHT instances. You may include as many URLs of CHT instances as you like. To preserve data consistency parameters in this file should NOT be adjusted after it is initially set.
```
jkuester, line 53:
```suggestion
2. Edit the `cht-instances.yml` file to have the URLs of your CHT instances. You may include as many URLs of CHT instances as you like.
```

I guess we don't really need this caution any more since they can only set the root urls.

content/en/apps/guides/hosting/monitoring/setup.md
```diff
@@ -0,0 +1,168 @@
...
+    docker compose up -d
+    ```
+
+4. Grafana is available at [http://localhost:3000](http://localhost:3000). See the output from step 1 for your username and password.
+
+If you would like to do more customizing of your deployment, see ["Addition Configuration"](#additional-configuration).
+
+### Upgrading
+
+Before upgrading, you should back up both your current configuration settings and your Prometheus time-series data.
```
jkuester, line 76:
```suggestion
Before upgrading, you should back up both your current configuration settings as well as your Prometheus/Grafana data directories.
```

The Grafana data directory should also be backed up since it might contain custom dashboards/configuration/etc.

content/en/apps/guides/hosting/monitoring/setup.md
```diff
@@ -0,0 +1,168 @@
...
+
+#### Prometheus, Grafana and JSON Exporter
+
+To upgrade these dependencies, update the version numbers set in your `.env` file (or leave them set to `latest`).  Then run the following commands:
+
+```shell
+docker compose pull
+docker compose up -d --remove-orphans
```
jkuester, line 84:
```suggestion
docker compose up -d
```

Does not hurt to have the `--remove-orphans` here but might be confusing to a reader since there is no way to get orphan containers from running `docker compose pull`

content/en/apps/guides/hosting/monitoring/setup.md
```diff
@@ -0,0 +1,168 @@
...
+When you see a new version in the [GitHub repository](https://github.com/medic/cht-monitoring), first review the release notes and upgrade instructions. Then, run the following commands to deploy the new configuration:
+
+1. Note the version number to derive the branch to pull.  For example,  `1.1.0` would be `1.1.x`
+2. Run this command to update your instance and restart it. Be sure to replace `BRANCH_NAME` with the value from the first step:
+   ```shell
+   cd ~/cht-monitoring
+   git fetch
+   git checkout BRANCH_NAME
+   docker compose pull
+   docker compose down
```
jkuester, line 98:
```suggestion
```
I don't think we need `down` here.  My understanding is that, `up` is smart enough to re-create any container that has a new image (and just leaves the rest alone.)

jkuester (approved):
:rocket: Couple additional minor suggestions in line, but overall this is great!

content/en/apps/guides/hosting/monitoring/setup.md
mrjones-plip: So - I'm learning some good stuff about longitudinal docker compose health - thank you!

That said, I had `restart` in there originally because we may update something like the `prometheus.ini` or, as I often do the `cht-instances.yml`.  Will `docker compose up` stat those file to see if they've changed such that we can safely remove the `down`?  My (untested!!) assumption was that it wouldn't and the `down` would be needed.

jkuester: > we may update something like the prometheus.ini

Ohh, good point!  Yeah, probably safer to stick with this `down` in that case.  (AFIK, both Prometheus and Grafana do a good job hot-reloading their config files, but I like to be on the safe side)

content/en/apps/guides/hosting/monitoring/setup.md
```diff
@@ -0,0 +1,168 @@
...
+1. Run the following commands to clone this repository, initialize your `.env` file, create a secure password and create your data directories:
+
+    ```sh
+    cd ~
+    git clone https://github.com/medic/cht-monitoring.git
+    cd cht-monitoring
+    cp cht-instances.example.yml cht-instances.yml
+    cp grafana/grafana.example.ini grafana/grafana.ini
+    mkdir -p grafana/data && mkdir  -p prometheus/data 
+    apt install -y wamerican  # ensures /usr/share/dict/words is present for shuf call below
```
jkuester, line 43: Is it just me that still has to `sudo` apt commands?

jkuester (approved):

content/en/apps/guides/hosting/monitoring/setup.md
mrjones-plip: fair! 3933f29e

PR #1022, opened 2023-04-18
chore: More details to commit message

latin-panda:
# License 

Adds more details to commit message docs to match [PR template info. ](https://raw.githubusercontent.com/medic/cht-core/master/.github/PULL_REQUEST_TEMPLATE.md)

# License


garethbowen:
> Remove features that we don't need anymore, but it's not a breaking change like Guided Tour (feat? perf?)

If the feature isn't used anywhere then it's a chore IMO because the app is unchanged. If it was seriously annoying to users then call it a "feat" because we've improved the user experience. If you want to call it a "perf" then you'd better measure and publish the performance improvement!

I think I'd call removing the tour a "chore" because most people won't notice any difference - it's just cleaning up our code a little.

> Upgrade of dependencies

I'd go for chore here too, unless there's some specific improvement we're unlocking and documenting. Most of the time it's just cleaning stuff up.

> Adding telemetry to a feature 

I think this is a "feat" because it produces something that some users (app devs) may appreciate.

garethbowen (commented):
Up till now we've restricted ourselves to the set used by angular: https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-changelog-angular

This is completely arbitrary and we can add more, but if we do then we risk breaking automation because it doesn't know what "style" and "test" means. The way to achieve this without breaking automation would be with "feat" and "chore" respectively. We could configure the automation by providing our own configuration to the semantic-release package used by cht-conf but if we're deviating from convention we should have a good reason.

Why is it better to use "style" and "test" rather than "feat" and "chore"?

garethbowen:
Sounds good.

There may be value in separating things out one day (security, ui/ux, etc) and we can build that in to the automation so it's clear in release tests, but that would have to be combined with changes to the automation config for cht-conf.

latin-panda:
@garethbowen can you please have a look at the wording? Thanks

content/en/contribute/code/workflow.md
```diff
@@ -110,13 +110,13 @@ The main branch is `main` (or `master`) which must be kept stable so as not to i
...
-Features | feat(#456): add home tab | minor
-Non-code | chore(#123): update README | none
-Breaking| perf(#2): remove reporting rates feature <br/> BREAKING CHANGE: reporting rates no longer supported | major
+Type | Description | Example commit message | Release type
+-- | -- | -- | --
+Bug fixes | A bug fix. | fix(#123): infinite spinner when clicking contacts tab twice | patch
+Performance | A code change that improves performance. Measure the performance improvement to inform the community. | perf(#789): lazily loaded angular modules | patch
+Features | A new feature or change in the code that users will notice. | feat(#456): add home tab | minor
+Non-code | A change that user won't notice, like a change in a README file, adding e2e tests, update dependencies, removing unused code, etc. | chore(#123): update README | none
+Breaking | Breaking changes should be explained under performance type. | perf(#2): remove reporting rates feature <br/> BREAKING CHANGE: reporting rates no longer supported | major
```
garethbowen, line 119: "Breaking" is special because it could apply to any of the above types. "perf" was just this example, but it's highly likely to be a "feat" but could even be a "fix". Maybe we need to split it out of the table to make this clear?

garethbowen (changes requested):
Love the descriptions. One change requested.

content/en/contribute/code/workflow.md
latin-panda: I've created a note and added that piece there. What do you think?

content/en/contribute/code/workflow.md
```diff
@@ -108,17 +108,26 @@ The main branch is `main` (or `master`) which must be kept stable so as not to i
...
-Type | Example commit message | Release type
--- | -- | --
-Bug fixes | fix(#123): infinite spinner when clicking contacts tab twice | patch
-Performance | perf(#789): lazily loaded angular modules | patch
-Features | feat(#456): add home tab | minor
-Non-code | chore(#123): update README | none
-Breaking| perf(#2): remove reporting rates feature <br/> BREAKING CHANGE: reporting rates no longer supported | major
+Type | Description | Example commit message | Release type
+-- | -- | -- | --
+Fixes | A bug fix. | fix(#123): infinite spinner when clicking contacts tab twice | patch
```
garethbowen, line 115:
```suggestion
Bug fixes | Change code that wasn't working as intended. | fix(#123): infinite spinner when clicking contacts tab twice | patch
```

content/en/contribute/code/workflow.md
```diff
@@ -108,17 +108,26 @@ The main branch is `main` (or `master`) which must be kept stable so as not to i
...
+Performance | A code change that improves performance. Measure the performance improvement to inform the community. | perf(#789): lazily loaded angular modules | patch
+Features | A new feature or change in the code that users will notice. | feat(#456): add home tab | minor
```
garethbowen, line 117:
```suggestion
Features | A new feature or improvement that users will notice. | feat(#456): add home tab | minor
```

content/en/contribute/code/workflow.md
```diff
@@ -108,17 +108,26 @@ The main branch is `main` (or `master`) which must be kept stable so as not to i
...
+Non-code | A change that user won't notice, like a change in a README file, adding e2e tests, update dependencies, removing unused code, etc. | chore(#123): update README | none
```
garethbowen, line 118: Update for consistent tense.

```suggestion
Non-code | A change that user won't notice, like a change in a README file, adding e2e tests, updating dependencies, removing unused code, etc. | chore(#123): update README | none
```

garethbowen (changes requested):
Love the new "breaking changes" explanation - thanks!

Three minor suggestions inline.

garethbowen (approved):
Perfect. Thanks!

PR #1026, opened 2023-04-19
chore(#8185) Remove tour from docs

latin-panda:
# Description

medic/cht-core#8185

[CHT Core PR](https://github.com/medic/cht-core/pull/8188)

# License


michaelkohn (approved):
👍🏼

PR #1032, opened 2023-04-20
Update feature flags documentation 

michaelkohn:
The admin app now allows management of multiple user roles directly from the UI.  Here's the issue that added it https://github.com/medic/cht-core/issues/5903


latin-panda (approved):
LGTM!

PR #1035, opened 2023-04-21
test slack alerts w/ broken link

mrjones-plip:


abbyad (approved):
Approving for the sake of breaking the image link and testing the Slack push in the action.

PR #1049, opened 2023-05-01
Add container diagram for cht-monitoring

jkuester:
I have tweaked this diagram slightly from the original PR so that none of the labels are overlapping.  Unfortunately, I have not found any way to avoid the _lines_ from overlapping the other shapes.  Reading the issues, etc, it seems like this is not a big deal to the UML people. :shrug: 

An alternative would be to drop the C4 UML styling and just switch to a regular flowchart diagram.  It would not include as much data (e.g. descriptions of the containers), but is better supported in Mermaid (C4 diagrams are still beta) and has more style options.


mrjones-plip:
Can we simplify the docs UML one?  Requests are:
* remove "Reads" action, replace with directional arrows
* move frequency of update to protocol (eg `1m`), use `@` instead of `-` .  The `-` makes it look like a negative
* remove "<<containers>>" type labels
* make way less tall

I edited the SVG in inkscape (which is totes cheating), but gives you an idea of the goal:

docs setup

mrjones-plip:
First draft of that revamped image had text values instead of text -> path values, so the fonts were rendering wonky.  this has been updated in comment above.

mrjones-plip:
closing in favor of #1062

PR #1055, opened 2023-05-04
fix: Fix or exclude broken links

andrablaj:


.github/scripts/muffet.sh
```diff
@@ -33,5 +33,8 @@ muffet http://localhost:1313 \
   --exclude "http[s]*://127\.0\.0*" \
   --exclude "http[s]*://.*rapidpro.io.*" \
   --exclude "http[s]*://.*africastalking.com*" \
+  --exclude "http[s]*://.*udemy.com.*" \
+  --exclude "http[s]*://.*notion.so.*" \
+  --exclude "http[s]*://.*medium.com.*" \
```
dianabarsan, line 38: Apologies if this is supposed to be obvious. Are medium links behind a paywall and that's why they don't resolve? If muffet not able to bypass the browser check?

dianabarsan (approved):
I left one question inline, for my own knowledge, but otherwise looking good!

mrjones-plip (approved):
lgtm

PR #1057, opened 2023-05-04
feat: remove acceptance testing stages

garethbowen:


andrablaj (approved):
So happy to see this PR 🙌, thanks for moving this forward!

I have one comment, but it is not blocking merging this PR, so I approve it.

As we are removing the `Ready for AT section`, I was wondering if we could keep the **Testing template** as part of the development process (in progress step).  QA Engineers work together with developers during the development phase, but it might still be useful for them to have a template to share testing results early in the process.

Issue #1058, opened 2023-05-05
Update Google Analytics to GA4


abbyad:
By July 1, 2023 we need to migrate from our UA ID (`UA-168418637-1`) for Google Analytics to the new GA4 ID (`362515267`).

<img width="582" alt="image" src="https://user-images.githubusercontent.com/789512/236574036-a7058018-a95a-42a7-b773-0e4083d85539.png">

> On July 1, 2023, standard Universal Analytics properties will no longer process data. You'll be able to see your Universal Analytics reports for a period of time after July 1, 2023. However, new data will only flow into Google Analytics 4 properties.

abbyad:
Closing as I can confirm that the data is coming in properly now.

PR #1059, opened 2023-05-05
Switches Analytics from UA to GA4

abbyad:
For #1058

Uses the new Analytics ID and format, as per docs: https://www.docsy.dev/docs/adding-content/feedback/


garethbowen (approved):

garethbowen:
Do we need to do this for the forum and root site too?

abbyad:
Thanks, merging this and can move onto the other sites next.

PR #1067, opened 2023-05-11
fix: Exclude twitter link from check

andrablaj:
The [build](https://github.com/medic/cht-docs/commit/98aa1459e6879b7fc3929bf8d64f6a1c9ed73aec/checks) failed for 3 consecutive days, due to twitter sending `too many redirections` error.


latin-panda (approved):
Thanks!

PR #1068, opened 2023-05-11
Add docs about how to set log level

jkuester:


content/en/apps/guides/debugging/sharing-4x-logs.md
jkuester, line 1: I could not seem to create a local in-line link to this page with the extra period `.` in the file name...

mrjones-plip (changes requested):
The page you added this to  "Obtaining Browser and Phone Logs" (`apps/guides/debugging/obtaining-logs`) - is all about getting client side logs.  Even the "On the server" section is actually about feedback docs generated on the client which are pushed to the server.

i think we should move your new (excellent!) content to "Viewing server logs in CHT 4.x" (`apps/guides/hosting/4.x/logs/`)

jkuester:
Ah, good call @mrjones-plip!  That page makes way more sense!  :sweat_smile:  Thanks for the review!

mrjones-plip (approved):
🛃

PR #1073, opened 2023-05-15
Document how to enable languages from `app_settings.json`

m5r:
document how to use the feature implemented in medic/cht-core#6281


garethbowen (approved):
Nice one, thanks!

content/en/core/overview/translations.md
```diff
@@ -22,6 +22,7 @@ New languages must be added and configured in several places:
 - *In GitHub*
   - Create a new `messages-XX.properties` file in the [`api/resources/translations`](https://github.com/medic/cht-core/tree/master/api/resources/translations) folder, replacing XX with the 2 or 3 letter language code.
   - Add the language to the [`LOCAL_NAME_MAP` in api](https://github.com/medic/cht-core/blob/master/api/src/translations.js#L8). Use the language code for the key, and the local name followed by the English name for the language in brackets, eg: "fr: 'Français (French)'".
+  - Add an object to the `languages` array in `app_settings.json` with the `locale` and `enabled` properties representing respectively the 2 or 3 letter language code and whether that language should be enabled
```
jkuester, line 25: So, I think this docs page is focused on adding a new translation file _to the cht-core codebase_.  This is not what a partner should read when they are trying to setup the config _for their instance_.  

Maybe we should move this documentation to this page: https://docs.communityhealthtoolkit.org/apps/reference/translations/

The first paragraph on that page says:

> In the app_settings.json file the default language for the application is set, along with a separate default language for outgoing messages that are sent via SMS.

Right after that seems like a good place to add some more details...

content/en/apps/reference/app-settings/_index.md
```diff
@@ -28,19 +28,20 @@ The following settings do not need to be specified. They should only be defined
...
+| task_days_overdue                | Display number of overdue days in tasks list                                                                                                                                                                                                                                                                                                                                                                                            | false | 3.13.0  |
+| languages                        | Array of objects with `locale` and `enabled` properties representing respectively the 2 or 3 letter language code and whether that language should be enabled.  <br/>If unset it falls back to the previous behavior of relying on the `enabled` property of each translation document `messages-XX.properties`. This fallback behavior will be deprecated and this property will be required starting with the next major version 5.0. | | 4.2.0   |
```
jkuester, line 44:
```suggestion
| languages                        | Array of objects with `locale` and `enabled` properties representing respectively the 2 or 3 letter language code and whether that language should be enabled.  <br/>If unset it falls back to the previous behavior of relying on the `enabled` property of each translation document `messages-XX.properties`. This fallback behavior is now deprecated and will be removed in the next major version (5.0). This `languages` configuration property will be required for CHT 5.0+. | | 4.2.0   |
```

Minor, but I think the `enabled` in the translation documents is considered "deprecated" _now_ (since it has been superseded by a new go-forward config).  Then, in `5.0` it will be "removed".

jkuester (changes requested):
Excellent! Just a wording nit-pick and some thoughts about placement.,,,

content/en/core/overview/translations.md
m5r: Good call, I agree this is a better place to document this

content/en/apps/reference/translations.md
```diff
@@ -7,7 +7,10 @@ description: >
 keywords: localization languages translations
 ---
 
-Given that CHT apps are used around the world, the Core Framework was designed with localization in mind. The Core Framework itself is available in English, French, Hindi, Nepali, Spanish, Swahili, and Indonesian. In the `app_settings.json` file the default language for the application is set, along with a separate default language for outgoing messages that are sent via SMS. 
+Given that CHT apps are used around the world, the Core Framework was designed with localization in mind. The Core Framework itself is available in English, French, Hindi, Nepali, Spanish, Swahili, and Indonesian.  
+In the `app_settings.json` file the default language for the application is set by the `locale` property, along with a separate default language for outgoing messages that are sent via SMS with the `locale_outgoing` property.  
+Additionally, languages can be enabled and disabled through the `languages` property which contains an array of objects. These objects should contain the `locale` and `enabled` properties representing respectively the 2 or 3 letter language code and whether that language should be enabled.
```
jkuester, line 12:
```suggestion
Additionally, languages available to the user can be enabled and disabled through the `languages` property which contains an array of objects. These objects should contain the `locale` and `enabled` properties representing respectively the 2 or 3 letter language code and whether that language should be enabled.
```

jkuester (approved):
LGTM!

PR #1081, opened 2023-05-16
fix(na) remove extra mermaid conf fix numbering in watchdog

mrjones-plip:
This PR:
* removes the extra mermaid config
* fixes the numbering on the CHT Watchdog page

The [extra mermaid config](https://github.com/medic/cht-docs/blob/58e3ced0a9230f71adf3f86430fe7b3ae31dc1e4/config.toml#L213) removes the colors and tightens up the charts.  Top is with this PR, bottom is [what is live now](https://docs.communityhealthtoolkit.org/apps/guides/hosting/monitoring/integration/#additional-flows):

image

[The numbering that is off](https://docs.communityhealthtoolkit.org/apps/guides/hosting/monitoring/integration/#steps-to-new-integrations) that this PR fixes:

image

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -209,11 +209,3 @@ RelPermalink= "communityhealthtoolkit.org"
  [[related.indices]]
     name = "keywords"
     weight = 1
-
-[params.mermaid]
-theme = "neutral"
-
-[params.mermaid.flowchart]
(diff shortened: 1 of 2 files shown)


jkuester (approved):
LGTM!

PR #1084, opened 2023-05-16
add boxes to people hierarchcy

mrjones-plip:
This PR adds boxes around the people hierarchy subgraphs to make it more clear what lines are associated with which nodes.  It also lines up the families and family members horizontally. 

Left is what is live, right is what this PR does:

image

### DIFF
--- a/content/en/apps/concepts/hierarchy.md
+++ b/content/en/apps/concepts/hierarchy.md
@@ -56,11 +56,12 @@ The hierarchy can be modeled after the health system, health program and/or the 
 For example, a Family Member is part of a Family. A Family and CHWs are part of a CHW Area. A Family Member, a Family, and CHWs are part of a CHW Supervisor Area. 
 
 ```mermaid
-%%{init: { "flowchart": { "rankSpacing": 15, "nodeSpacing": 15 } } }%%
+%%{init: { "flowchart": { "rankSpacing": 20, "nodeSpacing": 10 } } }%%
 flowchart TB
 linkStyle default stroke-width:1px,stroke:lightgrey
 
-classDef none fill:none,stroke:none
+classDef node fill:none,stroke:none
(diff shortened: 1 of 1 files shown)


esthermmoturi:
This is great @mrjones-plip ,is it possible to add a line that explains the relationship between the CHW area and CHW especially in regards to creating/adding users?

mrjones-plip:
Thanks @esthermmoturi ! 

I don't understand what your request is.  Could give more details? I notice we already have a line between the CHW Area and the CHW. 

If it helps, here's the [page the](https://docs.communityhealthtoolkit.org/apps/concepts/hierarchy/#people) chart is currently on. Maybe this is an improvement enough that we can ship it and then make another PR for the change you're requesting?

esthermmoturi:
@mrjones-plip I meant adding a statement to show the relationship. Sure go ahead and ship it, we can always add the statement in another PR.

mrjones-plip:
Thanks @esthermmoturi !  

I'll need an approval from you or @jkuester  before I can merge.

jkuester (approved):
I think the boxes do a good job of making the connections less ambiguous!  (From an aesthetic perspective, I wish there was a way to put a little rounding on the box corners, but I could not quite figure out how to do that here... Not a big deal though!)

My only concern is how you have change the layout for the "unbalanced" part of the graph (the families with no CHW Areas in their hierarchy.  Personally, it makes more sense to me to have it like you did where the families are displayed at the same level as the other families, but the original graph did not have them like this:

graph

I am fine with moving them down, but it is not super clear to me why the original graph had them moved up...

mrjones-plip:
Thanks for your review @jkuester !  

Imma push it :sunglasses: 

@abbyad - lemme know if you have any concerns!
@esthermmoturi - you as well! 

Open to input from you both.

abbyad:
Looks better than it was, nice work!

PR #1088, opened 2023-05-19
fix: Replace placehold.it URLs with PNGs

andrablaj:
Screenshot 2023-05-19 at 09 57 34


andrablaj:
Per @latin-panda great suggestion, I replaced the `placeholder.it` links with the `PNGs` that I downloaded from the links

latin-panda (approved):
✨✨Obrigado✨✨

PR #1091, opened 2023-05-23
Include tip for restarting Watchdog when making config changes

jkuester:
Closes https://github.com/medic/cht-watchdog/issues/49


mrjones-plip (approved):
🕺

PR #1104, opened 2023-06-01
fix 404  on DBT link

mrjones-plip:


mrjones-plip:
Thanks for the fast turn around @njogz ! Accidentally clicked "close" instead of "merge" :facepalm: 

i'll merge once the rebuild passes CI again!

PR #1105, opened 2023-06-02
Certbot only works on CHT 4.2, instruct to use own domain

mrjones-plip:


garethbowen (commented):
@mrjones-plip Looks good to me! Can you please fix the conflict and I'll get this merged.

garethbowen (approved):

Issue #1106, opened 2023-06-05
local-ip.medicmobile.org doesn't resolve on GH runners, causes link check errors


mrjones-plip:
When the [link checker script](https://github.com/medic/cht-docs/blob/main/.github/scripts/muffet.sh) runs, it fails to resolve [local-ip URLs](https://local-ip.medicmobile.org), which gives an error.

```
2m 51s
Run ./.github/scripts/muffet.sh
  ./.github/scripts/muffet.sh
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    GOROOT: /opt/hostedtoolcache/go/1.18.10/x64
http://localhost:1313/apps/tutorials/local-setup/
	lookup local-ip.medicmobile.org: Temporary failure in name resolution	https://local-ip.medicmobile.org/
http://localhost:1313/apps/guides/hosting/3.x/offline/
	lookup local-ip.medicmobile.org: Temporary failure in name resolution	https://local-ip.medicmobile.org/
Error: Process completed with exit code 1.
```

mrjones-plip:
so far unable to reproduce:

*    found the [docker image GH uses](https://github.com/catthehacker/docker_images/pkgs/container/ubuntu) (or close enough hopefully): `docker run -d --name muffet ghcr.io/catthehacker/ubuntu:act-22.04 tail -f /dev/null`
*    spun up a local container, installed go 1.18 and muffet 1.5.7 (`docker exec -it muffet bash` and then some curl love with `https://golang.org/dl/go1.16.7.linux-amd64.tar.gz`  and then muffet with `GO111MODULE=on go install github.com/raviqqe/muffet@v1.5.7`)
*    ran [the script](https://github.com/medic/cht-docs/blob/main/.github/scripts/muffet.sh) against a local webserver with a local-ip link

I suspect this is endemic to GH runner DNS - so must reproduce in situ....my last favorite situ-ation to be in :(

mrjones-plip:
Given the recent issues with DNS on local-ip (see [here](https://github.com/medic/cht-core/issues/8289) and [here](https://github.com/medic/cht-core/issues/8100)), I'm doing a bit of a deep dive to see if this is endemic to GH CI or [localtls](https://github.com/Corollarium/localtls) instance on https://local-ip.medicmobile.org

mrjones-plip:
[...]

PR #1109, opened 2023-06-08
fix(#1106) Muffet DNS Issues & CI Names

mrjones-plip:
This PR:
* Adds a new `[set_dns.sh](https://github.com/medic/cht-docs/compare/1106-fix-muffet-dns-issues?expand=1#diff-41bc4f19b66c95e56aa6a4eb3e44559e4f15369d246de7e9c3125a6c681ebfa5)`  script which sets GH CI runner DNS to `9.9.9.9`
* Updates CI to call the new `set_dns.sh` script
* Updates runner names to no longer have "All Branches -" and "Main only" as this script is always called on `main`, never on branches.

### DIFF
--- a/.github/scripts/set_dns.sh
+++ b/.github/scripts/set_dns.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# At one point we had issues with the DNS server that GH CI Runners use,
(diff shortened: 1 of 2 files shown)


andrablaj (approved):
Thanks for solving this issue! 💪

Issue #1108, opened 2023-06-08
Improve 'API to interact with CHT Applications': Empty place holders


yrimal:
Page: /apps/reference/api/
Version: 2023-05-12 08:26:27  0700  0700
---

<img width="623" alt="image" src="https://github.com/medic/cht-docs/assets/6102813/c3a64248-761a-4d1c-83c0-dff23af2ae08">

There are empty placeholders, refer to above image.

garethbowen:
Fixed, and while I was there, removed the deep nesting so the toc is a little easier to use.

PR #1110, opened 2023-06-09
fix(#1108): fix empty placeholders

garethbowen:
The empty placeholders were because of using heading sizes that weren't incrementally smaller then the parent (eg: 3 hashes with a subhead of 5 hashes). I manually fixed these, then found there's a config change we can make to simplify the TOC to make it more usable too. As far as I can see this is the only place where we use a TOC so it makes no difference anywhere else.

@mrjones-plip I'm happy to revert the heading size change _or_ the toc change _or_ neither - let me know which you prefer!

#1108

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -81,7 +81,7 @@ time_format_blog = "02.01.2006"
 
 [markup]
   [markup.tableOfContents]
-    endLevel = 3
+    endLevel = 4
   [markup.goldmark]
     [markup.goldmark.renderer]
       unsafe = true
--- a/config.toml
+++ b/config.toml
@@ -81,7 +81,7 @@ time_format_blog = "02.01.2006"
 
 [markup]
   [markup.tableOfContents]
-    endLevel = 4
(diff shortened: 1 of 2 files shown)


mrjones-plip (approved):
@garethbowen  - thanks for diving in to tidy up these little annoyances !

The PR is such a minor change, approving to unblock, but am requesting a change: I made [the  `endLevel` change](https://github.com/medic/cht-docs/commit/2f3b3ab3bbc97f39ad97ab06ed38bc1a0c7315e2) and find it's addition very beneficial! I think should do one of the other options you suggested, but keep this at `4`. Sample screenshots below.

Thanks!

image

image

garethbowen:
Good spotting! Thanks I missed that use case altogether.

I've reverted that, but added an inline style to reduce the toc depth just on this one page because I think it makes the API toc much easier to use. Would you mind having another look?

mrjones-plip (approved):
:ship: 

image

Issue #1122, opened 2023-06-20
Improve 'CHT Watchdog Setup'


yrimal:
Page: /apps/guides/hosting/monitoring/setup/
Version: 2023-05-23 14:58:17 -0500 -0500
---

On the documentation, mention that absolute path needs to be used for docker compose version 2.x and later.

mrjones-plip:
looks like #1123 fixed this - closing this.

Thanks @yrimal !

PR #1123, opened 2023-06-20
#1122 Update documentation about absolute path

yrimal:
Update cht-watchdog setup guide about absolute path

For #1122


jkuester (approved):
Good call on updating this doc!  I was originally hoping that not too many folks would be on `docker-compose 2.x`, but now I think that is unrealistic (it is going to take awhile for everyone to get uplifted to `docker compose`.)  This will help avoid confusion!

PR #1128, opened 2023-06-26
 update TLS docs to include AUTO_GENERATE and other important steps/warnings

### LINKED ISSUE #1127: Improve 'Adding TLS certificates in CHT 4.x'
Page: /apps/guides/hosting/4.x/adding-tls-certificates/
Version: 2023-04-25 06:31:07 -0700 -0700
---

We were following the [certbot certificate section](https://docs.communityhealthtoolkit.org/apps/guides/hosting/4.x/adding-tls-certificates/#certbot-certificate) to add TLS certificate, but we got this error:
```
certbot    | certbot.errors.PluginError: /etc/nginx/private/certbot/ does not exist or is not a directory
```
In our case, the environment variable `CERTIFICATE_MODE` was set to `"SELF_SIGNED"`.
It looks like it needs to be set to `"AUTO_GENERATE"` for `cht-nginx` to create the correct directories and scripts.

We should add this information to the docs.

mrjones-plip:
This PR updates the 4.x TLS page to:
* have info about the `AUTO_GENERATE` value being set in CHT
* warns that the certbot stuff only works with >4.2
* talks about using `STAGING` certbot servers to do lots of TLS testing without hitting rate limit from certbot: `too many certificates (5) already issued for this exact set of domains in the last 168 hours`

closes #1127

### DIFF
--- a/content/en/apps/guides/hosting/4.x/adding-tls-certificates.md
+++ b/content/en/apps/guides/hosting/4.x/adding-tls-certificates.md
@@ -54,11 +54,25 @@ Your certificates are now safely stored in the native docker volume. Restart you
 
 ## Certbot certificate
 
-If you have a deployment with a publically accessible domain name, you can have Certbot automatically create free TLS certificates by using [their Docker image](https://hub.docker.com/r/certbot/certbot/). 
+_The Certbot integration requires CHT 4.2.0 or later_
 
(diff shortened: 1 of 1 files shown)


binokaryg (approved):
Great! Thanks!

Issue #1145, opened 2023-08-03
Add documentation for /api/v1/express-metrics


kennsippell:
@kennsippell Please also update the API documentation here: https://docs.communityhealthtoolkit.org/apps/reference/api/

_Originally posted by @garethbowen in https://github.com/medic/cht-core/issues/8426#issuecomment-1662737501_

PR #1146, opened 2023-08-03
#1145 - Documentation for /api/v1/express-metrics

kennsippell:
#1145


garethbowen (approved):
I was going to suggest using the standard test for "added in" but realised we're not very consistent, so went ahead and made the other changes too. For example, you need two new lines to actually create a new paragraph in the resulting html.

Otherwise everything looks good. Would you mind reviewing _my_ changes?

kennsippell:
Yep! Looks great! Thanks!

PR #1150, opened 2023-08-11
chore: default filter clarification

latin-panda:


latin-panda:
Hi @esthermmoturi, I want to add a note on performance implications when enabling default filters in the Reports tab. 
Online users with many reports, such as 10k, 20k, 50k etc, can experience bad performance whenever 1 or more filters are applied that match thousands of reports.

content/en/apps/reference/app-settings/user-permissions.md
```diff
@@ -67,7 +67,7 @@ Permissions can be assigned to user roles either directly in `base_settings.json
...
-| `can_default_facility_filter` | Defaults the Place Filter in Reports tab to the user's associated facility. The user should have a contact associated that belongs to a facility. This feature is not available for Admin and Offline type of users. Added in 4.3.                                                                                                                                                        |
+| `can_default_facility_filter` | Defaults the Place Filter in Reports tab to the user's associated facility. The user should have a contact associated that belongs to a facility. This feature is not available for Admin and Offline type of users. Use with caution, online users that can access thousands of reports can experience slow performance specially where network is slow. Added in 4.3.  |
```
esthermmoturi, line 70:
```suggestion
| `can_default_facility_filter` | Defaults the Place Filter in Reports tab to the user's associated facility. The user should have a contact associated that belongs to a facility. This feature is not available for Admin and Offline type of users. Use with caution, online users that can access thousands of reports can experience slow performance especially where the network is slow. Added in 4.3.  |
```

esthermmoturi (changes requested):

content/en/apps/reference/app-settings/user-permissions.md
latin-panda: Fixed! Thank you!

esthermmoturi (approved):

PR #1152, opened 2023-08-14
Update doc added new type phone_number

PrjShrestha:
Update doc added new type phone_number in app settings section.


dianabarsan (changes requested):
Cool! Thanks for the update! 
I think we also need to specify that `registration` can also have a config for which field to accept as phone number and that the value is saved on the newly added patient.

garethbowen:
@PrjShrestha Can you update this PR addressing @dianabarsan 's feedback?

PR #1155, opened 2023-08-16
feat: add SonarCloud documentation

jkuester:
Adds general documentation regarding the use of SonarCloud, how it can be configured, and how to set it up on new repos.

These docs just capture my initial thoughts on how we should use Sonar. This can provide a framework for further discussion to determine how we actually want to proceed!  See https://github.com/medic/cht-core/pull/8466 for an initial repo-level configuration.

## Org Level Settings

There are several org-level settings that should be discussed (and this PR seems as good as anywhere).

### Analysis Method

Since we are planning to use SonarCloud to analyze public repos, we can just use their default "Automatic" analysis method.  This will automatically analyze the default branch for the repo as well as any new PRs.

Alternatively, we could use the [GitHub Action](https://github.com/SonarSource/sonarcloud-github-action). It would be less convenient, but more configurable...

### New Code Definition

As noted in the docs, we need to specify a NCD for the repos we are going to analyze.  As far as I can tell, this only really matters for the repo-level reporting. The NCD does not apply to PRs (where the code changes in the PR are what is considered "new").  I have not been able to think of any reason not to just keep the default value of code from the last `30 days`, but perhaps someone else has more thoughts here....

### Quality Gate

The quality gate defines the conditions that must be met for the code to "pass" the Sonar analysis.  So, if the code in your PR fails the quality gate, your PR will be blocked from merging.  

The default `Sonar Way` quality gate has the following conditions:
	- Coverage < 80.0%
	- Duplicated Lines (%) > 3.0%
	- Maintainability Rating  < A
	- Reliability Rating < A
	- Security Hotspots Reviewed < 100%
	- Security Rating < A

However, all of this can be fully customized with these available conditions:

<details>
<summary>All Conditions</summary>

- Coverage:
	- Condition Coverage
	- Conditions to Cover
	- Coverage
	- Line Coverage
	- Lines to Cover
	- Uncovered Conditions
	- Uncovered Lines
- Duplications:
	- Duplicated Blocks
	- Duplicated Lines (%)
	- Duplicated Lines
- Issues
	- Blocker issues
	- Critical Issues
	- Info Issues
	- Issues
	- Major Issues
	- Minor Issues
- Maintainability:
	- Added Technical Debt
	- Maintainability Rating
	- Code Smells
	- Technical Debt Ratio
- Reliability:
	- Bugs
	- Reliability Rating
	- Reliability Remediation Effort
- Security:
	- Vulnerabilities
	- Security Rating
	- Security Remediation Effort
- Security Review:
	- Security Hotspots Reviewed
	- Security Review Rating
- Size:
	- Lines

</details>

### Quality Profiles

The quality profiles define the lists of rules that will be applied when doing analysis.  As noted in my documentation, Sonar maintains sensible default profiles, but they are broadly configurable.

The two main rules I have identified that we may want to configure right from the beginning are:

- [`S3776`](https://rules.sonarsource.com/javascript/RSPEC-3776/) Cognitive Complexity of functions should not be too high
	- Can configure max complexity level - default = `15`
- [`S107`](https://rules.sonarsource.com/javascript/RSPEC-107/) Functions should not have too many parameters
	- Can configure max parameters - default = `7`

We can leave these as their default values, but now is a good of time as any to consider if we want to set these to something else!

### DIFF
--- a/content/en/contribute/code/static-analysis.md
+++ b/content/en/contribute/code/static-analysis.md
@@ -1,0 +1,126 @@
+---
+title: "Static Analysis"
+linkTitle: "Static Analysis"
+weight: 9
+description: >
+  Guidelines for CHT SonarCloud static analysis.
+---
+
+Sonar static analysis supports development by providing feedback on code quality and security issues. Sonar analysis must pass on all _new code_.
+
+SonarCloud can be enabled on any public repo in the `medic` organization.
+
+## Workflow
+
+### During development
+
+While writing code, the [SonarLint](https://www.sonarsource.com/products/sonarlint/) plugin can be used to get real-time code analysis in your IDE. This is useful to avoid committing code with Sonar issues in the first place. 
+
+### PR Analysis
+
+SonarCloud is integrated with the CHT GitHub repositories and runs on every pull request. The results are posted as a comment on the pull request. If the Sonar analysis fails the quality check, the pull request will be blocked from merging.
+
+#### What should I do if Sonar finds an issue?
+
+When Sonar flags an issue with the code in your Pull Request, use this decision tree to determine the proper mitigation:
+
+1. If the issue is a genuine concern that should be addressed:
+   1. Fix it and push the updated code to your PR. The PR will automatically be unblocked once the Sonar analysis succeeds.  
+1. If the issue is a "false-positive" (i.e. Sonar has flagged some particular code as violating a rule, but it does not make sense to apply the rule in that context):
+   1. If the rule is one that should not be applied to any CHT code:
+      1. [Remove the rule]({{< ref "#removing-a-rule" >}}) from the default Quality Profile.
+   1. If it does not make sense to apply the rule to this particular code, you can do one of the following:
+      1. Completely ignore Sonar issues [on that line of code](https://docs.sonarsource.com/sonarqube/latest/user-guide/issues/#technical-review) by adding the `// NOSONAR` comment to the end of the line.
+      1. Update the `.sonarcloud.properties` to [ignore _that rule_ for that particular file]({{< ref "#ignoring-a-specific-rule-for-a-file" >}}).
+      1. Update the `.sonarcloud.properties` to [ignore _all rules_ for that particular file]({{< ref "#ignoring-all-rules-for-a-file" >}}) (useful if the file has been copied from an external dependency).
+
+## Adding a new repo to SonarCloud
+
+In the GitHub UI, navigate to the settings for the `medic` org > Integrations > Applications > SonarCloud and click the `Configure` button.
+
+In the Repository access section, select your desired repository from the drop-down and click `Save`.
+
+You will be automatically redirected to the SonarCloud UI where you can configure the repo-level settings. Use the `+` button and choose `Analyze new project`.  Select the repo from the list and click `Set Up`.
+
+### New Code Definition
+
+When setting up a new repository in SonarCloud, you will be asked to define what is considered to be "new code". This is used to determine which code in the default branch is considered "new" (affects reporting of issues, etc).
+
+Consult [the documentation](https://docs.sonarcloud.io/improving/new-code-definition/) for more details on the options available. For projects that do not use Gradle or Maven for version management, the `Number of days` option is recommended (since `Previous version` would require maintaining a version number in the `.sonarcloud.properties` file).
+
+## Sonar Configuration
+
+Broadly speaking, Sonar configuration is separated into repo-level and org-level configuration.
+
+### Repo-level configuration
+
+Each repository can include a `.sonarcloud.properties` file in the root directory. 
+
+This file must specify the path to the source code in the repository as well as which source files should be considered to be test code. See [the documentation](https://docs.sonarsource.com/sonarqube/latest/project-administration/analysis-scope/) for more details.
+
+```properties
+# Path to sources
+sonar.sources=.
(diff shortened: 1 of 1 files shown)


content/en/contribute/code/static-analysis.md
```diff
@@ -0,0 +1,128 @@
...
+
+In the Repository access section, select your desired repository from the drop-down and click `Save`.
+
+You will be automatically redirected to the SonarCloud UI where you can configure the repo-level settings. Use the `+` button and choose `Analyze new project`.  Select the repo from the list and click `Set Up`.
+
+### New Code Definition
+
+When setting up a new repository in SonarCloud, you will be asked to define what is considered to be "new code". This is used to determine which code in the default branch is considered "new" (affects reporting of issues, etc).
+
+Consult [the documentation](https://docs.sonarcloud.io/improving/new-code-definition/) for more details on the options available. For projects that do not use Gradle or Maven for version management, the `Number of days` option is recommended (since `Previous version` would require maintaining a version number in the `.sonarcloud.properties` file).
```
garethbowen, line 51: How will the NCD be used? If I'm submitting a PR will it block if someone else changed that same file within the NCD period? Or is it only when analysing the default branch? If so why not also analyse the old code?

I guess what I'm asking is, what is the workflow which is improved by the distinction of new vs old code?

garethbowen:
Thanks @jkuester ! The metrics in the quality gate are really complicated but after reading the docs I think they make sense.

Other than my confusion about NCD this looks good to me, but I'll only know for sure once we've been using it for a while.

andrablaj (commented):
I added a few minor non-blocking comments. This is a great start about how we can setup Sonar and manage the rules.

content/en/contribute/code/static-analysis.md
jkuester: As far as I can tell, if we use the "Automatic" analysis method and require that all PRs pass the Quality Gate, then the NCD is not really very important to us.  In general, I think the point of an NCD was as a generic way to incrementally measure code quality (that did not rely on specific implementation details of an SCM like git/GitHub).  If your code creation workflow does not involve branches/PRs/etc, then you need some way of deciding what should be analyzed as "changed" when trying to determine if the code base is in an acceptable state.  

Like I said, though, I don't think this really matters much to us since, with a PR, the NCD is not applied and just code code changed in the PR is considered "new".  The NCD will get used when Sonar analyzes the `master` branch, but since all our PRs should already be passing the Quality Gate, then it should not be possible for there to be any issues in "new" code in the `master` branch.

jkuester:
I imagine @craig-landry will have some wise thoughts particularly on our Quality Profile configuration I mentioned in the PR description (should we customize the parameter values of those two rules?)!

craig-landry:
I like the idea of starting off with a standard base set of things. I agree that the cognitive complexity and max params noted in the description are good cases to go further from the start. My preference is cognitive complexity set to 5 and max params set to 4. 

_[pause for riot]_

For both cognitive complexity and max params I favor radical reductions in the enforced values to bring about impactful change. We could take the value of 15 and make it 10, but that really just says "keep doing what you're doing and kinda try to cut some stuff out". That probably just annoys developers and doesn't give much benefit for the trouble. The goal here is to change how we're thinking about organizing logic in our code though.

Things that need a lot of parameters tend to be hiding something bigger. It may be as simple as a [data clump](https://refactoring.guru/smells/data-clumps) code smell to clean up or take more consideration like rethinking logic to break up logic into more general components. 

Code that has a lot of manual looping is probably missing a simple abstraction, and code with a lot of conditional logic is likewise missing a representation of a concept, which in current code is just written out as line-by-line logic.

image

content/en/contribute/code/static-analysis.md
```diff
@@ -0,0 +1,128 @@
+---
+title: "Static Analysis"
+linkTitle: "Static Analysis"
+weight: 9
+description: >
+  Guidelines for CHT SonarCloud static analysis.
+---
+
```
garethbowen, line 8: Given the page is called "static analysis" we should also include eslint here.

```suggestion

# eslint

All code must pass an eslint check which runs early in the CI cycle and uses the [standard medic eslint configuration](https://github.com/medic/eslint-config).

# sonar

```

content/en/contribute/code/static-analysis.md
```diff
@@ -0,0 +1,128 @@
+---
+title: "Static Analysis"
+linkTitle: "Static Analysis"
+weight: 9
+description: >
+  Guidelines for CHT SonarCloud static analysis.
```
garethbowen, line 6:
```suggestion
  Guidelines for static analysis of CHT code.
```

garethbowen (approved):
Suggestions inline for documenting eslint too.

andrablaj (approved):
- Analysis Method
  - I agree with using their `Automatic` analysis method to start with, and we can always integrate the Github Action if we find a scenario where that would serve us better. 
- New Code Definition & Quality Gate
  - Same, we can give the default values a try and adjust them when we would get more experience with the tool. 
 - Quality Profiles
   - I second @craig-landry on having radically lower numbers as a target, to intentionally push towards better code quality and code design. I am positive this would trigger fantastic discussions in future PRs. 
 
Let's do this 🦾! Thanks @jkuester for starting the discussion here!

garethbowen (approved):
Awesome, thanks!

andrablaj (approved):
👌

Issue #1161, opened 2023-08-28
fix 2 broken links in docs


mrjones-plip:
per CI:

>  http://localhost:1313/apps/guides/integrations/rapidpro/
>	timeout	https://youtu.be/WcFhpSFhFug
> http://localhost:1313/apps/concepts/interoperability/
>	timeout	https://wiki.ohie.org/display/CP/Use+Case+Summary%3A+Request+Community+Based+Follow-Up
> Error: Process completed with exit code 1.

andrablaj:
Closing this issue as those links seem to work fine now.

PR #1163, opened 2023-08-29
fix(#1161): increase muffet timout from 10 to 35

mrjones-plip:
back in a8ea000cae45cc0b68b73a082b5beaad1de739a6 the `timeout` was decreased from `35` to `10` let's revert that in hopes of avoiding timeouts per #1161


andrablaj (approved):
👌

PR #1173, opened 2023-09-06
feat(scaling): horizontal vs vertical scaling

mrjones-plip:
this PR:
* Adds documentation about how the CHT Scales horizontally
* Is intentionally not specific to docker compose vs k8s hosting
* Is meant to give self hosting and national scale deployments confidence about when and why to scale horizontally
* Ensure mentioning vertically scaling is a good first step


content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
+---
+title: "Vertical vs Horizontal scaling in CHT 4.x"
```
garethbowen, line 2:
```suggestion
title: "Vertical vs Horizontal scaling"
```

The first sentence of the intro specifies horizontal was introduced in 4.0.0.

content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
...
+    The power of clustered CouchDB to horizontally scale the CHT  
+relatedContent: >  
+    apps/guides/hosting/4.x/self-hosting-multiple-nodes
+    apps/guides/hosting/4.x/data-migration
+    core/overview/architecture/
+---
+
+## Introduction
+
+CHT Core 4.0.0 introduces [a new architecture]({{< relref "core/overview/architecture" >}}) for hosting.  Specifically, 4.0.0 gives the ability to add more servers, known as horizontally scaling, so that large deployments can support more users.  This is a big improvement over naively adding more RAM and CPU to a single server where the CHT Core is running, known as vertically scaling.
```
garethbowen, line 15: Is it worth specifically mentioning CouchDB servers, at least until we support horizontally scaling other services?

```suggestion
CHT Core 4.0.0 introduces [a new architecture]({{< relref "core/overview/architecture" >}}) for hosting.  Specifically, 4.0.0 gives the ability to add more CouchDB servers, known as horizontally scaling, so that large deployments can support more users.  This is a big improvement over naively adding more RAM and CPU to a single server where the CHT Core is running, known as vertically scaling.
```

garethbowen, line 15: This sentence is hard to read. Perhaps pull out definitions of horizontal and vertical scaling first, and then explain how that applies to the CHT?

content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
...
+
+When thousands of users are simultaneous trying to synchronize with the CHT, the concurrency can sometimes overwhelm CouchDB. As discovered [through extensive research](https://forum.communityhealthtoolkit.org/t/how-we-tested-scalability-of-cht-infrastructure/1532) and [large production deployments](https://github.com/medic/cht-core/issues/8324#issuecomment-1691411542), administrators will start to see errors in their logs and end users will complain of slow sync times.  Before moving to more CouchDB nodes, administrators should consider adding more RAM and CPU to the single server where the CHT is hosted. This applies to both CHT 3.x and CHT 4.x. Given the easy of allocating more resources, presumably in virtualized environment like [EC2](https://aws.amazon.com/ec2/), [Proxmox](https://www.vmware.com/content/vmware/vmware-published-sites/us/products/esxi-and-esx.html.html) or [ESXi](https://www.vmware.com/content/vmware/vmware-published-sites/us/products/esxi-and-esx.html.html), this is much easier than moving to [from a single to multi-node CouchDB instance]({{< relref "apps/guides/hosting/4.x/data-migration" >}}).
```
garethbowen, line 21:
```suggestion
When thousands of users are simultaneously trying to synchronize with the CHT, the load can overwhelm CouchDB. As discovered [through extensive research](https://forum.communityhealthtoolkit.org/t/how-we-tested-scalability-of-cht-infrastructure/1532) and [large production deployments](https://github.com/medic/cht-core/issues/8324#issuecomment-1691411542), administrators will start to see errors in their logs and end users will complain of slow sync times.  Before moving to more CouchDB nodes, administrators should consider adding more RAM and CPU to the single server where the CHT is hosted. This applies to both CHT 3.x and CHT 4.x. Given the easy of allocating more resources, presumably in virtualized environment like [EC2](https://aws.amazon.com/ec2/), [Proxmox](https://www.vmware.com/content/vmware/vmware-published-sites/us/products/esxi-and-esx.html.html) or [ESXi](https://www.vmware.com/content/vmware/vmware-published-sites/us/products/esxi-and-esx.html.html), this is much easier than moving to [from a single to multi-node CouchDB instance]({{< relref "apps/guides/hosting/4.x/data-migration" >}}).
```

content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
...
+
+Here we see a normal deployment following our [bare minimum]({{< relref "apps/guides/hosting/requirements" >}}) hosting requirements for hosting the CHT.  We'll call this a "short" deployment because it is not yet vertically scaled:
```
garethbowen, line 23:
```suggestion
Here we see a normal deployment following the bare minimum [hosting requirements]({{< relref "apps/guides/hosting/requirements" >}}) for the CHT.  We'll call this a "short" deployment because it is not yet vertically scaled:
```

content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
...
+flowchart TD
+
+subgraph couch2[" CouchDB - Single ''tall'' Node "]
+    couchInner2["16 CPU/16 GB RAM"]
+end
+
+API["API"] --> HAProxy -->  couch2
+```
+
+Since both CHT 3.x and 4.x support this, vertical scaling is an easy, good first step in addressing performance issues in the CHT.
```
garethbowen, line 47: Consider also adding a sentence for autoscaling which is one of the main benefits of vertical over horizontal scaling, but leave the details as an exercise for the reader based on their host.

content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Horizontal scaling 4.x
+
```
garethbowen, line 50: I think we need to add something about what to do if you're starting a new project as opposed to migrating. Right now the doc reads like you should start with single node, then vertically scale, then horizontally scale, but if you know your instance needs to support many thousands of users we'd recommend horizontally scaling from day 1 to avoid the migration later (eg: Uganda MoH).

garethbowen (changes requested):
Thanks for kicking this off! I didn't get around to running it locally so I can't vouch for the diagrams, but will do that in the next round...

content/en/apps/guides/hosting/vertical-vs-horizontal.md
mrjones-plip: @craig-landry  gave the feedback that "CouchDB" is a bit scary when you're just trying to get your head around all this crazy scaling business.  I'll push back on this suggestion, but we can re-visit if you feel strongly!

content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
...
+When:
+
+* you can no longer vertically scale your CHT instance because of hardware limitations
+* vertically scaling stops yielding better performance
```
kennsippell, line 54: internally i believe we don't scale beyond `r4.8xlarge` instance sizes. this could be useful info for others

content/en/apps/guides/hosting/vertical-vs-horizontal.md
```diff
@@ -0,0 +1,85 @@
...
+subgraph couch2[" CouchDB - Single ''tall'' Node "]
+    couchInner2["16 CPU/16 GB RAM"]
+end
+
+API["API"] --> HAProxy -->  couch2
+```
+
+Since both CHT 3.x and 4.x support this, vertical scaling is an easy, good first step in addressing performance issues in the CHT.
+
+## Horizontal scaling 4.x
```
kennsippell, line 49: is it useful to mention that horizontal scale with the CHT means downtime vs vertical scale can be done without disruption to users? maybe i missed this

content/en/apps/guides/hosting/vertical-vs-horizontal.md
mrjones-plip: gotcha! I tried to split this up in 7f9f2ce  - lemme know how i did!

content/en/apps/guides/hosting/vertical-vs-horizontal.md
mrjones-plip: Vertical auto-scaling is not that big of a win for us, and is my understanding you need to A) be running k8s and friends and B) have a VM/node/thingy already pre-provisioned which has the more cores/ram that you'd need to add.  As such, largely a manual process.

otherwise, we don't support any horizontal auto-scaling and don't think this is (currently) a benefit of any orchestrator.  while we may get there in the future, [I'd really like wait until then ](https://medic.slack.com/archives/CBQH2HNJC/p1692819508799289?thread_ts=1692808476.137739&cid=CBQH2HNJC) (private slack link) to add this language as one of our 4.x benefits.

content/en/apps/guides/hosting/vertical-vs-horizontal.md
mrjones-plip: good idea! fixed in 647f818

content/en/apps/guides/hosting/vertical-vs-horizontal.md
mrjones-plip: ah cools!  That's 32 cores and 244 ram, so put that as a generic stat per 8b6a889  - that work?

content/en/apps/guides/hosting/vertical-vs-horizontal.md
mrjones-plip: oh - hello @kennsippell !  So glad you stopped by. Great question!

Honest question: Does it really matter because we're 100% offline first?  I ask this because you're a lot closer to our end users and I wouldn't want to say this unless it was true.  So if the view warming takes 3 days, but the the data transfer and changing internal load balancer only takes 3 hours, I'd say that's 3 hours of downtime for an offline first app, and likely that doesn't matter to a CHW synchronizing once a week.

But maybe in practice we'd need to call this out for reasons not covered above?

garethbowen (approved):
Nice! Diagrams look good too. Thanks!

content/en/apps/guides/hosting/vertical-vs-horizontal.md
mrjones-plip: reflecting on this, I think we can improve the messaging as one takes almost no time (likely just a container/vm reboot) and the other takes hours of app admin effort, at a minimum, and some down time.  Thanks again!

see ff73e85

PR #1176, opened 2023-09-07
Add recommended phone specifications

derickl:
This adds a baseline recommended phone spec based on what we've been recommending for deployments.


garethbowen (approved):
Perfect.

PR #1178, opened 2023-09-13
Be explicit about moving shards being manual

dianabarsan:


content/en/apps/guides/hosting/4.x/data-migration.md
```diff
@@ -195,18 +195,18 @@ EOF
...
 ```shell
-cd ~/couchdb-migration/ 
+cd ~/couchdb-migration/
 shard_matrix=$(docker-compose run couch-migration generate-shard-distribution-matrix)
 docker-compose run couch-migration shard-move-instructions $shard_matrix
-``` 
+```
 
-j) Follow the instructions from the step above and move the shard files to the correct location, according to the shard distribution matrix. 
+j) Follow the instructions from the step above and move the shard files to the correct location, according to the shard distribution matrix. This is a manual step that requires to physically move data around on disk.
```
henokgetachew, line 209: My recommendation is to use `rsync` with the flags `-avz --progress`. The `-a` ensures that symbolic links, devices, attributes, permissions, ownership, etc., are preserved in the copied files. The `v` just for verbose mode and the `z` to send the files in a compressed format to preserve bandwidth. The `--progress` is good because you often wonder during file transfers if things are really working or if it's stuck somewhere.

henokgetachew (approved):
I have added one minor comment about using `rsync`. I have had issues with `scp`. I have also suggested important flags to use with `rsync`.

content/en/apps/guides/hosting/4.x/data-migration.md
dianabarsan: I think this will heavily depend on deployment. Some deployments hold all data locally,  maybe rsync doesn't make sense for those.

garethbowen:
@dianabarsan Looks like this is ready for merge, unless you wanted to change something based on the comment?

dianabarsan:
I don't think we're ready to include that sort of detail in that documentation. Merging.

PR #1181, opened 2023-09-15
Adds sections on purging and CouchDb 3 to 4.4 release notes

dianabarsan:


content/en/core/releases/4.4.0.md
```diff
@@ -52,9 +52,32 @@ As part of this work, some translation keys changed. We have provided translatio
...
+3. Adding a new `_prometheus` endpoint, which exposes CouchDb metrics and can be used out of the box in monitoring visualisation tools, like the CHT Watchdog.
+4. Improved security, where database creation and writes/reads are no longer allowed for all users by default.
+5. Improved security, where only CouchDb admins are allowed to add/edit/delete users.
+6. Dramatically improved multi-core CPU usage, resulting in better performance when running on powerful machines. This results in a 1.5x improvement of request time across the board, with some endpoints being up to 10x faster.
+
+### Replication performance improvement by persistent caching of purged doc ids
+
+CouchDb request performance analysis has revealed that approximately 25% of CPU time was used querying to determine which documents have been purged for users that are replicating.
+Because purging is an action that runs on a schedule, relatively infrequently due to how performance expensive it is, caching results from these very expensive queries significantly improve server-side performance.
```
garethbowen, line 71:
```suggestion
Because purging is an action that runs on a schedule, relatively infrequently due to how performance expensive it is, caching results from these very expensive queries significantly improves server-side performance.
```

content/en/core/releases/4.4.0.md
```diff
@@ -52,9 +52,32 @@ As part of this work, some translation keys changed. We have provided translatio
...
+
+As of 4.4.0, the CHT caches the purged doc ids for each user that replicates. The cache is reset when purging runs next time and when user replication settings change. This means that, for every connected user, purged docs are only queried once per purging timeframe, when the user first syncs or replicates.
+
+| Requests URL | Request Count | Response time | Sum Response time sum percent |
```
garethbowen, line 75: I think this would be easier to follow if it was "average response time" rather than cumulative. Then we can just drop the "request count" column.

content/en/core/releases/4.4.0.md
```diff
@@ -52,9 +52,32 @@ As part of this work, some translation keys changed. We have provided translatio
...
+| -- | -- | -- | -- |
+| /medic/_design/medic/_view/docs_by_replication_key | 103503	| 8322(h) |	46.34% |
+| /medic-purged-role-.*/_changes  | 105890	 | 4815(h) | 	26.82% |
```
garethbowen, line 78: It's worth emphasizing that this is the row that has changed or been eliminated on subsequent requests.

garethbowen (changes requested):
Amazing. This makes me want to upgrade  :D 

Suggestions inline.

tatilepizs (approved):
Thank you for the details.

content/en/core/releases/4.4.0.md
dianabarsan: I bolded it.

PR #1182, opened 2023-09-15
fix broken link for dbt install

mrjones-plip:
should fix infinite redirect [per this check](https://github.com/medic/cht-docs/commit/ad3e363f387156878d816a9768963b955dd18161/checks)


andrablaj (approved):
👌

PR #1195, opened 2023-10-02
Run link checker only on Sundays

mrjones-plip:
We get a lot of timeouts in the link checker.

I think because we check so often (every 24 hrs) we catch a lot of websites that are down. This greatly increases our false positives and reduces the trust in this link check script.  With all the noise in the channel, we end up ignoring the reports.

To reduce the noise and to hopefully reduce the false positives, I'm going to reduce the frequency of this script being run to 1/week instead of 1/day.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -9,8 +9,8 @@
 #    the default or base branch
 on:
   schedule:
-    # runs at 3:15am every day
-    - cron:  '15 3 * * *'
+    # runs at 3:15am every Sunday
+    - cron:  '15 3 * * 0'
 
 name: Test site for broken links
 jobs:


nydr (approved):
👍 Could be worth trying as a workaround, it would at minimum reduce the frequency.

Looking at the errors it seem to be mostly pages being down, or access denied, presumably because it's hitting some bot detection or rate limiting issue from GH servers.

Other options I can think of:
1. Pick an even [more random](https://www.random.org/clock-times/?num=5&earliest=03%3A00&latest=06%3A00&interval=1&format=html&rnd=new) time
2. Save state and only error if the same site is down twice in a row
    - This would increase the complexity of the tests
3. Do a retry of the failed pages an hour or so later
    - This might use up a lot of CI runtime, depending on how it is structured

mrjones-plip:
Ohhh!  Those three are good suggestions.  We could probably log failed URLs in [the script](https://github.com/medic/cht-docs/blob/main/.github/scripts/muffet.sh) and then retry them with a back off of 2X minutes, say 4 iterations maybe? (eg 2,4,8,16). 

Let's see how the 1 week scrape does and remember these good ideas as needed.

PR #1197, opened 2023-10-03
Add release notes for multiple backports

garethbowen:
medic/cht-core#8593
medic/cht-core#8594
medic/cht-core#8595
medic/cht-core#8596


dianabarsan (approved):
Very nice!

PR #1198, opened 2023-10-04
Use asdf to specify hugo dependencies and versions

nydr:


.github/workflows/daily.yml
```diff
@@ -25,24 +25,41 @@ jobs:
         with:
           fetch-depth: 0         # Fetch all history for .GitInfo and .Lastmod
 
+      - name: Parse .tool-versions for hugo-version
+        id: hugo-version
+        run: |
+          grep 'hugo extended_' .tool-versions |
```
garethbowen, line 31: I'm not overly familiar with asdf, but it looks like they publish [github actions](https://github.com/asdf-vm/actions) to install tools which not only saves us from grepping the file but might also help simplify the rest of the file. Is that useful?

garethbowen (commented):
Thanks!

One suggestion inline.

.github/workflows/daily.yml
nydr: That's amazing! I'll check them out and update PR if it can be used

garethbowen (approved):
Awesome, thanks!

PR #1205, opened 2023-10-16
Fix muffet workflow after asdf update

nydr:


nydr:
Issue introduced in #1198

mrjones-plip (approved):
ah  - nice fix!

:ship:

PR #1214, opened 2023-10-23
chore: Exclude Baserow demo from link checker

andrablaj:
Related to [this build fail](https://github.com/medic/cht-docs/commit/281eaa7084677c98ac531386dc225f28bf4039b4/checks).


nydr (approved):
Thanks for sorting this out!

PR #1229, opened 2023-11-07
improve 4.4 performance table -> chart

mrjones-plip:
# Description

This PR changes a table to a graph to better illustrate the time savings of the purge cache improvements:

Before:
image

After:

image

# License


content/en/core/releases/4.4.0.md
```diff
@@ -70,14 +70,11 @@ Most notable features are:
...
 
-As of 4.4.0, the CHT caches the purged doc ids for each user that replicates. The cache is reset when purging runs next time and when user replication settings change. This means that, for every connected user, purged docs are only queried once per purging timeframe, when the user first syncs or replicates.
-
-| Requests URL | Avg Response time (min) | Sum Response time sum percent |
-| -- | -- | -- | -- |
-| /medic/_design/medic/_view/docs_by_replication_key 	| 4.8 |	46.34% |
-| **/medic-purged-role-.*/_changes**   | **2.7** | 	**26.82%** |
-| /medic/_design/medic/_view/contacts_by_depth 	 | 0.2 | 	3.42% |
-| /medic/_changes   | 0.4 |  3.26% |
+As of 4.4.0, the CHT caches the purged doc ids for each user that replicates. The cache is reset when purging runs next time and when user replication settings change. This means that, for every connected user, purged docs are only queried once per purging timeframe, when the user first syncs or replicates:
```
garethbowen, line 73: I think we need to call out that this performance improvement doesn't occur on the first replication by the user...

```suggestion
As of 4.4.0, the CHT caches the purged doc ids for each user that replicates. The cache is reset when purging runs next time and when user replication settings change. This means that, for every connected user, purged docs are only queried once per purging timeframe, when the user first syncs or replicates.

This leads to significant improvements in subsequent replication requests.
```

garethbowen (changes requested):

garethbowen (approved):
Thanks!

PR #1231, opened 2023-11-07
fix: Remove H1 for test page

craig-landry:
The Hugo template takes this from the markdown metadata and prints that as an H1 already. Having it again on the page resulted in a duplicate H1.


jkuester (approved):
LGTM!

PR #1232, opened 2023-11-08
feat(#7462): add doc for new cht-form npm scripts

jkuester:
https://github.com/medic/cht-core/issues/7462
https://github.com/medic/cht-core/pull/8153

See [this thread](https://github.com/medic/cht-core/pull/8153#discussion_r1387234158) for more context regarding the style guide changes around implicit returns.

# License


content/en/contribute/code/style-guide.md
```diff
@@ -248,6 +248,38 @@ const getResult = () =>
...
 
+### Implicit Returns
+
+As [noted above]({{< ref "#use-arrow-functions" >}}), implicit returns should be used for one-line arrow functions. However, for the sake of readability, they should not be used when returning a multi-line value.
+
+*Right:*
+
+```js
+const oneLineString = () => 'World';
+const oneLineObject = () => { hello: 'World' };
```
m5r, line 259: is this supposed to return an object? if so, it should be wrapped with parenthesis  
this code is valid JS but it does not return anything, it defines a [labeled statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/label)

```suggestion
const oneLineObject = () => ({ hello: 'World' });
```

m5r (approved):
LGTM, there is a detail in my comment I'd like to clear up before merging but I'll approve to not block it while I'm offline

content/en/contribute/code/style-guide.md
jkuester: Good catch!

Issue #1240, opened 2023-11-20
Improve 'contact-summary.templated.js'


binokaryg:
There is a typo in the Care Guides section:
> The context **infomation** from the profile is accessible as the variable summary.

Page: /apps/reference/contact-page/
Version: 2023-11-06 12:40:37 -0700 -0700
---

andrablaj:
Fixed via https://github.com/medic/cht-docs/pull/1241.

PR #1241, opened 2023-11-20
fix(#1240): Fix a typo

binokaryg:
Fixed a typo #1240

# Description
There was a typo in the Care Guides section:

> The context **_infomation_** from the profile is accessible as the variable summary.

# License


andrablaj (approved):
Good catch! Thanks 👍

Issue #1095, opened 2023-05-26
Add info about when/how often syncing is attempted


michaelkohn:
Would be useful to document when syncing is attempted.

Here are some relevant resources that could be rolled up and documented in the "[Sync Status](https://docs.communityhealthtoolkit.org/apps/concepts/navigation/#sync-status)" section (or elsewhere). 

- [Forum Post with all (well, most of) the details](https://forum.communityhealthtoolkit.org/t/for-offline-users-how-often-does-the-app-try-to-refresh-if-there-is-an-available-internet-connection/503/2?u=michael )
- https://github.com/medic/cht-core/issues/5738#issuecomment-621934539
- https://github.com/medic/cht-core/issues/6299

latin-panda:
@michaelkohn, I guess what you need is telemetry, right? Is the number of times people sync okay?

michaelkohn:
@latin-panda this is just about **_documenting_** the current sync strategy on the docs site :-) ... I believe the forum post contains most of the details (though I'm not sure if things have changed at all since 2019), I just want that to be on the docs site somewhere.

latin-panda:
Oh okay, easy then. Thanks @michaelkohn

latin-panda:
Ideas to document:
- Internal process
- Different types of sync (manual, automated)
- Type of documents / data
Not sure if to also cover when it prompts to reload the app? (when config changes, when there's a new version of the app (src), when there's a new form, etc)

michaelkohn:
Yep.... some important details we'd want to cover:

1. That offline users have (I think) two databases that get synced (the main one and the meta DB)
2. The types of docs you'll find in each of those DBs (telemetry, feedback, etc... in meta | reports, contacts, etc... in the main DB)
3. When do each of these databases normally get synced (automatically while there's a connection, whenever there is a change, when the app notices it comes back online, etc...)  
4. Any differences in how things are handled if there are server changes vs. client changes 
5. What other things will trigger a sync (logging in, reloading the app, etc....)

> Not sure if to also cover when it prompts to reload the app?

Jenni did some good investigation in [this slack thread
](https://medic.slack.com/archives/C02GME3BQE6/p1648194230211909?thread_ts=1648024569.425659&cid=C02GME3BQE6). I'd love for us to document this as well, but am OK doing it as a separate issue.

Benmuiruri:
Closed by #1235

michaelkohn:
Quick link to where this got updated on the docs site

[...]

PR #1243, opened 2023-11-22
feat(#1095): Update data sync documentation

Benmuiruri:
# Description

Document the current sync strategy on the docs site. The documentation has been updated with information that cover the following topics: 
- data replication
- sync status states
- sync intervals
- sync retry mechanism

#1095 

# License

### DIFF
--- a/content/en/apps/concepts/navigation.md
+++ b/content/en/apps/concepts/navigation.md
@@ -38,9 +38,19 @@ Tap the menu icon in the upper right corner of the header to access other pages,
 
 Data synchronisation is important for [offline users]({{< relref "apps/concepts/users" >}}). These users keep a copy of the data they have access to on their device. They can work from their device while disconnected from the internet (offline), by reading from and writing to their copy of the data. “Sync” (synchronisation) is when data on the device is made to match the data on the server. Synchronization depends on the user's connectivity status. The CHT app monitors the online status and attempts sync accordingly. It adapts to changes in connectivity to optimize data transfer.
 
+#### Replication Types
+
 Synchronisation consists of upward replication and downward replication. 
 - **Upward Replication**: uploading all new or altered data from the device to the server. It includes a retry mechanism for handling larger data batches, ensuring a robust and reliable upload process.
 - **Downward Replication**: downloading new or altered data from the server to the device. Downward replication may include the download of software updates to the CHT core or CHT app when available.
+
+The CHT app has two different Sync Intervals: 
+- **Regular Sync**: Occurs every 5 minutes. Regular sync replicates application data, including user-generated content and any changes to existing documents.
+- **Metadata Sync**: Occurs every 30 minutes, ensuring critical metadata is regularly updated. Metadata sync is designed to synchronize meta-information, which is critical for the app's functionality but not directly related to user-generated content. Metadata Sync replicates system-level data such as configurations, settings, or system updates
+
+The two sync interval values are configurable in the `db-sync.service.ts` file.
+
+#### Sync Status Notification
 
 At the bottom of the menu is a notification which provides important information about data synchronisation.
 
@@ -66,13 +76,6 @@ The synchronization process can be in one of the following states:
 - **Success**: The last sync operation was successful.
 - **Required**: There is data pending synchronization.
 
-#### Replication Types
-
-The CHT app has two different Sync Intervals: 
-- **Regular Sync**: Occurs every 5 minutes. Regular sync replicates application data, including user-generated content and any changes to existing documents.
-- **Metadata Sync**: Occurs every 30 minutes, ensuring critical metadata is regularly updated. Metadata sync is designed to synchronize meta-information, which is critical for the app's functionality but not directly related to user-generated content. Metadata Sync replicates system-level data such as configurations, settings, or system updates
-
-The two sync interval values are configurable in the `db-sync.service.ts` file.
 
 #### Data Synchronization Retry Mechanism 
 
--- a/content/en/apps/concepts/navigation.md
+++ b/content/en/apps/concepts/navigation.md
@@ -36,7 +36,11 @@ Tap the menu icon in the upper right corner of the header to access other pages,
 
 {{< figure src="sync-status.png" link="sync-status.png" class="right col-6 col-lg-4" >}}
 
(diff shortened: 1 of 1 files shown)


Benmuiruri:
Hi @garethbowen , my first draft of updating our Sync documentation is ready for your review to fact-check my understanding of syncing. The write up is based on the [discussion between Michael and Jenni ](https://github.com/medic/cht-docs/issues/1095)

Let me know if there's anything I could add, delete, or correct. 

Thanks.

content/en/apps/concepts/navigation.md
```diff
@@ -52,3 +71,27 @@ If the indicator is red, it means you have data changes waiting to be uploaded t
...
 It will also allow to retry the sync process in case of failure.
+
+The user-initiated sync process initatied by clicking the "Sync now" buttoninvolves uploading local changes to the server and downloading updates from the server to the local database. It syncs both the main database and the metadata database to ensure that both user data and application metadata are up-to-date.
+
+#### Synchronization Triggers
+- **On Login**: Synchronization is automatically initiated upon successful user login if the app is connected to the internet. The CHT app checks for changes since the last known sequence, identifying document revisions that differ from the ones on the server, and sending or receiving bulk document updates.
+- **Manual**: Clicking the "Sync now" button.
+- **Periodic Sync**: The application performs regular checks and background processes that may trigger synchronization if there are local changes to replicate to the server, or changes from the server that need to be replicated to the local database.
+- **On Reload**: Synchronization is automatically initiated when you click the reload button in the "Update available" modal.
+
```
garethbowen, line 82: Also add one more...

```suggestion
- **On Connect**: The app also detects when an internet connection becomes available and attempts to sync immediately.
```

content/en/apps/concepts/navigation.md
```diff
@@ -52,3 +71,27 @@ If the indicator is red, it means you have data changes waiting to be uploaded t
...
+
+#### Sync Status States
+The synchronization process can be in one of the following states:
+
+- **Unknown**: The sync status is not determined yet.
+- **Disabled**: Sync is disabled, typically when the user is in online-only mode.
```
garethbowen, line 88:
```suggestion
- **Disabled**: Sync is disabled - only applies to online-only users.
```

content/en/apps/concepts/navigation.md
```diff
@@ -52,3 +71,27 @@ If the indicator is red, it means you have data changes waiting to be uploaded t
...
+- **InProgress**: Synchronization is currently ongoing.
+- **Success**: The last sync operation was successful.
+- **Required**: There is data pending synchronization.
+
+
+#### Data Synchronization Retry Mechanism 
+
+The CHT app has a retry mechanism to handle replication failures, particularly due to forbidden documents and challenging network environments with restricted data access. CHT core is configured to attempt replication retry up to a maximum of 3 times for each document.
+
```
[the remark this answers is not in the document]
Benmuiruri: I inferred it from the [db-sync-retry.service.ts](https://github.com/medic/cht-core/blob/d93b9162e21b180b1c920d58a6890812ae3ea9e7/webapp/src/ts/services/db-sync-retry.service.ts#L5C26-L5C26) file. 

Was my understanding incorrect ?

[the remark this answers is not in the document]
garethbowen: Ahh I see! I think this is an implementation detail and not worth complicating these docs for. Essentially if the replication happens in the wrong order you might sync a report that the server thinks you shouldn't have access to. If that happens we change the order and try again.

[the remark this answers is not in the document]
Benmuiruri: Got it, in that case let me delete that section.

content/en/apps/concepts/navigation.md
```diff
@@ -36,9 +36,27 @@ Tap the menu icon in the upper right corner of the header to access other pages,
...
 
-At the bottom of the menu is a notification which provides important information about data synchronisation.
+#### Replication Types
+
+Synchronization consists of upward replication and downward replication. 
+- **Upward Replication**: Uploading all new or updated data from the device to the server. It includes a retry mechanism for handling larger data batches, ensuring a robust and reliable upload process.
+- **Downward Replication**: Downloading new or updated data from the server to the device. Downward replication may include the download of software updates to the CHT app when available.
+
+The CHT application manages data synchronization across two types of databases: the main application database and a meta database.
```
garethbowen, line 47:
```suggestion
The CHT application manages data synchronization across two types of databases:
```

I don't think this is necessary because we list them out next.

content/en/apps/concepts/navigation.md
```diff
@@ -36,9 +36,27 @@ Tap the menu icon in the upper right corner of the header to access other pages,
...
+- **User-Metadata Database(`medic-user-{username}-meta`)**: Each user has a dedicated database that stores operational metadata, including [telemetry data]({{< relref "apps/guides/performance/telemetry" >}}) and error messages. Synchronization occurs at predefined intervals to ensure up-to-date monitoring and analysis.
+
+The CHT application has two different Sync Intervals: 
+- **Regular Sync**: Occurs every 5 minutes. It replicates application data, including user-generated content and any changes to existing documents.
+- **Metadata Sync**: Occurs every 30 minutes, focusing on operational metadata rather than user-generated content. This ensures that telemetry data and system logs are regularly synchronized, providing insights into system performance and user activities
```
garethbowen, line 54: We should keep the names of the two databases consistent rather than using "regular" here. Suggestion:

```suggestion
The two databases sync on different schedules: the main application database syncs every 5 minutes, while the user metadata database syncs every 30 minutes.
```

garethbowen (changes requested):
Nice!

I have two more suggestions, and two left over from the original review.

content/en/apps/concepts/navigation.md
Benmuiruri: I deleted the content all-together because the sync interval is mentioned below in the Synchronization Triggers 

`- **Periodic Sync**: The application performs regular checks and attempts to synchronize. The main application database syncs every 5 minutes, while the user metadata database syncs every 30 minutes.`

garethbowen (approved):
Great work! Ship it!

PR #1244, opened 2023-11-24
chore: remove S2699 sonar rule that requires tests to have assertions

dianabarsan:
# Description

# License


jkuester (approved):
I believe that global assignments of things like `expect` are actually the root cause of the issue here and (since it leads to inconsistencies when requiring `chai`, which is where Sonar ran into problems with this rule...).

That being said, I will not block this change as I know that my opinions here are not universal. Also, I feel that this particular rule is of only limited use anyway...

PR #1251, opened 2023-12-01
move new k3s to be next to other self hosting, update titles to match

mrjones-plip:
# Description

With k3s hosting documentation having been just added, this PR moves it to be next to the other self hosting docs.  As well it updates the titles so they're have parity.  Finally it adds a redirect (alias) to ensure we don't loose any traffic.

# License


mrjones-plip:
@esthermmoturi  - when you get a sec!  I didn't realize that @nsitaula was on holiday.

esthermmoturi (approved):
LGTM!

PR #1253, opened 2023-12-11
fix(#): consistent naming for public key file

kennsippell:
# Description

Appears this should be cert.pem.  [This line](https://github.com/medic/cht-docs/blob/main/content/en/apps/guides/hosting/4.x/adding-tls-certificates.md?plain=1#L46) uses cert.pem and it is the default [here](https://github.com/medic/cht-core/blob/34edb11f7f5b9a2f945945387e70a762098c8148/nginx/ssl-install.sh#L5). Docker helper maybe uses chain.pem but I don't think this is the default for cht-core self-host. 

# License


mrjones-plip (approved):
nice catch - thanks @kennsippell !

PR #1263, opened 2024-01-15
fix: Fix broken links

andrablaj:
# Description

Tackles [these link errors.](https://github.com/medic/cht-docs/commit/b1b0f959843cd4a0266762c9984526388d35c723/checks) 

# License

### DIFF
--- a/content/en/apps/tutorials/localizing-cht.md
+++ b/content/en/apps/tutorials/localizing-cht.md
@@ -71,7 +71,7 @@ Populate the `messages-sw.properties` file with the appropriate translation stri
 ```
 cht --local upload-custom-translations
 ```
-The default Swahili translations that come pre-added to CHT can be found [here](https://github.com/medic/cht-core/blob/ac573692f449a36220e108e759bd86d9e6c1087e/config/standard/translations/messages-sw.properties).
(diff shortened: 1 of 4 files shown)


content/en/apps/tutorials/localizing-cht.md
```diff
@@ -71,7 +71,7 @@ Populate the `messages-sw.properties` file with the appropriate translation stri
 ```
 cht --local upload-custom-translations
 ```
-The default Swahili translations that come pre-added to CHT can be found [here](https://github.com/medic/cht-core/blob/master/config/standard/translations/messages-sw.properties).
+The default Swahili translations that come pre-added to CHT can be found [here](https://github.com/medic/cht-core/blob/ac573692f449a36220e108e759bd86d9e6c1087e/config/standard/translations/messages-sw.properties).
```
nydr, line 74: Optional: Would it make sense to use a reference to a version branch instead of commit? For example: https://github.com/medic/cht-core/blob/4.5.x/config/standard/translations/messages-sw.properties

nydr (approved):
Thanks for going through these, LGTM!

content/en/apps/tutorials/localizing-cht.md
andrablaj: Yes, you are right, that is cleaner! Updated 💯

PR #1267, opened 2024-01-17
feat(sonar): recommend setting NCD to 1 day

jkuester:
# Description

Sonar analysis [failed](https://sonarcloud.io/summary/new_code?id=medic_cht-core&branch=master) on the latest commit to cht-core `master`.  The [PR](https://github.com/medic/cht-core/pull/8744) that actually added the code causing the issue was merged 9 days ago.  However, at the time the PR was merged, it passed the Sonar check!  It turns out, Sonar added [a new rule](https://sonarcloud.io/organizations/medic/rules?open=typescript%3AS1444&rule_key=typescript%3AS1444) today and failed that code [ex post facto](https://en.wikipedia.org/wiki/Ex_post_facto_law) because it was considered to be "new" by our Sonar New Code Definition for cht-core (which is currently set to consider all changes from the last `30 days` as "new").

My proposal is to lower our NCD value to just `1 day` to avoid these issues in the future.

# License

### DIFF
--- a/content/en/contribute/code/static-analysis.md
+++ b/content/en/contribute/code/static-analysis.md
@@ -52,9 +52,11 @@ When Sonar flags an issue with the code in your pull request, use this decision 
 
 #### New Code Definition
 
-When setting up a new repository in SonarCloud, you will be asked to define what is considered to be "new code". This is used to determine which code in the default branch is considered "new" (affects reporting of issues, etc).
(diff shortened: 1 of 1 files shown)


m5r (approved):

m5r:
Just to be clear, has the NCD been lowered to 1 day for cht-core already?

jkuester:
I have not made any changes to Sonar yet.  For the config like this that has to be manually set on the Sonar UI, my general approach is to do this docs change (as a way of recording our decisions and discussion around the change) and then make the change in the UI once this PR is approved/merged.

m5r (approved):

PR #1280, opened 2024-01-31
chore(NA): Improve magic link docs

mrjones-plip:
# Description

Improve magic link docs

# License


content/en/apps/concepts/access.md
```diff
@@ -58,13 +58,13 @@ Users may log out by going to the options menu available in the top right corner
...
+When creating users, the admin has the option to enable a user to login in by simply clicking a link sent via SMS. The link is only valid for 24 hours and can only be used once to log in. This ensures the link is used only by the intended recipient. By clicking the magic link, the user is logged into their project's instance directly, bypassing the need to manually enter a username and password.  
 
-By clicking the magic link to log in, the user is able to enter their project's instance directly, bypassing the need to enter their username and password. If the app is not installed on their phone, it will open in their default browser.
+With token login, the password is never known by the admin or the user because the password is changed to a random string after every successful token login. If the user needs to login again, they need to contact the admin so that the admin can either send a new magic link or revert their account back to using a manual login and password.
```
garethbowen, line 65: "revert" sounds technical and scary.

```suggestion
With token login, the password is never known by the admin or the user because the password is changed to a random string after every successful token login. If the user needs to login again, they need to contact the admin so that the admin can either send a new magic link or switch their account back to using a manual login and password.
```

content/en/apps/concepts/access.md
```diff
@@ -58,13 +58,13 @@ Users may log out by going to the options menu available in the top right corner
...
 
-To recover a password, the user needs to contact the admin so that they may regenerate a new magic link and repeat the workflow. 
+When the token login link is clicked and the app is not installed on the user's phone, it will open in their default browser. If no gateway is set up on the CHT server, the message may be sent via another messaging app.
```
garethbowen, line 67: I think it'd be better to move the second sentence up to line 63 immediately following the sentence "When creating users, the admin has the option to enable a user to login in by simply clicking a link sent via SMS.". It seems to fit the context there better.

garethbowen (changes requested):
A couple of quick suggestions.

content/en/apps/concepts/access.md
mrjones-plip: word!  fixed in 4eba15d5

garethbowen (approved):
Awesome, thanks!

PR #1285, opened 2024-02-06
chore(na) Simplify macos steps to always build couchdb image

mrjones-plip:
# Description

This PR reduced the number of manual steps that Apple Silicon users need to take when setting up the CHT Core dev env.  While it sends both Intel  and Apple Silicon user through these steps, it's a safe abstraction (technically all users could take this route).

The `npm run local-images` is a very heavy call, building every image for every service.  We only want the couchDB one.  As well, we don't want different env vars and docker files for macOS folks - this PR keeps all that standard.  

With less steps and less setup needed, I believe this will mean more success for new devs first getting setup!

# License

### DIFF
--- a/content/en/contribute/code/core/dev-environment.md
+++ b/content/en/contribute/code/core/dev-environment.md
@@ -125,7 +125,8 @@ echo $COUCH_NODE_NAME && echo $COUCH_URL
 
 Create a `docker-compose.yml` and `couchdb-override.yml` files under the `~/cht-docker` folder with this code:
 
-```
+{{< tabpane persistLang=false lang=shell >}}
+{{< tab header="Linux (Ubuntu) & Windows (WSL2)" >}}
 mkdir -p ~/cht-docker
 curl -s -o ~/cht-docker/docker-compose.yml https://staging.dev.medicmobile.org/_couch/builds_4/medic:medic:master/docker-compose/cht-couchdb.yml
 cat > ~/cht-docker/couchdb-override.yml << EOF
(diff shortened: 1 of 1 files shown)


mrjones-plip:
@JenniferMuli  @esthermmoturi  and @andrablaj - I know you both have been working on this so mentioning y'all for awareness of this change.  I also note this was a recent change to the docs and there's an active forum post [discussing](https://forum.communityhealthtoolkit.org/t/couchdb-installation-issue-on-apple-silicon-chip/3205/35) this!

content/en/contribute/code/core/dev-environment.md
```diff
@@ -136,54 +135,29 @@ services:
           - "5984:5984"
           - "5986:5986"
 EOF
-```
+{{< /tab >}}
+{{< tab header="macOS" >}}
+docker build -t couchdb-apple-silicon ~/cht-core/couchdb/.
```
andrablaj, line 140: This command only works if `cht-core` is cloned in the `~/` directory. Can we specify that? Or advise the readers to `cd` into `cht-core` and run `docker build -t couchdb-apple-silicon couchdb/.` instead?

andrablaj (changes requested):
This is so much simpler now, thanks @mrjones-plip! 👏  I have a small comment, but otherwise everything worked correctly. 

I committed directly the removal of sone text mentioning Apple Silicon section, I hope that's OK.

content/en/contribute/code/core/dev-environment.md
mrjones-plip: Ah - good thinking! The whole guide is predicated on using the default location of `~/cht-core/` - I'll add a heads up that this continues to use that.  

The guide is meant to be followed all in one go with a "happy path" of `~/cht-core/`, but we can call this out for folks who didn't notice.

andrablaj: Perfect, then just removing the `~/cht-core/` will help, as the readers will already be in the `cht-core` directory. I will add a suggestion.

andrablaj, line 140:
```suggestion
docker build -t couchdb-apple-silicon couchdb/.
```

andrablaj (commented):
Cool, this works out of the box if the reader follows the guide and clones `cht-core` to `~/`. I skipped that step as I had already the repo cloned elsewhere, hence the need for me to edit the path mentioned above.

andrablaj (approved):
:shipit:

PR #1303, opened 2024-03-07
Direct to client two way texting workflows on cht

nsitaula:
Could you please review direct-to-patient workflow documentation? which is in:
cht-docs-> content -> en -> apps -> examples


EvelynWaweru (approved):
Thanks @nsitaula, the documentation looks good to me.

PR #1308, opened 2024-03-13
feat: update Android dev environment docs to note Java 17

jkuester:
# Description

After https://github.com/medic/cht-android/pull/349, Java 17 will be required to build cht-android.  Also added documentation about how to upgrade the Android System Webview version in an emulated Android device.

# License

### DIFF
--- a/content/en/contribute/code/android/development-setup.md
+++ b/content/en/contribute/code/android/development-setup.md
@@ -176,6 +176,12 @@ To upgrade the necessary apk on an emulated Android device:
     - For Android 10+, you should download a new version of `com.google.android.webview` (e.g. from [apkMirror](https://www.apkmirror.com/apk/google-inc/android-system-webview/)).
 - Then, you can use adb to install the apk into the device: `adb install -r *your_apk*.apk`
 
+{{% alert title="Note" %}}
+Testing using an Android 5-6 device can be more challenging. The `com.android.webview` package seems to be baked into the system partition regardless of the image variant. It is possible use a custom ROM (e.g. CyanogenMod) with an upgraded WebView version on a physical device. Another approach to testing with older versions of Android is to simply test against an older version of the CHT (e.g. CHT `3.x.` only requires Chrome/Webview 53). The functionality of cht-android can still be validated by interacting with this old CHT version.
+
+If using the docker-helper to deploy CHT instances for testing with old Android versions, be aware that the `local-ip.medicmobile.org` SSL certificates may not work with the device (because the root certificate is not recognized). On Android 5-6 you can manually install the root certificate via Settings > Security > Install from SD Card. 
+{{% /alert %}}
+
 ### Android Studio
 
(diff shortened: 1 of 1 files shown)


m5r (approved):
Coming back to this, requesting some changes :eyes:

content/en/contribute/code/android/development-setup.md
```diff
@@ -162,6 +162,20 @@ _Only CHT Android_
...
+
+To upgrade the necessary apk on an emulated Android device:
+
+- Start by using the `Google APIs` variant of the Android image. The base AOSP image bakes in the `com.android.webview` package, making it difficult to update.
+- Depending on your version of Android, you will need to manually download a new version of the apk (make sure to get the correct variant for your architecture):
+    - For Android 7-9, you should download a new version of `com.android.chrome` (e.g. from [apkMirror](https://www.apkmirror.com/apk/google-inc/chrome/)).
+    - For Android 10+, you should download a new version of `com.google.android.webview` (e.g. from [apkMirror](https://www.apkmirror.com/apk/google-inc/android-system-webview/)).
```
m5r, line 176: What's the steps for Android 5-6? I've tried installing APKs of both Chrome and Android System WebView, both versions 90+, and the CHT app would freeze on the loading screen with the following error being logged by the browser:
```
https://gamma.dev.medicmobile.org/scripts.js:1 | Uncaught SyntaxError: Unexpected token {
```
The emulated device is running Android 5.0 with Google Play Services, I installed the APKs from APKMirror using `adb` like shown below

The funny thing is that opening the same instance directly in Chrome worked...

m5r (changes requested):

content/en/contribute/code/android/development-setup.md
jkuester: Yep. After hours of tinkering I have not come up with any solution for Android 5-6.  For these Android versions, it seems like even the "Google API's" images have the `com.android.webview` package baked in. Technically, it should be possible to upgrade that package, but I guess it might require actually editing the system APKs (e.g. the device carrier releasing an updated ROM with a newer WebView).  It seems of dubious value to invest much more effort into getting this to work since I am a bit skeptical of finding a device in the wild that is Android 5 with WebView 90+....

_However_, looking at the [support matrix](https://docs.communityhealthtoolkit.org/core/releases/), it does seem like a more likely use-case we should perhaps test for is connecting the latest cht-android on an Android 5 device to a CHT `3.x` instance. Theoretically we would just need WebView 53+ for this (though it is worth noting that the emulated Android 5 device seems to still only have WebView 39....). 

I am going to try and do some more testing and maybe see about adding some more docs here....

m5r: I gave it a shot on my HTC One S running Android 5 (rooted cyanogenmod) & WebView 95 and it's working normally with a CHT 4.5 instance. Not sure why the HTC picks up on the WebView APK I installed but not the emulator though...

jkuester: > cyanogenmod

:star_struck: Ahh the good ol days!

jkuester: Okay, I have added a bit more info about testing with older Android versions.  Let me know if you think there is anything else we can clarify here!

m5r (approved):
All good! Thanks for adding the notes for older Android versions

PR #1310, opened 2024-03-18
feat: add doc for new countdown timer form configuration

jkuester:
# Description

https://github.com/medic/cht-core/issues/8681

Update documentation for the `countdown-timer` widget to specify how to use the new `trigger` functionality.  

cc @SheilaAbby

# License

### DIFF
--- a/content/en/apps/reference/forms/app.md
+++ b/content/en/apps/reference/forms/app.md
@@ -92,7 +92,15 @@ Calendar widget using Bikram Sambat calendar, which is used by default for appro
 {{< see-also page="apps/reference/forms/app" title="`to-bikram-sambat` XPath function" anchor="to-bikram-sambat" >}}
 
 ### Countdown Timer
-A visual timer widget that starts when tapped/clicked, and has an audible alert when done. To use it create a `note` field with an `appearance` set to `countdown-timer`. The duration of the timer is the field's value, which can be set in the XLSForm's `default` column. If this value is not set, the timer will be set to 60 seconds.
+
+A visual timer widget that starts when tapped/clicked, and has an audible alert when done. To use it, first make sure you have the [`namespaces` column](https://getodk.github.io/xforms-spec/#namespaces) in the "settings" tab of your XLSForm populated with a value that includes `cht=https://communityhealthtoolkit.org`. Then, you can add the timer as a `trigger` field with the _appearance_ set to `countdown-timer`. The duration of the timer can be set in a column named _instance::cht:duration_ (the default value is 60 seconds). 
+
+If you want to make the timer mandatory so users must wait for the timer to complete before continuing to the next page or submitting the form, you can populate the _required_ column with an XPath expression as you would do for any other required question. A value of `"OK"` will be set for the `trigger` field when the timer completes.
+
+{{% alert title="Note" %}}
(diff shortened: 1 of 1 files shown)


content/en/apps/reference/forms/app.md
```diff
@@ -92,7 +92,15 @@ Calendar widget using Bikram Sambat calendar, which is used by default for appro
 {{< see-also page="apps/reference/forms/app" title="`to-bikram-sambat` XPath function" anchor="to-bikram-sambat" >}}
 
 ### Countdown Timer
-A visual timer widget that starts when tapped/clicked, and has an audible alert when done. To use it create a `note` field with an `appearance` set to `countdown-timer`. The duration of the timer is the field's value, which can be set in the XLSForm's `default` column. If this value is not set, the timer will be set to 60 seconds.
+
+A visual timer widget that starts when tapped/clicked, and has an audible alert when done. To use it, first make sure you have the [`namespaces` column](https://getodk.github.io/xforms-spec/#namespaces) in the "settings" tab of your XLSForm populated with a value that includes `cht=https://communityhealthtoolkit.org`. Then, you can add the timer as a `trigger` field with the _appearance_ set to `countdown-timer`. The duration of the timer can be set in a column named _instance::cht:duration_ (the default value is 60 seconds).
```
latin-panda, line 96: You can add a sample to make this easier for readers, something that looks like the xlsx 

```
| type | name | label | required | appearance | instance::cht:duration | ... |
| -- | -- | -- | -- | -- |
| X | Y | Z....
```

latin-panda: I asked this in the other PR
why do we need to hardcode this: `cht=https://communityhealthtoolkit.org`  Are there any other possible values?

latin-panda (changes requested):
Thanks Josh, I added a question and a suggestion for your consideration

content/en/apps/reference/forms/app.md
jkuester: So, the namespaces stuff in the XLSForm is from [the spec](https://xlsform.org/en/#advanced-use-and-extensibility) which notes:

> The settings sheet has support for defining (multiple space-separated) additional custom namespaces and namespace prefixes using the namespaces column. You'll then be able to use those namespaces in the survey sheet, for example, to properly define a custom attribute with your organisation's own namespace.

The idea is that we are going to use `cht` as our namespace (and eventually migrate all our custom form features to have the `cht` prefix). We already started with pattern with `cht:extension-lib`.  The difference with the custom `instance` column is that the column name is incorperated in the xform xml structure and so Pyxform wants the namespace to be a real [xml namespace](https://www.w3.org/TR/REC-xml-names/#dt-NSName). 

Regarding the _value_ of the namespace (e.g. https://communityhealthtoolkit.org/), as far as I can tell from some general searching of the internet, there is no particular meaning to the value except that it should be unique to the set of functionality that it is associated with (no one else is going to use the same namespace). The most common values I saw as examples for namespace all were URLs. As far as I can tell, Pyxform does not really care what value you set for the `cht` namespace as long as you give something.  The https://communityhealthtoolkit.org url just felt like a nice and relevant value to recommend.  Open to other ideas, though!

PR #1313, opened 2024-03-19
fix(na): Remove macOS specific `buildx` instructios for Core Dev Environment

mrjones-plip:
# Remove macOS specific `buildx` instructios for Core Dev Environment

Once we [release](https://github.com/medic/cht-core/issues/8823) `4.6.0`, we'll be on to `4.7.0` and `master` will pick up the new [multi-platform builds](https://github.com/medic/cht-core/issues/8841).   At that time, we should be able to remove the `buildx` macOS specific section of the CHT Core Dev setup docs.

# License

### DIFF
--- a/content/en/contribute/code/core/dev-environment.md
+++ b/content/en/contribute/code/core/dev-environment.md
@@ -123,8 +123,7 @@ curl -X GET "http://medic:password@localhost:5984/_membership" | jq
 
 Create a `docker-compose.yml` and `couchdb-override.yml` files under the `~/cht-docker` folder with this code:
 
-{{< tabpane persistLang=false lang=shell >}}
-{{< tab header="Linux (Ubuntu) & Windows (WSL2)" >}}
+```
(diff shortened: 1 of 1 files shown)


andrablaj (approved):
🎆

mrjones-plip:
I've confirmed that current `master` resolves to multi-platform images even before `4.6.0` is released, we're good to go!

Note two last lines:

```
docker run --rm mplatform/mquery public.ecr.aws/medic/cht-couchdb:4.6.0-alpha
Image: public.ecr.aws/medic/cht-couchdb:4.6.0-alpha (digest: sha256:190f54d421e5938e6d2f8c28aa6db30b14be1cf107bc4b60a52996022188e0b2)
 * Manifest List: Yes (Image type: application/vnd.docker.distribution.manifest.list.v2+json)
 * Supported platforms:
   - linux/amd64
   - linux/arm64
```

PR #1309, opened 2024-03-14
feat: add cht-core 4.6.0 release notes

jkuester:
# Description

https://github.com/medic/cht-core/issues/8823

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #1316, opened 2024-03-20
fix: include dates for cht-core 4.6.0 release

jkuester:
# Description

When adding the [4.6.0 release notes](https://github.com/medic/cht-docs/pull/1309) I forgot to update the dates for the release. :sweat: 

# License


mrjones-plip (approved):
not sure on the diff between `TBA` and `TBD`, but the math checks out - :3rd_place_medal:  

(that was the random emoji I came up with - no judgement on the quality of this PR :upside_down_face: )

PR #1330, opened 2024-03-26
feat(na): Bump hugo to 0.122 & add Docker Compose file, update readme

mrjones-plip:
# Description

This PR: 
* Adds ability to run `hugo server` with Docker compose in addition to native packages via a newly add `compose.yml` file
* updates readme with steps to use newly added compose file
* bumps `hugo` in `.tool-version`  to `0.122`

# License

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-hugo extended_0.101.0
+hugo extended_0.122.0
 golang 1.21.0
 nodejs 16.20.2
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The documentation is built using Markdown pages, which can be converted into a n
 
 ## Build dependencies
 
-The documentation site uses [Hugo](https://gohugo.io/) including the extended version of Hugo. Version v0.122 is used per (see #https://github.com/medic/cht-docs/issues/1302) as denoted in `.tool-versions` [file](https://github.com/medic/cht-docs/blob/main/.tool-versions).
+The documentation site uses [Hugo](https://gohugo.io/) including the extended version of Hugo (see `.tool-versions` [file](https://github.com/medic/cht-docs/blob/main/.tool-versions) for specific version).
 
 To proceed, please choose one of the three options below of Docker, native packages or using `asdf` then proceed to ["Run Hugo" below](#run-hugo) to start your local instance.
 
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ The documentation site uses [Hugo](https://gohugo.io/) including the extended ve
 
 To proceed, please choose one of the three options below of Docker, native packages or using `asdf` then proceed to ["Run Hugo" below](#run-hugo) to start your local instance.
 
+For folks using `asdf` and native packages, be sure the correct `hugo` version is installed per `.tool-versions` [file](https://github.com/medic/cht-docs/blob/main/.tool-versions).
+
 ### Docker
 
 Running `hugo` locally using Docker is likely the easiest way to update the docs.  You only need to ensure you have [Docker and Docker compose installed](https://docs.docker.com/compose/install/).
--- a/README.md
+++ b/README.md
@@ -10,20 +10,38 @@ The documentation is built using Markdown pages, which can be converted into a n
 
 ## Build dependencies
 
-The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.76.0 and later. See `.tool-versions` for the currently used version
+The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.76.0 and later. See `.tool-versions` for the currently used version.
 
-[asdf](https://asdf-vm.com/guide/getting-started.html) is the recommended way to manage `hugo` and `golang` versions for local development.
+### Docker
 
-After installing it run:
+Running `hugo` locally using Docker is likely the easiest way to update the docs.  Ensure you have Docker and Docker compose installed, then:
+
+1. Clone the `cht-docs` repo
+1. From the command line, navigate to the local directory where you cloned the repo
+1. Type `docker compose up`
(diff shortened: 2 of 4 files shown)


.tool-versions
```diff
@@ -1,3 +1,3 @@
-hugo extended_0.101.0
+hugo extended_0.122.0
```
jkuester, line 1: So, I have been trying to decide if it is a problem that this value does not match the [min Hugo Version](https://github.com/medic/cht-docs/blob/b8e560d9f9b1f58efbe3e61b6bb512af562b3a97/config.toml#L74) in the config.toml.  Ultimatly, I think this could create problems in the future since the Hugo version [used for CI testing](https://github.com/medic/cht-docs/blob/b8e560d9f9b1f58efbe3e61b6bb512af562b3a97/.github/workflows/ci.yml#L26) is based on this `.tool-versions` value.

Currently the min version is set to `0.93.0`.  But if I am running `0.101.0` locally and use a feature that requires `0.101.0`, then the CI will not catch it and anyone actually using `0.93.0` will be broken....

Perhaps we should just remove the min version from the `config.toml` and just say that the value here is the source of truth for the supported version of Hugo...

README.md
```diff
@@ -10,26 +10,39 @@ The documentation is built using Markdown pages, which can be converted into a n
 
 ## Build dependencies
 
-The documentation site uses [Hugo](https://gohugo.io/), and specifically uses features found in the extended version of Hugo, v0.76.0 and later. See `.tool-versions` for the currently used version
+The documentation site uses [Hugo](https://gohugo.io/) including the extended version of Hugo. Version v0.122 is used per (see #https://github.com/medic/cht-docs/issues/1302) as denoted in `.tool-versions` [file](https://github.com/medic/cht-docs/blob/main/.tool-versions).
```
jkuester, line 13:
```suggestion
The documentation site uses the extended version of [Hugo](https://gohugo.io/). See the `.tool-versions` [file](https://github.com/medic/cht-docs/blob/main/.tool-versions) for the latest version of Hugo supported for these docs.
```

Hard-coding the actual version value here gives us another place to have to update when the supported version changes.  Seems better to just point folks to the `.tool-versions` file as the source of truth.

README.md
```diff
@@ -10,26 +10,39 @@ The documentation is built using Markdown pages, which can be converted into a n
...
+To proceed, please choose one of the three options below of Docker, native packages or using `asdf` then proceed to ["Run Hugo" below](#run-hugo) to start your local instance.
 
-After installing it run:
+### Docker
+
+Running `hugo` locally using Docker is likely the easiest way to update the docs.  You only need to ensure you have [Docker and Docker compose installed](https://docs.docker.com/compose/install/).
+
+### Native packages
+
+If you'd like to install packages directly on your workstation, follow the [installation instructions for your Operating System](https://gohugo.io/getting-started/installing/), and be sure to get the extended version. Most users will be able to simply install using their native package manager like `brew`, `apt` or `snap`.
```
jkuester, line 23: Probably should also warn folks here that they need to be aware of the Hugo version they are installing (and if it matches the `.tool-versions`.

README.md
```diff
@@ -10,26 +10,39 @@ The documentation is built using Markdown pages, which can be converted into a n
...
 asdf plugin add golang
 asdf plugin add hugo
 ```
-### Installing Hugo Manually
 
-If using `asdf` fails you can try installing Hugo manually. To install, follow the [installation instructions for your Operating System](https://gohugo.io/getting-started/installing/), and be sure to get the extended version. Most users will be able to simply install using their native package manager like `brew`, `apt` or `snap`.
+### Run Hugo
 
-## Building the Documentation
+For both native packages or `asdf`, you can start your local CHT Docs instance by:
```
jkuester, line 37:
```suggestion
You can start your local CHT Docs instance by:
```

README.md
```diff
@@ -10,26 +10,39 @@ The documentation is built using Markdown pages, which can be converted into a n
...
 
 1. Clone the `cht-docs` repo
-1. From the command line, navigate to the local directory where you cloned the repo
-1. Type `hugo server`
-1. Preview your site in your browser at: http://localhost:1313/
-1. As you make changes to the site, your browser will automatically reload your local dev site so you can easily and quickly see your changes.
+2. From the command line, navigate to the local directory where you cloned the repo
+3. To start `hugo` run:
+   * `hugo server` for `asdf` and native packages **OR**
+   * `docker compose up` for Docker compose
```
jkuester, line 43: So, one downside to running via Docker Compose is that it will end up leaving a stopped container sitting on your host _forever_ (unless you manually remove it).  It is not a huge deal, but it seems like, out of politeness, we should provide instructions for how to delete the container.

Alternatively, we could suggest running the container via the `docker compose run --rm -it -p 1313:1313 hugo` command and then I think the container will be automatically removed when the command is stopped.

jkuester (approved):
Most excellent to have a Docker deployment documented!  I left some minor comments/questions inline.

Fun fact: I have actually been using [Distrobox](https://distrobox.it/) to run Hugo from a Docker container on my machine for several months. Same basic concept as our approach here, but Distrobox abstracts out the docker part of things for me...

README.md
mrjones-plip: yeah - love the ephemeral approach!  That said, testing it, I get ~20 second start up ever time for the `--rm` approach and only that on the first start for the vanila compose.  It's otherwise ~3 seconds on subsequent starts.  

I propose we keep it as I have, add an explicit name in the compose file and call out how to delete it.  Done in 21e999fc

.tool-versions
mrjones-plip: yo!! I knew I picked you for good reason to review my PR.  totally didn't dive this deep - thank you!

fixed in 64fa0eed

README.md
mrjones-plip: natch 0a832ef1

README.md
mrjones-plip: yup yup!  Fixed in 685431f6

Issue #977, opened 2023-03-07
Add copy to clipboard button for code snippets


eljhkrr:
While going through the [v4 setup guide](https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/), I had to manually copy text blocks containing shell commands:
![image](https://user-images.githubusercontent.com/4529018/223527055-89082aba-5d7b-4e4f-b889-5d369937b488.png)
It would be very useful if users could have a copy to clipboard button to grab the whole set of commands at a go like in the example below:
![image](https://user-images.githubusercontent.com/4529018/223527373-c3e592dc-4fdb-410b-b1d1-7e86fc0f3cb2.png)
There are a number of workarounds both in the hugo ecosystem and using vanilla JS that can be adopted to add this functionality. A few examples:
https://digitaldrummerj.me/hugo-add-copy-code-snippet-button
https://www.dannyguo.com/blog/how-to-add-copy-to-clipboard-buttons-to-code-blocks-in-hugo/

eljhkrr:
Copy code snippet is available in docsy v0.6.0+ 
https://github.com/google/docsy/pull/1447
https://www.docsy.dev/docs/adding-content/lookandfeel/#code-highlighting-with-chroma

PR #1355, opened 2024-04-18
Docsy theme update from '0.6.0' to '0.7.0'

esthermmoturi:
# Description

This update from version **'0.6.0'** to **'0.7.0'** of the Docsy theme will enable the copy to clipboard feature for blocks of code on the docsite.

This solves **#977**


eljhkrr:
I'd recommend upgrading to the latest docsy release [v0.9.1](https://github.com/google/docsy/releases/tag/v0.9.1)

esthermmoturi:
@eljhkrr , we are doing slow upgrades because 0.9.1 had a lot of new dependencies that were a mismatch with the hugo version we use. I plan to upgrade hugo then upgrade docsy to the latest version.

PR #1356, opened 2024-04-18
Upgrading docsy from 0.6.0 to 0.7.0

esthermmoturi:
# Description

This update from version '0.6.0' to '0.7.0' of the Docsy theme will enable the copy to clipboard feature for blocks of code on the docsite.

This solves https://github.com/medic/cht-docs/issues/977.

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -80,7 +80,7 @@ h2, h3 {
   }
   .td-sidebar-link {
     &__page {
-      //color: $gray-900;
+      color: #222222;
       font-weight: $font-weight-normal;
     }
     &__page.active {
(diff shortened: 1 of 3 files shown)


assets/scss/_variables_project.scss
```diff
@@ -80,7 +80,7 @@ h2, h3 {
   }
   .td-sidebar-link {
     &__page {
-      color: $gray-900;
+      //color: $gray-900;
```
jkuester, line 83: So, I guess this affects the color of the _unselected_ page links in the sidebar. It looks like the old `gray-900` variable was resolving to the [`#222222` color](https://www.color-hex.com/color/222222).  By removing this here, we are dropping back to using [`#495057`](https://www.color-hex.com/color/495057) which is just a lighter shade of gray.

image

Personally, the lighter shade is fine with me, but if we wanted to stick with what we had before, we could just hard-code the color value (either way we should remove the commented-out code):

```suggestion
      color: #222222;
```

jkuester (approved):
Most excellent!  I had one minor comment about the css change, but otherwise this is ready to go!  The copy buttons work great!

PR #1363, opened 2024-04-25
Chore: reword to avoid ambiguity

Benmuiruri:
# Description

I found this line in the docs abit confusing [Also, do not use quotes "" to enclose your password, and do not use spaces in your password](https://docs.communityhealthtoolkit.org/contribute/code/core/deploy-on-eks/#:~:text=Also%2C%20do%20not%20use%20quotes%20%22%22%20to%20enclose%20your%20password%2C%20and%20do%20not%20use%20spaces%20in%20your%20password). it seems to imply one should not enter the password as a string in the values file.

While you do not explicitly need to wrap `strings` in quotes in `yaml` as explained in [SO post](https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml) and random [blog post](https://brettops.io/blog/yaml-strings/), there are some edge cases. 

Therefore, to avoid any possible issues with setting the bugs lets have the docs mention it's best to use quotes. 

# License

### DIFF
--- a/content/en/contribute/code/core/deploy-on-eks.md
+++ b/content/en/contribute/code/core/deploy-on-eks.md
@@ -88,7 +88,7 @@ After you have created a ticket per "Request permission" above, you should get a
    * Put a UUID in `secret` - the command `uuidgen` is great for this
    * Update `host` to be your `username`.  For example: `mrjones.dev.medicmobile.org`
 4. Use `uuidgen` to fill in the `secret` in `values.yaml`
-5. Use a good passphrase (diceware!) to fill in `password` in `values.yaml`. _Please note that a few special characters are unsupported in this field like `:`, `@`, `"`, `'`, etc. Also, do not use quotes `""` to enclose your password, and do not use spaces in your password. This will not impact the deployment but will not let you log in to the CHT instance._
+5. Use a good passphrase (diceware!) to fill in `password` in `values.yaml`. _Please note that a few special characters are unsupported in this field like `:`, `@`, `"`, `'`, etc. Add your password as a string by enclosing it in quotes `""`, and do not use spaces in your password. This will not impact the deployment but will not let you log in to the CHT instance._
 6. Ensure you have the latest code of `cht-core` [repo](https://github.com/medic/cht-core):
    ```shell
    git checkout master;git pull origin


content/en/contribute/code/core/deploy-on-eks.md
```diff
@@ -88,7 +88,7 @@ After you have created a ticket per "Request permission" above, you should get a
    * Put a UUID in `secret` - the command `uuidgen` is great for this
    * Update `host` to be your `username`.  For example: `mrjones.dev.medicmobile.org`
 4. Use `uuidgen` to fill in the `secret` in `values.yaml`
-5. Use a good passphrase (diceware!) to fill in `password` in `values.yaml`. _Please note that a few special characters are unsupported in this field like `:`, `@`, `"`, `'`, etc. Also, do not use quotes `""` to enclose your password, and do not use spaces in your password. This will not impact the deployment but will not let you log in to the CHT instance._
+5. Use a good passphrase (diceware!) to fill in `password` in `values.yaml`. _Please note that a few special characters are unsupported in this field like `:`, `@`, `"`, `'`, etc. Add your password as a string by enclosing it in quotes `""`, and do not use spaces in your password. This will not impact the deployment but will not let you log in to the CHT instance._
```
sugat009, line 91: question: I'm confused here, the password is supposed to be not enclosed within the quotes(at least at the writing of this document and when I tried. Ref: https://github.com/medic/medic-infrastructure/issues/863#issuecomment-2033514003). Was it changed?

Benmuiruri: Hi @sugat009, a couple of things. 
1. The password entry is a string. You either enter it explicitly as a string by wrapping it in quotes or your entry is converted to a string in the end if you did not use quotes. 
2. As the two links, [StackOverflow](https://stackoverflow.com/questions/19109912/yaml-do-i-need-quotes-for-strings-in-yaml) and the [blog post](https://brettops.io/blog/yaml-strings/) explain there are at least six ways to represent a string in YAML. 

- If you have this `password: my-password` the password will be cast into a `string` as `"my-password"`
- If you have `password: "my-password"` the password is still a `string` as `"my-password"`

Here is the problem when you do not explicit string. 
- If you have `password: !-password` this will throw an error. 
- if you have `password: "!-password"` this will correctly be converted to a string as `"!-password"`

That is why I am suggesting we explicitly require a person to add the password as a string.

sugat009: My concern is whether or not it works. When I tried this and set my password in the manner `"pass-word"`, it did not work. If it has started working then, this change does make sense.

Benmuiruri: Reading through https://github.com/medic/medic-infrastructure/issues/863#issuecomment-2033514003) it seems the problem was host more than the password. 

I deployed successfully yesterday and my password was enclosed in a string.

sugat009:
Update:
I tried deploying to EKS with my old config yaml file i.e. the file with `password` field with a non-enclosed string and I'm getting the following error.

image

However, it was deployed successfully after enclosing the `password` field with double quotes. 
Approving this PR as the premise has changed.

sugat009 (approved):

Benmuiruri:
Thank you for confirming it works with explicit strings.

nydr:
With that change all commonly used special characters should be supported

PR #1367, opened 2024-04-25
feat(na): update couch2pg watchdog setup 

mrjones-plip:
# update couch2pg watchdog setup 

see medic/cht-watchdog/issues/81

# License


mrjones-plip:
Sorry @lorerod ! I meant to request a review from you not to assign it to you :sweat_smile:

content/en/apps/guides/hosting/monitoring/setup.md
```diff
@@ -119,34 +119,23 @@ docker compose up -d
...
-   
+1. Copy the example config file, so you can add the correct contents in them:
    ```shell
    cd ~/cht-watchdog
-   cp exporters/postgres/postgres-instances.example.yml exporters/postgres/postgres-instances.yml
-   cp exporters/postgres/postgres_exporter.example.yml exporters/postgres/postgres_exporter.yml
+   cp exporters/postgres/sql_servers_example.yml exporters/postgres/sql_servers.yml
    ```
-2. Edit `postgres-instances.yml` you just created and add your target postgres connection URL along with the proper root URL for your CHT instance as the label value. For example, if your postgres server was `db.example.com` and your CHT instance was `cht.example.com` the config would be:
+2. Edit `sql_servers.yml` you just created and add your target postgres connection URL along with the proper root URL for your CHT instance. For example, if your postgres server was `db.example.com`, your CHT instance was `cht.example.com`, your user was `db_user` and your password was `db_password`,  the config would be:
```
lorerod, line 127: Is there any specification for that Postgres database we need to define here? 
I'm trying this, and I need a Postgres database.

lorerod (approved):
Thank you, @mrjones-plip, for this update. I just left a question.

content/en/apps/guides/hosting/monitoring/setup.md
mrjones-plip: This is explicitly meant to work with Postgres that couch2pg is fed into. From [the intro](https://docs.communityhealthtoolkit.org/apps/guides/hosting/monitoring/setup/#couch2pg-data) above this step:

> Watchdog now supports easily ingesting [couch2pg](http://localhost:1313/apps/tutorials/couch2pg-setup/) data read in from a Postgres database (supports Postgres >= 9.x).

If you need a database to test with you can use the `cht-couch2pg` docker compose file to start both a couch2pg instance and a Postgres server.  The [default values](https://github.com/medic/cht-watchdog/blob/ee24210c1a82f7e5f1c9c5a4116eeebd25792f24/exporters/postgres/sql_servers_example.yml#L28) in the `- targets:` should work, but you'll need to update the port for your cht instance. Note that the `172.17.0.1` IP is the universal IP for your localhost where your running docker.

See [this comment](https://github.com/medic/cht-watchdog/issues/81#issue-1824781379) for a step by step instructions to do this. Also happy to pair if wanna debug real time!

lorerod: Is the universal IP the same if I'm using docker desktop?

mrjones-plip: oh...maybe :sweat_smile: 

lemme check on my test mac here...

mrjones-plip: Thanks for the call just now to debug this - the issue was, as you suspected, that IP is not universal and did not work on macOS running Docker Desktop.  The trick was to get the IP for you computer on the WiFi and then use that every where the `172.17.0.1` IP was used above.

thanks for checking that this all works!

lorerod: Thanks for your help @mrjones-plip. Feel free to merge this anytime.

mrjones-plip: yay! thanks @lor

Issue #1369, opened 2024-04-29
Advice to have the OS language set in English before running make org=new_brand keygen


lorerod:
When running the command `make org=new_brand keygen`, it loops through the questions repeatedly when having the OS language set in English.

PR #1370, opened 2024-04-29
Advice to change OS language to English

lorerod:
# Description

closes https://github.com/medic/cht-docs/issues/1369

# License


latin-panda (approved):
Thanks @lorerod! I wouldn't think on this first when experiencing this type of errors 😅

PR #1386, opened 2024-05-10
Add some extra steps and clarification for easier understanding of release process for cht-android

sugat009:
# Description

Add some necessary steps and clarification I faced when releasing `cht-android`.

# License

### DIFF
--- a/content/en/contribute/code/android/releasing.md
+++ b/content/en/contribute/code/android/releasing.md
@@ -23,7 +23,7 @@ All Medic's android projects automatically build, sign, and release builds via G
 
 1. Make sure all issues for this release have passed AT and been merged into `master`. You can also create an alpha release from a feature branch, to provide the needed `.apk` files to the QA team.
 2. Create a git tag starting with `v` and ending with the alpha version, e.g. `git tag v1.2.3-alpha.1` and push the tag to GitHub (`git push --tags`). For features branches, you can add the name of the branch or whatever keyword helps to identify the the release, e.g. `v1.2.3-alpha.dark-theme.4`.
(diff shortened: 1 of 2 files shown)


content/en/contribute/code/android/releasing.md
```diff
@@ -23,14 +23,15 @@ All Medic's android projects automatically build, sign, and release builds via G
...
-3. Creating the tag will trigger the building and signing of the app in CI. The release-ready APKs are available for side-loading from the Releases section in the project (e.g. [CHT-Android Releases](https://github.com/medic/cht-android/releases)), along with the AABs that may be required by the Google Play Store.
+3. Creating the tag will trigger the building and signing of the app in CI. The release-ready APKs are available for side-loading from the Releases section in the project (e.g. [CHT-Android Releases](https://github.com/medic/cht-android/releases)), along with the AABs that may be required by the Google Play Store. Although a new release with the same name as the tag is automatically generated when you create the tag, the associated assets are not immediately visible under that tag. To make the assets visible, you need to explicitly link the release to the tag. Once you've linked the release to the tag, the assets will be listed under that tag.
```
jkuester, line 26: @garethbowen do you know of any reason why we need to keep _draft GitHub releases_ around for these alpha builds once we are done testing?  The cht-android CI will create a draft release for every tag that gets pushed. These drafts are what the release _artifacts_ (apks) are attached to (so we cannot just stop creating the drafts).  However, once we are done testing and we go to cut the production release, it seems like we could add a step to go back and delete the _draft_ release since the [releases page](https://github.com/medic/cht-android/releases) is getting cluttered....

@sugat009 perhaps instead of adding the extra step of "explicitly link the release to the tag" here we could just specify something like the following to be sure no one gets confused:

> Note that the draft release on GitHub with the generated artifacts will not be explicitly linked to the GitHub page for the tag because the release is not published.

jkuester (approved):
Excellent updates!  Thank you!

PR #1388, opened 2024-05-13
fix(na): update sql exporter docs to use new format

mrjones-plip:
# Description

update sql exporter docs to use new format per medic/cht-watchdog/issues/107

# License

### DIFF
--- a/content/en/hosting/monitoring/setup.md
+++ b/content/en/hosting/monitoring/setup.md
@@ -129,9 +129,9 @@ With the [release of 1.1.0](https://github.com/medic/cht-watchdog/releases/tag/1
 2. Edit `sql_servers.yml` you just created and add your target postgres connection URL. For example, if your postgres server was `db.example.com`, your user was `db_user` and your password was `db_password`,  the config would be:
    ```yaml
    - targets:
-      "db1": 'postgres://db_user:db_password@db.example.com:5432/cht?sslmode=disable' # //NOSONAR - password is safe to commit
+      "db-example-com": 'postgres://db_user:db_password@db.example.com:5432/cht?sslmode=disable' # //NOSONAR - password is safe to commit
    ```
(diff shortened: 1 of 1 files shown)


content/en/hosting/monitoring/setup.md
```diff
@@ -126,13 +126,12 @@ With the [release of 1.1.0](https://github.com/medic/cht-watchdog/releases/tag/1
...
-2. Edit `sql_servers.yml` you just created and add your target postgres connection URL along with the proper root URL for your CHT instance. For example, if your postgres server was `db.example.com`, your CHT instance was `cht.example.com`, your user was `db_user` and your password was `db_password`,  the config would be:
+2. Edit `sql_servers.yml` you just created and add your target postgres connection URL. For example, if your postgres server was `db.example.com`, your user was `db_user` and your password was `db_password`,  the config would be:
    ```yaml
    - targets:
-      "cht.example.com": 'postgres://db_user:db_password@db.example.com:5432/cht?sslmode=disable' # //NOSONAR - password is safe to commit
-
+      "db1": 'postgres://db_user:db_password@db.example.com:5432/cht?sslmode=disable' # //NOSONAR - password is safe to commit
    ```
-   You may add as many targets as you would like here - one for each CHT Core instance in your `cht-instances.yml` file.
+   You may add as many targets as you would like here - one for each CHT Core instance in your `cht-instances.yml` file. Be sure to give each entry a unique name (eg `db1` as shown, then `db2`, `db3` etc.).
```
jkuester, line 134: Depending on the resolution of https://github.com/medic/cht-watchdog/pull/108/files#r1604005494, we might want to tweak this example.

jkuester (approved):
:+1:

content/en/hosting/monitoring/setup.md
mrjones-plip: yups!  fixed in a20d050f

PR #1396, opened 2024-05-20
chore: Remove production setup from CHT Sync

andrablaj:
# Description

Per [the discussion on Slack](https://medic.slack.com/archives/C03SP4KRVEX/p1715811365800719), remove the CHT Sync production setup page until the team validates that the approach is appropriate for production use. 

# License


andrablaj:
> we have a solid draft here of where to start

Exactly! Plus, we will be EKS pros :bowtie: and we will have a whole new approach to deployment!

PR #1411, opened 2024-05-31
Fix: correcting typos

deining:
This PR corrects various typos I found in the project.


hugo.toml
```diff
@@ -202,7 +202,7 @@ enable = false
 RelPermalink= "communityhealthtoolkit.org"
 
 [related]
-  treshold = 1
+  threshold = 1
```
garethbowen, line 205: I'm curious about this change. The file name changed and the threshold value is now set which might have unintended consequences. What was the logic behind changing this file name?

garethbowen: @deining Do you know?

deining: > I'm curious about this change. The file name changed (...)
> What was the logic behind changing this file name?

Have a look [here](https://gohugo.io/getting-started/configuration/#configuration-file):

```
With v0.109.0 and earlier the basename of the site configuration file was config
instead of hugo. You can use either, but should transition to the new naming
convention when practical.
```

Hope this make the motivation for the name change clear.

garethbowen: Ok, that's great to know!

I'm still cautious about the typo correction, because it'll go from not applying to applying the threshold. It might be safer to delete it instead of fixing the typo...

deining: > Ok, that's great to know!
> 
> I'm still cautious about the typo correction, because it'll go from not applying to applying the threshold. It might be safer to delete it instead of fixing the typo...

Agreed. I removed the line  treshold = 1`, I think this the way to go here.

garethbowen (commented):
This looks good to me. 

@esthermmoturi Handing this over to you while I'm on leave.

hugo.toml
esthermmoturi: Hey @deining , I have undone the name change and the `treshold` change. I will review the aforementioned changes in [this](https://github.com/medic/cht-docs/issues/1420) issue. Thank you for the suggestion.

esthermmoturi (approved):
@deining Thank you for taking your time to contribute to our documentation site. I am curious to know if you used software/script to find and fix the typos or you did it manually. Other than that, everything looks good. I have approved the changes.

deining:
> @deining Thank you for taking your time to contribute to our documentation site.

You are welcome.

> I am curious to know if you used software/script to find and fix the typos or you did it manually.

Yes, I did use spell checkers to detect the typos.

> Other than that, everything looks good. I have approved the changes.

Great. Looking forward seeing my PR merged.

PR #1434, opened 2024-06-20
chore: Add reference links to cht-core files

tatilepizs:
# Description

Add reference links to cht-core files

# License

### DIFF
--- a/content/en/contribute/code/core/automated-tests.md
+++ b/content/en/contribute/code/core/automated-tests.md
@@ -41,7 +41,7 @@ High coverage of functionality. If measured in branch coverage percentage,  aim 
 | Extremely fast | Extremely low | Extremely stable |
 
 ### Implementation
-In cht-core unit tests are located in the `tests` directories of each app  (e.g. in `webapp/tests` you can find unit test for the webapp). Run them locally with: `npm run unit`.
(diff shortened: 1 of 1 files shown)


content/en/contribute/code/core/automated-tests.md
```diff
@@ -41,7 +41,7 @@ High coverage of functionality. If measured in branch coverage percentage,  aim
 | Extremely fast | Extremely low | Extremely stable |
 
 ### Implementation
-In cht-core unit tests are located in the `tests` directories of each app  (e.g. in `webapp/tests` you can find unit test for the webapp). Run them locally with: `npm run unit`.
+In cht-core unit tests are located in the `tests` directories of each app  (e.g. in [`webapp/tests`](https://github.com/medic/cht-core/tree/master/webapp/tests) you can find unit test for the webapp). Run them locally with: [`npm run unit`](https://github.com/medic/cht-core/blob/master/package.json#L42).
```
andrablaj, line 44: I wonder how often we update the `package.json` file, and the line number changes, and then this doc links to the wrong lines. Do you think we can specify at the beginning that the commands can be found in the `package.json` file for those who want to know more?

andrablaj (commented):
As we modify the `package.json` file [quite often](https://github.com/medic/cht-core/commits/master/package.json), can we review the way we link specific lines in this doc, as those might change regularly and we might forget to update this doc?

content/en/contribute/code/core/automated-tests.md
tatilepizs: That is a really good point, I will add it, thanks Andra!

Issue #1373, opened 2024-05-02
Fix grep results from the moving hosting section PR


esthermmoturi:
This is in regards to the comment left by Mrjones in [#1361](https://github.com/medic/cht-docs/pull/1361)

PR #1361, opened 2024-04-23
Move available hosting resources and fix links

esthermmoturi:
# Description

Moving the hosting section from a 3rd level navigation section to a top-level navigation section. 

per #1351

PR #1436, opened 2024-06-24
Fix grep results for aliases

esthermmoturi:
To fix concern raised on grep results when doing hosting moving, for aliases. - https://github.com/medic/cht-docs/pull/1361

This fixes this issue raised - https://github.com/medic/cht-docs/issues/1373

### DIFF
--- a/content/en/contribute/code/core/dev-environment.md
+++ b/content/en/contribute/code/core/dev-environment.md
@@ -6,6 +6,7 @@ description: >
   Getting your local machine ready to do development work on CHT Core.
 aliases: >
   apps/guides/hosting/core-developer
+  apps/guides/hosting/hosting/4.x/app-developer
 ---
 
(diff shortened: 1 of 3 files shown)


content/en/contribute/code/core/dev-environment.md
```diff
@@ -6,6 +6,7 @@ description: >
   Getting your local machine ready to do development work on CHT Core.
 aliases: >
   apps/guides/hosting/core-developer
+  apps/guides/hosting/hosting/4.x/app-developer
```
mrjones-plip, line 9: These both need a leading `/` to work. Note that  I had a cached redirect in hugo that I had to clear with `hugo serve --cleanDestinationDir` and then testing these with `curl -v localhost:1313/apps/guides/hosting/hosting/4.x/app-developer/  -o /dev/null` would return a `404`.

```suggestion
  /apps/guides/hosting/core-developer
  /apps/guides/hosting/hosting/4.x/app-developer
```

mrjones-plip (approved):
yay - nice work!

There's one minor fix needed - go ahead and make the change per the suggestion and then you can safely merge to `main`.  Approving to make this zoooom!

esthermmoturi:
Thank you @mrjones-plip

PR #1452, opened 2024-07-12
chore: Update Infrastructure FWG roles

jonathanvq:
# Description

Housekeeping, removes roles that are no longer active in the Infrastructure FWG
# License


garethbowen (approved):
Approved because this is a step in the right direction. I would go further and remove Infrastructure from this section (or page?) altogether as they are no longer operating as a FWG.

jonathanvq:
@garethbowen , you're right about removing Infra FWG instead of just the two roles that are no longer with the team. Although I was hesitating a bit, it is fair to say we are not a full FWG like the others, and for the time being, the documentation should reflect this. I have pushed another small commit for this. Thanks

garethbowen (approved):
Thanks!

PR #1460, opened 2024-07-17
feat: add doc for disabling SONAR rule preventing TODO comments

jkuester:
# Description

[RSPEC-1135](https://rules.sonarsource.com/javascript/RSPEC-1135/) prevents adding any `TODO` comments to the code base. The idea is to guard against merging unfinished code, but in reality, this is not much of a danger since those situations are easily identified via code review.  

Instead, this rule is preventing use from using `TODO` comments to mark places in the code that will need updated in the future (but cannot be updated as a part of the current project).  I think the benefits are not worth the downsides and we should just disable this rule.

# License

### DIFF
--- a/content/en/contribute/code/static-analysis.md
+++ b/content/en/contribute/code/static-analysis.md
@@ -176,6 +176,9 @@ Java:
       - `threshold` 7 -> 4
     - [`S3776`](https://rules.sonarsource.com/javascript/RSPEC-3776/) - Cognitive Complexity of functions should not be too high
       - `threshold` 15 -> 5
+  - Disabled
+      - [`S1135`](https://rules.sonarsource.com/javascript/RSPEC-1135/) - Track uses of "TODO" tags
+          - Disabled because we typically use "TODO" tags to mark code as needing future updates which cannot be currently applied for reasons outside our direct control (e.g. requiring a dependency uplift).
 
 JavaScript:
 
@@ -186,6 +189,8 @@ JavaScript:
       - [`S3776`](https://rules.sonarsource.com/javascript/RSPEC-3776/) - Cognitive Complexity of functions should not be too high
          - `threshold` 15 -> 5
    - Disabled
+       - [`S1135`](https://rules.sonarsource.com/javascript/RSPEC-1135/) - Track uses of "TODO" tags
+           - Disabled because we typically use "TODO" tags to mark code as needing future updates which cannot be currently applied for reasons outside our direct control (e.g. requiring a dependency uplift).
       - [`S2699`](https://rules.sonarsource.com/javascript/RSPEC-2699/) - Tests should include assertions
          - Disabled due of rigidity of the rule when detecting `expect` imports and calls to imported functions that have assertions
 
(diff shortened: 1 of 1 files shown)


garethbowen:
This is my personal point of view and I'll go along with the majority if I'm outvoted, but personally I hate TODO comments. IMO they should only be used for things you want to fix before merging, which means having sonar pick this up during PR review is perfect. If you want to remember to come back and fix something after merging then it should be captured in a GH issue for visibility, tracking, discussion, including in release notes, etc etc. Putting it inline makes it hard to find, and in reality nobody ever goes back to fix it - if it wasn't important enough to fix in the original PR it's not going to be important enough to go back and fix next week.

m5r:
I rarely use TODO comments but I agree that a GitHub issue is the better place to track future changes. With that said, they come in handy when the code can move around and it becomes harder to find the place we were supposed to change.
The most recent example is https://github.com/medic/cht-conf/pull/623 where I added 2 TODO comments to pinpoint the place in the code where an eslint comment should be removed. In this case it's short-lived since I'm already working on the fix in a separate branch but the point is that it feels a bit too heavy to track this kind of change in an issue.

If we're going to keep this rule, I will need a solution to make Sonar happy in that PR because neither adding `// NOSONAR` at the end of the line or wrapping it with two lines `// NOSONAR_BEGIN` & `// NOSONAR_END` will make Sonar happy https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=623&id=medic_cht-conf&open=AZDBIkL15aSDoadCSmFw

garethbowen:
@m5r I find it useful to use permalink when linking to code so you can find the code as it was at the time you wrote the issue. It can still be difficult to find the line but at least you know what you were talking about at a point in time!

Having `// NOSONAR TODO` is the worst possible compromise, so if that's the compromise, then let's remove the rule...

m5r:
@garethbowen [The PR](https://github.com/medic/cht-conf/pull/623) containing the comment removal and the fix is already up for review, that's how short-term the `// NOSONAR` bandaid was supposed to last.

Other than that occurrence, this rule hasn't bothered me when it shouldn't. I'm fine with keeping it but let's do something about the `TODO` comments lying around [our repos](https://github.com/search?q=repo%3Amedic%2Fcht-core+OR+repo%3Amedic%2Fcht-conf+OR+repo%3Amedic%2Fcht-android+OR+repo%3Amedic%2Fcht-sync+OR+repo%3Amedic%2Fmedic-infrastructure+OR+repo%3Amedic%2Fcht-interoperability+OR+repo%3Amedic%2Fcht-pipeline+OR+repo%3Amedic%2Fcht-docs+%2F%28%5E%7C%5CW%29TODO%28%24%7C%5CW%29%2F&type=code)

jkuester:
Since it sounds like @m5r is okay with keeping the rule, and I am content with @garethbowen's explanation for why we should generally avoid TODOs (and no one else has spoken up), I will close this PR and keep the rule in place!  

Thank you everyone for the discussion! :clap: IMHO this is how I hoped Sonar rule conversations could be conducted and now this discussion/decision is captured here in case we ever decide to revisit this.

PR #1484, opened 2024-08-08
Change config to hugo toml

esthermmoturi:
# Description

Refer to [this](https://github.com/medic/cht-docs/pull/1411) PR raised earlier and the comments provided.


esthermmoturi:
I have not observed any significant changes with the change from 'treshold' to threshold. I have referred to [hugo's documentation](https://gohugo.io/content-management/related/) on the same.

esthermmoturi:
File name change was informed by [this](https://gohugo.io/getting-started/configuration/#configuration-file) release update by hugo.

> With v0.109.0 and earlier the basename of the site configuration file was config instead of hugo. You can use either, but should transition to the new naming convention when practical.

hugo.toml
```diff
@@ -202,7 +202,7 @@ enable = false
 RelPermalink= "communityhealthtoolkit.org"
 
 [related]
-  treshold = 1
```
garethbowen, line 205: I think we should just delete this line - I presume it wasn't doing anything before because of the typo, so unless we have a reason to start applying it I think it makes more sense to delete it.

garethbowen (changes requested):
Comment inline.

hugo.toml
esthermmoturi: Done, thanks.

garethbowen (approved):
LGTM. I haven't tested this locally though.

esthermmoturi:
Thanks @garethbowen , I have tested it locally.

PR #1485, opened 2024-08-09
update releasing doc on when to wait

mrjones-plip:
# Description

update releasing doc on when to wait

# License

### DIFF
--- a/content/en/contribute/code/releasing/_index.md
+++ b/content/en/contribute/code/releasing/_index.md
@@ -44,6 +44,17 @@ Given a version number `MAJOR.MINOR.PATCH`, increment the:
 {{% alert title="Info" %}} You can find the versions currently supported, dependencies, and release notes for the CHT Core 
  [on the Releases page]({{% ref "core/releases" %}}). {{% /alert %}}
 
+### When to wait to release
+
+In general, we do not want to wait to do a release.  The moment the release manager feels the release is ready, they should release it.  This may even be with just one ticket! 
+
+There are two conditions that we would wait:
+
+1. There is a new bug found in `master` that has not been released.  This should be fixed before the release.
+1. There is a bug found in an existing release that can be fixed in 5 days.  In less than 5 days the fixed code will be released.  In more than 5 days, the bug will be released and a new release will ship later with the fix for the bug.
+
+The reason we have these rules around when we wait, is so that everyone knows what they are and the release manager never needs to ask "should we wait?" - they'll know already!
+
 ## CHT Conf
 
 Follow the [instructions in the readme](https://github.com/medic/cht-conf/#user-content-releasing).


content/en/contribute/code/releasing/_index.md
```diff
@@ -44,6 +44,17 @@ Given a version number `MAJOR.MINOR.PATCH`, increment the:
 {{% alert title="Info" %}} You can find the versions currently supported, dependencies, and release notes for the CHT Core 
  [on the Releases page]({{% ref "core/releases" %}}). {{% /alert %}}
 
+### When to wait to release
+
+In general, we do not want to wait to do a release.  The moment the release manager feels the release is ready, they should release it.  This may even be with just one ticket! 
+
+There are two conditions that we would wait:
+
+1. There is a new bug found in `master` that has not been released.  This should be fixed before the release.
```
garethbowen, line 53: The wording here feels not specific enough. What about this?

```suggestion
1. There is a bug found in `master` that did not exist in the previous release.  This should be fixed before the release.
```

content/en/contribute/code/releasing/_index.md
```diff
@@ -44,6 +44,17 @@ Given a version number `MAJOR.MINOR.PATCH`, increment the:
...
+1. There is a bug found in an existing release that can be fixed in 5 days.  In less than 5 days the fixed code will be released.  In more than 5 days, the bug will be released and a new release will ship later with the fix for the bug.
```
garethbowen, line 54:
```suggestion
1. There is a high priority bug found in an existing release that we think can be fixed in less than 5 days. The way to determine if a bug is high priority is if we would release a service pack to fix it later. If so then it makes sense to wait for the fix to save us the effort of having to do two releases in quick succession. If it's going to take longer than 5 days then it's worth the additional effort of releasing a service pack. After 5 days if the fix isn't merged, then continue with the release regardless.
```

content/en/contribute/code/releasing/_index.md
```diff
@@ -44,6 +44,17 @@ Given a version number `MAJOR.MINOR.PATCH`, increment the:
...
+
+The reason we have these rules around when we wait, is so that everyone knows what they are and the release manager never needs to ask "should we wait?" - they'll know already!
```
garethbowen, line 56:
```suggestion
The reason we have these rules around when we wait, is so that everyone knows what they are and the release manager never needs to ask "should we wait?" - they'll know already!

While waiting the release manager should use the time effectively by doing as many release steps as possible including writing release notes, preparing the release branch, etc so that the release can be finalized as quickly as possible once the fixes are made.
```

garethbowen (changes requested):

content/en/contribute/code/releasing/_index.md
```diff
@@ -44,6 +44,17 @@ Given a version number `MAJOR.MINOR.PATCH`, increment the:
 {{% alert title="Info" %}} You can find the versions currently supported, dependencies, and release notes for the CHT Core 
  [on the Releases page]({{% ref "core/releases" %}}). {{% /alert %}}
 
+### When to wait to release
+
+In general, we do not want to wait to do a release.  The moment the release manager feels the release is ready, they should release it.  This may even be with just one ticket!
```
andrablaj, line 49: [Per the guidelines](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#using-we), using `we` can be confusing for the reader. There are several mentions of it in this new section and I suggest we rephrase it to avoid the `we`/`us` confusion. 

```suggestion
In general, release managers should not feel they have to wait to do a release. The moment the release manager feels the release is ready, they should release it. This may even be with just one ticket! 
```

content/en/contribute/code/releasing/_index.md
```diff
@@ -44,6 +44,17 @@ Given a version number `MAJOR.MINOR.PATCH`, increment the:
 {{% alert title="Info" %}} You can find the versions currently supported, dependencies, and release notes for the CHT Core 
  [on the Releases page]({{% ref "core/releases" %}}). {{% /alert %}}
 
+### When to wait to release
+
+In general, we do not want to wait to do a release.  The moment the release manager feels the release is ready, they should release it.  This may even be with just one ticket! 
+
+There are two conditions that we would wait:
```
andrablaj, line 51:
```suggestion
There are two scenarios where a release manager should wait before doing the release:
```

content/en/contribute/code/releasing/_index.md
andrablaj: Same here, please replace the mention of `we` with "release manager" or something more generic.

andrablaj (changes requested):
[Per the guidelines](https://docs.communityhealthtoolkit.org/contribute/docs/style-guide/#using-we), using `we` can be confusing for the reader. There are several mentions of it in this new section and I suggest we rephrase it to avoid the `we`/`us` confusion for the reader.

content/en/contribute/code/releasing/_index.md
mrjones-plip: taking Andra's revision in 20da0fce

content/en/contribute/code/releasing/_index.md
mrjones-plip: took both in 35a099a8  - thanks!

andrablaj (approved):
👍 Thanks for the updates!

garethbowen (approved):

Issue #1486, opened 2024-08-09
subtitle Anchor links missing after hugo upgrade


mrjones-plip:
docs site used to have anchor links on sub-titles but after the [recent upgrade](https://github.com/medic/cht-docs/pull/1437) in hugo they're gone.

here's what they used to do (below), now there's nothing when you hover.

https://github.com/user-attachments/assets/f31df4c2-0ca0-4b19-8d7c-e4575c8b6305

esthermmoturi:
[Documentation](https://www.docsy.dev/docs/adding-content/navigation/#heading-self-links) on self-heading links.

esthermmoturi:
Changed in Docsy v**0.9.0**

PR #1487, opened 2024-08-12
Adding heading self link

esthermmoturi:
Fixes [the issue on Heading self links](https://github.com/medic/cht-docs/issues/1486) missing after upgrade.


mrjones-plip (changes requested):
Nice work!  works as advertised

is it possible to get back the chain icon instead of the hashtag/pound?  Old, preferred is on the right, new one is on the left:

image

mrjones-plip:
I think the  :link: character you mentioned in slack would be plenty good if we can use that!

mrjones-plip (approved):
lgtm!

PR #1495, opened 2024-08-21
chore: document preference for ts over js

garethbowen:
# Description

Document preference for ts over js

# License

### DIFF
--- a/content/en/contribute/code/style-guide.md
+++ b/content/en/contribute/code/style-guide.md
@@ -8,14 +8,21 @@ description: >
 
 ## Language
 
-Prefer JavaScript or TypeScript wherever possible, including in webapps, on the server, and for scripting. This is because:
-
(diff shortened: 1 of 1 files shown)


andrablaj (approved):
Looks food as a first version of the guidelines. Thanks for putting it together.

In `cht-interoperability`, TypeScript was used to build the [mediator APIs](https://github.com/medic/cht-interoperability/tree/main/mediator/src). Having the request and response objects implemented with TS improves the clarity of the APIs and helps catch model-related errors early in the development process (especially when working with standards like FHIR, for example). 

Would this be a good case where TS could be preferred over JS on the server-side?

garethbowen:
Yes absolutely. Running TS on the server has the same upside as running it on the browser, but you have to pay the compile tax, so it's a tradeoff.

PR #1505, opened 2024-08-27
feat: update known Enketo issues for 4.x upgrade

jkuester:
# Description

We still have partners upgrading from `3.x` to `4.x`, so the Upgrade Prep guide remains relevant documentation.  However, there are some changes to the "Known issues" with Enekto in the more recent versions of `4.x` (which are likely the versions that partners will be upgrading to).  

This PR updates the docs around the known Enketo issues with the goal of providing useful and relevant information to partners currently trying to prepare for an upgrade.  

# License


mrjones-plip (approved):
Looks good - thank you!

Approving, but suggesting a minor tweak to make users not have to look as hard.  But it's minor for sure

PR #1516, opened 2024-08-30
Update README.md

esthermmoturi:
Just a quick update since we changed the configuration file from `config.toml` to `hugo.toml`


mrjones-plip (approved):
:ship:

Issue #1129, opened 2023-06-29
Update local setup guide


jkuester:
https://docs.communityhealthtoolkit.org/apps/tutorials/local-setup/ 

- Update macOS steps for Python install to point to Python3
- Update Docker Linux steps to use Docker Desktop for Linux (verify that works first...)

esthermmoturi:
Hey @jkuester , are you still up to updating this?

jkuester:
@esthermmoturi I can give this a shot.  Do you happen to have macOS?  I just need to confirm with someone about the commands for installing python....

esthermmoturi:
Hey @jkuester , I do have macOS.

PR #1531, opened 2024-09-09
feat(#1129): update local setup guide

jkuester:
# Description

Closes https://github.com/medic/cht-docs/issues/1129

Guilde now includes Docker Desktop for Linux and Python 3 for macOS.

# License


content/en/apps/tutorials/local-setup.md
```diff
@@ -77,12 +77,7 @@ nvm install {{< param nodeVersion >}}
 {{< tab header="macOS" >}}
 # Uses Homebrew: https://brew.sh/
 brew update
-brew install curl jq pyenv git make node@{{< param nodeVersion >}} gcc
-# Python no longer included by default in macOS >12.3
-pyenv install 2.7.18
-pyenv global 2.7.18
-echo "eval \"\$(pyenv init --path)\"" >> ~/.$(basename $SHELL)rc
-. ~/.$(basename $SHELL)rc
+brew install curl jq git make node@{{< param nodeVersion >}} gcc python
```
jkuester, line 80: @esthermmoturi I am confident that we want Python 3 installed (even on macOS) since that is what we are using on all the other platforms.  Reading [this guide](https://docs.python-guide.org/starting/install3/osx/) it seems like you can just use `brew` to directly install Python 3.  This seems like the most straightforward way to do it.  

Since you have macOS, let me know if there is a different way we should have people install Python (or if there is some reason that we cannot just do it with this one brew command....).  Thanks

esthermmoturi: Hey @jkuester , I ran the code and my python version still remains as 2.7.18(This is what I had before running the code).

jkuester: @esthermmoturi couple followup questions!  

- What version of macOS are you running?
- Do you remember how you previously installed Python?  (Or were you just using the version that shipped with macOS?)

esthermmoturi: > @esthermmoturi couple followup questions!
> 
> * What version of macOS are you running?
 - Apple M1
> * Do you remember how you previously installed Python?  (Or were you just using the version that shipped with macOS?)

esthermmoturi:
I see that python3 is already installed but when I run `python --version` I still get 2.7.18, I am not sure if I am missing something.
image

content/en/apps/tutorials/local-setup.md
jkuester: Okay, so based on this info from the article I linked above:

> Mac OS X comes with Python 2.7 out of the box between versions 10.8 and 12.3.

Can you check your exact macOS software version (by running the `sw_vers` command)?  If it is within that range, then we can definite assume that Python 2 is just coming from your macOS system utils.

In that case, can you run `brew info python` and see what it prints out?  Specifically, I think we need to [follow these steps](https://stackoverflow.com/a/51939526) and add brew's Python to your `PATH` .

esthermmoturi: Sorry, I totally missed this question. Here is the answer:
ProductName:	macOS
ProductVersion:	12.1
BuildVersion:	21C52

python@3.12: stable 3.12.6 (bottled)

jkuester: > ProductVersion: 12.1

Okay, so yes. I think you are in the situation described above where you have Python 2 pre-installed already by macOS.  Your `python` command is automatically mapped to this version.  Then, as described by the SO post that I linked to, when you installed Python via homebrew, it installed Python 3 like we wanted. However, it did not overwrite the `python` command, but instead left that mapped to the original Python 2.  I believe you should be able to run `python3 --version` and see the version installed by homebrew.  

To get the homebrew python to _override_ the actual `python` command, we need to add the new python binaries to the _front_ of your `PATH`.  You can run `brew info python` to double-check exactly where homebrew installed python, but the SO post suggests it is probably `/usr/local/opt/python/libexec/bin` (though this might not be the case for macOS 12...).  Assuming that is the correct directory, then all you should need to do is add the following line to your `~/.profile` file:

```shell
export PATH="/usr/local/opt/python/libexec/bin:$PATH"
```

Then you should see Python 3 when running: `source ~/.bash_profile && python --version`.

jkuester: Okay scratch that.  :sweat_smile: I had a call with Tatiana this afternoon and she helped me work through some issues with using the `python` from Homebrew.  Basically the challenge is that if Homebrew detects a version of python already installed (e.g. the python2 that comes with many versions of macOS) it will _not_ overwrite the `python` command, but instead just includes the Homebrew version of python on your PATH as `python3`.  It is possible to manually overwrite the `python` command by adding the Homebrew python to the front of your PATH.  That is what I was suggesting in my previous comment.  

Unfortunately, the directory path where Homebrew installs python does not seem to be the same on variouis version of macOS and even worse, on Tatiana's machine python was installed in a directory based on the actual version of python (so something like `~/.local/brew/python3.13`).  This makes it very difficult to add a simple line to your `.bashrc` file for adding Python to the PATH since the _brew python installation directory_ could change in the future when Homebrew updates python.  

Instead of taking our chances trying to figure out a universally workable manual solution here, it seems like a much more stable idea to just stick with `pyenv`.  The `pyenv init --path` command that we were already adding to the .bashrc file handles all this for us.  

So, I have updated the docs here to add back in the `pyenv` code (but this time installing Python 3).

esthermmoturi: Thanks @jkuester for taking the time to troubleshoot.

esthermmoturi (approved):
LGTM!

PR #1532, opened 2024-09-09
feat: update docker image to have proper Hugo version

jkuester:
# Description

The `jakejarvis/hugo-extended` is unmaintained (as is the `klakegg/hugo` which I have also used).  Instead, there is a different [community maintained image](https://discourse.gohugo.io/t/the-up-to-date-hugo-docker-images/43293).  

I am honestly a bit mystified by the terrible SEO for `hugomods/hugo`.  Just from the pull counts it is orders of magnitude more popular than other images, but does not surface at the top in the results list on Docker Hub or in a DuckDuckGo search....

It looks like the [base versions of the image](https://docker.hugomods.com/docs/tags/#latest) contain what we need!  (Extended + Go + Node...). Might be able to get away with one of the more specific ones, but not sure how useful it is to get super picky about the size of a docker image for development usage... :sweat_smile: 

I based the compose.yml updates the [documented compose config](https://docker.hugomods.com/docs/development/docker-compose/).  

# License

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,3 +1,3 @@
-hugo extended_0.122.0
+hugo extended_0.122.0 # Should match the version in compose.yml
 golang 1.21.0
 nodejs 16.20.2
--- a/compose.yml
+++ b/compose.yml
@@ -3,9 +3,9 @@ version: '3'
 services:
   hugo:
     container_name: cht-hugo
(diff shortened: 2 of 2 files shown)


mrjones-plip (approved):
:x_ray:

PR #1536, opened 2024-09-11
App Dev Hosting: work work manual steps, move docker helper to the top

mrjones-plip:
# Description

* update the manual app dev to be more like [prod docker](http://localhost:1313/hosting/4.x/self-hosting/single-node/) docs
* `docker-compose` -> `docker compose`
* always use `.env` files
* remove now redundant `.env` section
* move docker helper to be the top most section
* clearly call out that manual `docker compose` and Docker Helper achieve the same end result (as opposed to needing to run both)

# License

### DIFF
--- a/content/en/hosting/4.x/app-developer.md
+++ b/content/en/hosting/4.x/app-developer.md
@@ -27,26 +27,50 @@ docker kill $(docker ps -q)
 After meeting these requirements, create a directory and download the developer YAML files in the directory you want to store them. This example uses `~/cht-4-app-developer` as the directory:
 
 ```shell 
-mkdir  ~/cht-4-app-developer && cd ~/cht-4-app-developer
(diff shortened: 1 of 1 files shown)


nsitaula:
@mrjones-plip as mentioned in our discussion, I see the documentation "[App Developer Hosting in CHT 4.x](https://docs.communityhealthtoolkit.org/hosting/4.x/app-developer/)" needs to be restructured so that we are clearly mentioning there are two alternative ways and as we discussed may be we can move "[Docker Helper](https://docs.communityhealthtoolkit.org/hosting/4.x/app-developer#cht-docker-helper-for-4x)" to the top in documentation or considering removing the other alternative way.

mrjones-plip:
thanks @nsitaula !  I've committed this change in 9e6ce70e.

PR #1549, opened 2024-09-18
Another set of building section migration

esthermmoturi:
[This](https://docs.google.com/document/d/1Vnq3HgXDQxtiHi4m3UJORnUzB1dj5H4wfxt1Q4kIEgQ/edit#heading=h.smca4jwqayex) is the vision. This PR migrates pre-requisites and getting started to top level navigation.

### DIFF
--- a/content/en/building/concepts/_index.md
+++ b/content/en/building/concepts/_index.md
@@ -1,6 +1,6 @@
 ---
 title: "Concepts"
-weight: 2
+weight: 3
 description: >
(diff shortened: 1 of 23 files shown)


michaelkohn (approved):
I don't really know what to test or what is included in this PR so all I did was validate that it builds and loads and that i can navigate around a bit.

esthermmoturi:
Thanks @michaelkohn , it was just a migration of files as is to higher navigation(removed them from folders), then I added aliases and confirmed all links worked.

PR #1550, opened 2024-09-19
Fixing image displacement caused by the move

esthermmoturi:
Due to the move of folders in [this](https://github.com/medic/cht-docs/pull/1549) PR, the images were displaced. This PR moves the folder that contains the images so that they are accessible. They are currently inaccessible as seen on the docsite. - https://docs.communityhealthtoolkit.org/building/local-setup/#cht-conf


latin-panda (approved):
LGTM!

PR #1557, opened 2024-09-23
feat(#1464) Adds docker - k8s migration guide

eljhkrr:
# Description

#1464

# License

### DIFF
--- a/content/en/hosting/kubernetes-vs-docker.md
+++ b/content/en/hosting/kubernetes-vs-docker.md
@@ -33,6 +33,7 @@ The main components of a Kubernetes deployment include:
 * A CHT HAProxy pod.
 * Upgrade Service pod.
 * CHT-Sentinel pod.
+
 ## Docker Compose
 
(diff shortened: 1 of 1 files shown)


mrjones-plip (changes requested):
yay!  so great to see some solid additions to the docs around hosting & migration - thanks @eljhkrr !

This shouldn't go in the ["Kubernetes vs Docker" page](https://docs.communityhealthtoolkit.org/hosting/kubernetes-vs-docker/).  There's already a [Data migration to 4.x](https://docs.communityhealthtoolkit.org/hosting/4.x/data-migration/) page, so maybe create a peer to that page, "Migrate 4.x from Docker to Kubernetes" or something?

cc @esthermmoturi and @n-orlowski who have been doing a bunch of re-org around docs pages.

henokgetachew:
I think the latest version of the helm charts support the gaps that were the reasons for adding the custom templates here. There are also a couple guides in the docs site for how to do the migrations. From 3.x Docker to 4.x k3s. I think the Only thing you would not need to do would be the upgrading from 3.x. Let me know if I can answer questions there.

andrablaj:
@eljhkrr  what is the plan with this PR? Can we merge it? Does it need extra work? cc @Hareet @mrjones-plip

mrjones-plip:
thanks @andrablaj !  I've moved the document into the correct location in the new hierarchy and made some light formatting.  I see the PR is otherwise missing the helm calls and how to copy existing data from your old location in docker to your new location in k8s.

cc @lorerod in case the work you're doing recently covers this and you can swoop in and finish the PR.  No worries if not!

mrjones-plip:
Unfortunately, no, this is net new documentation that likely would still be good to have and the [recent work](https://github.com/medic/cht-docs/pull/2110) I did doesn't cover it. 

@eljhkrr - any interest in resuming work on this PR to cleaning it up? It will need to account for using [latest helm charts](https://github.com/medic/helm-charts/) I believe?

PR #1563, opened 2024-09-26
Rename index page to contacts

esthermmoturi:
To improve image management, I placed them under a folder and renamed the index file to contacts.md. This is how it will look.
image


esthermmoturi:
@michaelkohn , after merging, I realized that the images needed to be within a folder to look neat. This is how the index page will now look.
image
image

nsitaula (approved):

Issue #1552, opened 2024-09-19
change weekly muffet link checks to not be run by a teamate


mrjones-plip:
currently when there's a broken link, [CI will catch it](https://github.com/medic/cht-docs/commit/6d23ad42b29c45f3d1f8c7214091d06c058e04c4/checks). then [an alert is sent](https://medic.slack.com/archives/CUXB9SVT5/p1726370895810419) to slack, but it's currently from `nydr` , a former teammate. 

> APP  [Saturday at 8:28 PM](https://medic.slack.com/archives/CUXB9SVT5/p1726370895810419)
> [nydr](http://github.com/nydr)nydr
> Ref: refs/heads/main      Event schedule
> Actions URL https://github.com/medic/cht-docs/commit/6d23ad42b29c45f3d1f8c7214091d06c058e04c4/checks
> CHT DOCS BAD LINKS
> Broken Links Found
> [Powered By rtCamp's GitHub Actions Library](https://github.com/rtCamp/github-actions-library)

We should make this all run and be sent from either a current teammate or, even better, a bot/automated account that's not tied to anyone

mrjones-plip:
#1567 failed to fix this issue - I'll have another go!

PR #1567, opened 2024-09-26
fix(#1552): GH user link validation slack alerts come from

mrjones-plip:
# Description

Fix which GH user link validation slack alerts come from per medic/cht-docs/issues/1552 . I think the last person to edit the CI is the GH user that slack shows.  Let's find out!

# License


andrablaj (approved):
Let's find out!

PR #1570, opened 2024-09-30
chore: Deprecate couch2pg and cht-couch2pg

andrablaj:
# Description

https://github.com/medic/couch2pg/issues/46

# License


mrjones-plip (approved):
really thorough work here - nice job!  I especially appreciate keeping couch2pg specific pages to help those using it, but clearly marking them as deprecated \o/

Issue #827, opened 2022-11-08
Fix source of text on task


abbyad:
Page: [/design/best-practices/](https://docs.communityhealthtoolkit.org/design/best-practices/#anatomy-of-a-task)
Version: 2021-07-16 08:15:33 -0500 -0500
---

The second line in a task is specified both as the form title (incorrect), and the task name (correct). We should update for accuracy.

![image](https://user-images.githubusercontent.com/789512/200682540-19061421-cdd4-44c3-b043-2163f68f798a.png)

tmatheas:
Hi @abbyad , Can you please assign this issue to me. Also can you please add the `hacktoberfest` tag to this issue.

tmatheas:
@andrablaj Can you explain the issue a little bit ?

andrablaj:
@latin-panda do you think you can support @tmatheas with his question? I assume this is just about updating the highlighted line to match what the actual second line is about?

latin-panda:
@tmatheas Thanks for picking this up! 

[This page](https://docs.communityhealthtoolkit.org/design/best-practices/)'s image is showing wrong information 

To avoid confusion is better to say "task" instead of "form", and say "title" instead of "name" since it aligns better to the terms of task configuration

<img width="800" alt="Screenshot 2024-10-01 at 4 07 20 PM" src="https://github.com/user-attachments/assets/f152a225-56b8-4eff-b389-02e133d64173">

 <br> <br>

In [this other page](https://docs.communityhealthtoolkit.org/building/reference/tasks/), we already have another image that seems to use better terms. I think you can use it:  
<img width="800" alt="Screenshot 2024-10-01 at 4 14 03 PM" src="https://github.com/user-attachments/assets/285d92c4-13af-48cb-82ff-1ae67fa847b2">

tmatheas:
@latin-panda I have made the changes. Can you check this is fine ?
<img width="892" alt="image" src="https://github.com/user-attachments/assets/44608c8c-3e7c-4b4d-bb26-7b722b4030f2">

latin-panda:
@tmatheas That looks good! Once [you open a pull request](https://docs.communityhealthtoolkit.org/contribute/code/workflow/#pull-requests) to the `main` branch, feel free to add me as a reviewer

tmatheas:
#1580 Linking the PR

tmatheas:
[https://github.com/medic/cht-docs/pull/1581](url)

tmatheas:
[...]

PR #1581, opened 2024-10-01
fix(#827): Change source of text from form to task

tmatheas:
To avoid confusion changed to "task" instead of "form", and say "title" instead of "name" since it aligns better to the terms of task configuration.

Closes https://github.com/medic/cht-docs/issues/827


tmatheas:
Assigning @latin-panda as a reviewer

latin-panda (approved):
LGTM! Thank you!

PR #1590, opened 2024-10-04
Add more troubleshooting entry

kapelo:
This covers cases when a user session times out during E2E test run

# Description

[description]

# License

### DIFF
--- a/content/en/contribute/code/core/automated-tests.md
+++ b/content/en/contribute/code/core/automated-tests.md
@@ -231,6 +231,9 @@ If you experience errors such as:
 ```
 A user session timing out while running the test.
 ```
+This could be because there are some issues with data or there could be a large number of images and volumes that caused docker to crash.
+We recommend cleaning all the docker data to be able to start afresh and clean. It is important, however, to note that these commands will delete everything.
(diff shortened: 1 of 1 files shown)


tatilepizs (changes requested):
Thank you for being proactive about improving the documentation ⭐ 

We have to be very careful with those commands because they will delete everything, so can you please add more details about what will happen if someone runs them?

For example, you can explain that the error might happen because some data might be causing issues or because having a large number of images and volumes might crash docker depending on the resources assigned and that, as a solution, we recommend cleaning all the docker data to be able to start fresh and clean, but that is important to note that those commands will delete everything, and the next time that the e2e tests are run it might take a little longer because all the images need to be downloaded again.

Benmuiruri:
Typically for docs PR we need one reviewer from QA or a software developer for the technical part. 

Sometimes you may include @esthermmoturi if perhaps you have a question about which section of the docs the content should go, or the language of the content. 

Since you have Lorena and Tatiana, I'll remove myself as a reviewer not to block you.

tatilepizs (approved):
Thank you for the clarification @kapelo

PR #1592, opened 2024-10-05
fix(#1586): Removed ToC as it is a duplication of the side bar

### LINKED ISSUE #1586: Remove ToC as it is a duplication of the side bar
https://docs.communityhealthtoolkit.org/building/guides/privacy/policy/

tmatheas:
closes #1586


mrjones-plip (approved):
Thanks very much @tmatheas !

PR #1598, opened 2024-10-07
chore: Make Muffet ignore 403 HTTP Status Codes

andrablaj:
# Description

Make Muffet (the link checker) ignore `403` HTTP Status Codes, as tools like Cloudflare usually trigger them.

Removed link to nonexistent GitHub project.

# License


andrablaj:
@mrjones-plip @esthermmoturi I noticed that the main reason for which the link checker fails is `403` statuses (most probably triggered by tools like Cloudflare). I suggest we add `403` to the list of accepted codes. This will avoid adding more `exclude` links to `muffet.sh`.

Additionally, as most of the excluded links are due to `403`s, I suggest we remove them from the `muffet.sh` file.

mrjones-plip (approved):
Sure - let's give it a try!

andrablaj:
@mrjones-plip @esthermmoturi: 

- I removed the links that were returning `403`s from the `exclude` list in `muffet.sh`
- I upgraded `muffet` to the latest version so we can use the `accepted-status-codes` functionality.

Any concerns with this approach?

mrjones-plip:
@andrablaj - no concerns - still looks good!

Let's check the results of the run on ~Monday to ensure it's working as expected (false negatives being the main concern).

PR #1601, opened 2024-10-09
fix: updates training form with correct ID

latin-panda:
# Description

The training form has the incorrect ID

# License


latin-panda:
@esthermmoturi This is just fixing a bad copy/paste, and making the form ID about the "floating action buttons" instead of the "more options menu"

esthermmoturi (approved):
Looks good!

PR #1605, opened 2024-10-09
Dynamic shell for autocomplete on k8s docs

craig-landry:
Changes instructions and copy/paste code snippet to use the current shell for the autocomplete setup. This allows someone to run the code as documented without needing to change the shell value from zsh to whatever they are using (even though zsh is the correct answer ;) ).

### DIFF
--- a/content/en/contribute/code/core/deploy-on-eks.md
+++ b/content/en/contribute/code/core/deploy-on-eks.md
@@ -27,8 +27,8 @@ Be sure you have these tools installed and repos cloned:
 Both `helm` and `kubectl` have autocomplete libraries. For power users and beginners alike, it adds a lot of discoverability. This code is for `zsh`, but `bash`, `fish` and `powershell` are supported as well:
 
    ```shell
-   source <(kubectl completion zsh)
-   source <(helm completion zsh)
+   source <(kubectl completion "$(basename "$SHELL")")
+   source <(helm completion "$(basename "$SHELL")")
    ```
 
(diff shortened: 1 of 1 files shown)


craig-landry:
@mrjones-plip What do you think about this? I like that it allows someone to effortlessly copy/paste and run it, but it also makes the commands look kinda convoluted. I'm cool either way and would like your take.

craig-landry (commented):
A few bonus notes:

1. This needs testing on a Mac.
2. I understand `$0` can get such a value, but from reading around, I think it works differently on a Mac. (hence the item above)

mrjones-plip (approved):
works for me in my testing! Lemme add @tatilepizs as a reviewer to make sure macOS is cool too

tatilepizs:
@mrjones-plip I was checking the change, but I am not sure what do I need to try. Could you please help me with more detailed instructions?

mrjones-plip:
sure thing @tatilepizs ! basically, are you able to enable the `kubectl` and `helm` shell autocomplete libraries by following the [old steps](https://docs.communityhealthtoolkit.org/contribute/code/core/deploy-on-eks/) as well as the new steps in this PR - they should work the same.

Here's a little video of what they do without and with being enabled (AMAZINGESTEVAR AUTOCOMPLETE :heart_eyes: !!):

[Screencast from 2024-10-09 12-33-21.webm](https://github.com/user-attachments/assets/88a03fd5-ad8d-4a26-abf9-bf68c7e034ec)

tatilepizs (approved):
You are great @mrjones-plip, thank you for the details! ⭐ 

It works for me 😊 

[video](https://github.com/user-attachments/assets/225bbbc4-413e-4b14-84f6-c2ec404a5a20)

craig-landry:
Thank you both for verifying. I love the videos and knowing exactly what was attempted and how it ran!

PR #1606, opened 2024-10-09
fix(na): fix link for api docs in 4.11 release, correct cursor number in API docs

mrjones-plip:
# Description

* fix link for api docs in 4.11 release
* we're zero based for person API pagination, start at 0 instead of 1

# License

### DIFF
--- a/content/en/building/reference/api.md
+++ b/content/en/building/reference/api.md
@@ -874,7 +874,7 @@ Content-Type: application/json; charset=utf-8
 }
 ```
 
-2. Get the first 3 people of type `person` by using cursor `0`.
+2. Get 3 people of type `person` by using cursor `0`.
 
 ```
 GET /api/v1/person?type=person&cursor=0&limit=3
--- a/content/en/building/reference/api.md
+++ b/content/en/building/reference/api.md
@@ -874,10 +874,10 @@ Content-Type: application/json; charset=utf-8
 }
 ```
 
-2. Get 3 people of type `person` by using cursor `1`.
+2. Get the first 3 people of type `person` by using cursor `0`.
 
 ```
-GET /api/v1/person?type=person&cursor=1&limit=3
+GET /api/v1/person?type=person&cursor=0&limit=3
 ```
 
 ```
--- a/content/en/building/reference/api.md
+++ b/content/en/building/reference/api.md
@@ -874,10 +874,10 @@ Content-Type: application/json; charset=utf-8
 }
 ```
 
-2. Get 3 people of type `person` by using cursor `0`.
+2. Get the second set of 100 people of type `person` by using cursor `100`.
 
 ```
-GET /api/v1/person?type=person&cursor=0&limit=3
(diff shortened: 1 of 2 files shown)


content/en/building/reference/api.md
```diff
@@ -874,10 +874,10 @@ Content-Type: application/json; charset=utf-8
 }
 ```
 
-2. Get 3 people of type `person` by using cursor `1`.
+2. Get the first 3 people of type `person` by using cursor `0`.
```
jkuester, line 877: So, _technically_ passing `cursor=0` will give you the "first" people, but that is just because of an implementation detail of how we have implemented paging (and one that might change in future releases).

The actual value of a cursor is only promised to be a string "token" value that _the server_ will interpret to know which page to return. Consumers should only pass in token values that were returned from the server on previous calls to the same API. If you want the "first" page of people, you should not pass any cursor.

content/en/core/releases/4.11.0.md
```diff
@@ -29,7 +29,7 @@ None.
 The CHT datasource API has been expanded to allow querying places and persons by contact type.
 
 ##### Details
-See more details in the [api docs](http://docs.communityhealthtoolkit.org/cht-datasource/).
+See more details in the [API docs](https://docs.communityhealthtoolkit.org/building/reference/api/#get-apiv1person).
```
jkuester, line 32:
```suggestion
See more details in the [API docs]({{< relref "building/reference/api#get-apiv1person" >}}).
```

jkuester (approved):
Added comment inline regarding the changes to the API doc.  IMHO we can leave it as it was originally, but happy to review further iterations of the wording if what we have now does not make sense!

content/en/building/reference/api.md
mrjones-plip: ah - gotcha - I'm down with being technically accurate!

However, you could still miss a record (first or otherwise) but not starting at zero I believe.  Let's go with:

```suggestion
2. Get 3 people of type `person` by using cursor `0`.
```

yeah?

content/en/core/releases/4.11.0.md
mrjones-plip: ugh - that was lazy of me - thanks for the fix!

mrjones-plip:
no rush on this fix - will wait for another spot review!

content/en/building/reference/api.md
jkuester: :thinking: Still not a fan of using `0` for the example cursor value.  I do now see your concern though with putting `cursor=1` in the example request since it may not be clear that things are 0-based and this is not where things should start.  What do you think about using `cursor=100`?  The previous example does not pass any cursor/limit params so it will return the first 100 persons.  Then, to get the next persons, you would pass `cursor=100`.

mrjones-plip: oh - maybe?  I'm surprised about how hard this one change is :sweat_smile: 

Fixed in 5e5260d - thoughts?

jkuester (approved):

PR #1613, opened 2024-10-14
chore: downgrade muffet

andrablaj:
# Description

Reverting [these changes](https://github.com/medic/cht-docs/pull/1598/files), but keeping the links to catch "fresh" broken ones. 

# License


esthermmoturi (approved):
Thanks for this @andrablaj

PR #1616, opened 2024-10-15
fix(na): move medic EKS how to -> at medic 

mrjones-plip:
# Description

move medic EKS how to to at medic section, ad alias to redirect old URL

# License


esthermmoturi (approved):
Approved!

PR #1622, opened 2024-10-16
Move training section

esthermmoturi:
- Move this [training section](https://docs.communityhealthtoolkit.org/building/guides/training/) to top level navigation since we are moving away from guides, tutorials etc.

- Add aliases.
- Weighting will be done later to configure chronological order


mrjones-plip (approved):
All the alias redirect old traffic, all the images still load, related content links are updated - Nice work on this!

PR #1623, opened 2024-10-16
add where clause to exclude null versions on telemetry

mrjones-plip:
# Description

avoids returning results in SQL for `null` versions in telemetry which we already do in the [2nd query](https://docs.communityhealthtoolkit.org/building/guides/updates/preparing-for-4/#users-in-need-of-upgrade)

# License


Marina-Kenf (approved):
LGTM.

PR #1630, opened 2024-10-18
docs: update setup instructions for CHT Watchdog on CHT Sync to clarify target naming

### LINKED ISSUE #1631: Update the CHT Watchdog setup docs when setting up targets
**What feature do you want to improve?**
The [documentation](https://docs.communityhealthtoolkit.org/hosting/monitoring/setup/#cht-sync-data-local) currently does not make it clear that the key in the `targets` section in the `sql_servers.yml` file needs to be the same as the instance URL and not something generic.

njuguna-n:
# Description

Closes #1631 

# License

### DIFF
--- a/content/en/hosting/monitoring/setup.md
+++ b/content/en/hosting/monitoring/setup.md
@@ -131,7 +131,7 @@ With the [release of 1.1.0](https://github.com/medic/cht-watchdog/releases/tag/1
    - targets:
       "db-example-com": 'postgres://db_user:db_password@db.example.com:5432/cht?sslmode=disable' # //NOSONAR - password is safe to commit
    ```
-   You may add as many targets as you would like here - one for each CHT Core instance in your `cht-instances.yml` file. Be sure to give each entry a unique name based of the Postgres server (eg `db-example-com` as shown).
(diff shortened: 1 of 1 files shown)


content/en/hosting/monitoring/setup.md
```diff
@@ -131,7 +131,7 @@ With the [release of 1.1.0](https://github.com/medic/cht-watchdog/releases/tag/1
    - targets:
       "db-example-com": 'postgres://db_user:db_password@db.example.com:5432/cht?sslmode=disable' # //NOSONAR - password is safe to commit
    ```
-   You may add as many targets as you would like here - one for each CHT Core instance in your `cht-instances.yml` file. Be sure to give each entry a unique name based of the Postgres server (eg `db-example-com` as shown).
+   You may add as many targets as you would like here - one for each CHT Core instance in your `cht-instances.yml` file. Be sure to match the key (`db-example-com` in this example) to the exact value define in your your `cht-instances.yml` file while also enduring that the entry is unique.
```
andrablaj, line 134:
```suggestion
   You may add as many targets as you would like here - one for each CHT Core instance in your `cht-instances.yml` file. Be sure to match the key (`db-example-com` in this example) to the exact value defined in your `cht-instances.yml` file while ensuring the entry is unique.
```

andrablaj (changes requested):
As @njuguna-n is off next week, and @mrjones-plip is not online, adding my review here to try to have this merged before @njuguna-n's EOD.

andrablaj (approved):

PR #1656, opened 2024-10-25
fix(na):  specify correct port for production CHT Core deployments

mrjones-plip:
# Description

Add comment about correct port for production CHT Core deployments

closes medic/cht-sync/issues/175

# License

### DIFF
--- a/content/en/hosting/analytics/setup-docker-compose.md
+++ b/content/en/hosting/analytics/setup-docker-compose.md
@@ -24,6 +24,8 @@ This guide will walk you through setting up a deployment of CHT Sync with the CH
 
 In the `cht-sync` folder, copy the values from the `env.template` file to a `.env` file. For more information, see the references on the [Environment variables page]({{< relref "hosting/analytics/environment-variables" >}}).
 
+For production CHT Core deployments, most likely the port will need to be set to `443` like this: `COUCHDB_PORT=443`. This is because CHT Core uses an `nginx` reverse proxy on port `443`, instead the default `5984` port used in a stand alone CouchDB instance which the `env.template` [has](https://github.com/medic/cht-sync/blob/bbe9801b678761165262d0f1325007fc9eae31cf/env.template#L22).
+
 {{% alert title="Note" %}}
 The first time you run the commands from any of the sections below it will need to download many Docker images and will take a while. You'll know it's done when you see `#8 DONE 0.0s` and you are returned to the command line. Be patient!
 {{% /alert %}}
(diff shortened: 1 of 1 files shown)


content/en/hosting/analytics/setup-docker-compose.md
```diff
@@ -24,6 +24,8 @@ This guide will walk you through setting up a deployment of CHT Sync with the CH
 
 In the `cht-sync` folder, copy the values from the `env.template` file to a `.env` file. For more information, see the references on the [Environment variables page]({{< relref "hosting/analytics/environment-variables" >}}).
 
+For production CHT Core deployments, most likely the port will need to be set to `443` like this: `COUCHDB_PORT=443`. This is because CHT Core uses an `nginx` [reverse proxy]({{< relref "core/overview/architecture#overview" >}}) on port `443`, instead the default `5984` port used in a stand alone CouchDB instance which the `env.template` [has](https://github.com/medic/cht-sync/blob/bbe9801b678761165262d0f1325007fc9eae31cf/env.template#L22).
```
1yuv, line 27: What's the default port cht-sync tries to connect if the port is not provided? Can you also mention commenting out the COUCHDB_PORT entirely is also the way?

mrjones-plip: It's not the default CHT Sync port per se, it's [the Node Axios library](https://axios-http.com/) we're using. As with any http library, like `curl` for example, if you specify a secure connection (`COUCHDB_SECURE=true`) and don't specify a port, the implicitly declared port is `443`.  However, when I use this with Docker Helper, I specify `COUCHDB_SECURE=true` and am very happy I can easily set the port to `10433`.  

I think keeping this variable declared and explicit better than undeclared and implicit.

andrablaj, line 27:
```suggestion
For production CHT Core deployments, most likely, the port will need to be set to `443` like this: `COUCHDB_PORT=443`. This is because CHT Core uses an `nginx` [reverse proxy]({{< relref "core/overview/architecture#overview" >}}) on port `443`, instead of the default `5984` port used in a stand alone CouchDB instance which the `env.template` [has]({{< relref "hosting/analytics/environment-variables" >}}).
```

andrablaj (changes requested):
Suggestion to link to the env variables docs site rather than specific lines in the `env.template` + small typo. Otherwise, it looks good!

mrjones-plip:
good call!

andrablaj (approved):
📦

PR #1663, opened 2024-10-29
Move index page

esthermmoturi:
Closes https://github.com/medic/cht-docs/issues/1644

It adds an index overview page in [this](https://docs.communityhealthtoolkit.org/building/tasks/) section


content/en/building/tasks/_index.md
```diff
@@ -5,4 +5,62 @@ identifier: "Tasks"
...
+<div class="container">
+  <div class="row">
+{{< figure src="tasks-care1.png" link="tasks-care1.png" class="col-6 col-lg-3" >}}
+{{< figure src="tasks-care2.png" link="tasks-care2.png" class="col-6 col-lg-3" >}}
+{{< figure src="tasks-care3.png" link="tasks-care3.png" class="col-6 col-lg-3" >}}
+{{< figure src="tasks-care4.png" link="tasks-care4.png" class="col-6 col-lg-3" >}}
+  </div>
+</div>
+
+For more information on Care Guides, see the “[Decision Support for Care Guides]()” section of this overview.
```
andrablaj, line 46: The link is missing for the `Decision Support for Care Guides`

content/en/building/tasks/_index.md
```diff
@@ -5,4 +5,62 @@ identifier: "Tasks"
...
+    
+Alternatively, there is an option to configure Household Tasks. When this [permission](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/user-permissions/) is enabled, once a CHW has completed a task, they are taken to the Other Household Tasks page. This page shows the CHW all the additional outstanding tasks within the same household in which the initial task was completed.
+
+CHWs are able to complete tasks directly from this page, or exit by tapping on the “X”. If the household has no additional tasks, they will return directly to the main task list.
+    
+## Profile Page
+
+{{< figure src="tasks-profile.png" class="right col-6 col-lg-3" >}}
+
+Tasks are also accessed from a the People tab in the app.
```
andrablaj, line 62:
```suggestion
Tasks are also accessed from the People tab in the app.
```

content/en/building/tasks/_index.md
```diff
@@ -5,4 +5,62 @@ identifier: "Tasks"
...
+
+For more information on Care Guides, see the “[Decision Support for Care Guides]()” section of this overview. 
+
+When the user completes the care guide, the task will be cleared from the Tasks tab, and the report will be accessible from the Reports page or on the profile of the person or place whom the report is about.
+
+## Household Tasks
+    
+{{< figure src="tasks-household.png" class="right col-6 col-lg-3" >}}
+    
+Alternatively, there is an option to configure Household Tasks. When this [permission](https://docs.communityhealthtoolkit.org/apps/reference/app-settings/user-permissions/) is enabled, once a CHW has completed a task, they are taken to the Other Household Tasks page. This page shows the CHW all the additional outstanding tasks within the same household in which the initial task was completed.
```
andrablaj, line 54:
```suggestion
Alternatively, there is an option to configure Household Tasks. When this [permission]({{< relref "building/reference/app-settings/user-permissions" >}}) is enabled, once a CHW has completed a task, they are taken to the Other Household Tasks page. This page shows the CHW all the additional outstanding tasks within the same household in which the initial task was completed.
```

andrablaj (changes requested):
The images on the newly added page are broken. I made suggestions, but please double check that the suggestions fix the missing images.

content/en/building/tasks/_index.md
esthermmoturi: I'll fix this issue [here](https://github.com/medic/cht-docs/issues/1664)

content/en/building/tasks/_index.md
```diff
@@ -5,4 +5,62 @@ identifier: "Tasks"
...
 description: >
   Overview and configuration of CHT Tasks
+aliases:
+   - /building/features/tasks/
+   - /apps/features/tasks/
 ---
+
+Tasks help CHWs plan their day by prompting them to complete follow-up visits and other activities. The list might include upcoming scheduled ANC or Immunization visits, treatment or referral follow-ups, or other required activities such as a household survey.
+
+{{< figure src="tasks-mobile.png" link="tasks-mobile.png" class="left col-3 col-lg-3" >}}
```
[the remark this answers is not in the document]
esthermmoturi: Fixed in new commit

esthermmoturi:
@andrablaj , my localhost was caching changes so the images were not broken on my side. Thanks for noting that. I have moved the images, please let me know if they are still broken.

andrablaj:
@esthermmoturi, I have honest question before I submit my review: 
I noticed you moved the images to the same folder as the `.md` files. It looks like the docs site images are often kept in their own folder, separately from the `.md` files. I am OK with whatever option, but can we have the homogenous way of organizing the images across the docs site?

esthermmoturi:
@andrablaj , Images in hugo render when they are in the same folder as that of the .md page where they will be displayed. In our case, the images have to be in the same tasks folder as the _index.md page. [This](https://github.com/medic/cht-docs/tree/main/content/en/building/features/tasks) is how the folder I am moving looks like in the main branch. The situations where the images are placed in their own folders is when they don't appear in the index page. The only work around (If we want the images to be in a seperate folder in our case), is to have an empty index page and rename our current index page to something like tasks-overview.md then place the images in task-overview folder.

I hope I have provided clarification.

andrablaj:
Thanks for the clarification, @esthermmoturi! I looked at the `_index.md` file, and it contains a lot of documentation that could maybe better fit on a separate page, something like `Introduction`. The _index.html fils should be a more concise page about the section, and not a complete guide. What do you think?

esthermmoturi:
I agree @andrablaj , for now let me leave a blank index_page and rename the file to Introduction

PR #1665, opened 2024-10-29
fix(na): add detach to all docker calls

mrjones-plip:
# Description

add detach to all docker calls

This came up when testing for a [recent form post](https://forum.communityhealthtoolkit.org/t/error-while-installing-ssl-certificate-in-cht-4-0-0-with-docker/4240/5)

# License


mrjones-plip:
oops!  Sorry @andrablaj - I see now this is actually for running `certbot and you really want to see the interactive call to docker so you can see the errors on first run.  We don't want this!  As well, you don't want it detached in the cron job call.

ignore me :sweat_smile:

Issue #1680, opened 2024-11-04
Move condition cards section


esthermmoturi:
To be done on the week of 4/11/2024

PR #1689, opened 2024-11-06
Move condition cards section

esthermmoturi:
# Description

Closes https://github.com/medic/cht-docs/issues/1680

### DIFF
--- a/content/en/building/condition-cards.md
+++ b/content/en/building/condition-cards.md
@@ -1,7 +1,7 @@
 ---
 title: Condition Cards
 linkTitle: Condition Cards
-weight: 3
+weight: 7
 description: >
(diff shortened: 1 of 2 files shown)


content/en/building/condition-cards.md
```diff
@@ -1,13 +1,14 @@
 ---
-title: "Building Condition Cards"
+title: Condition Cards
 linkTitle: Condition Cards
-weight: 12
+weight: 3
```
andrablaj, line 4:
```suggestion
weight: 7
```

Just to make it different from the `Tasks` section. We can review all the weights we the migration is finalized

andrablaj (approved):
Non blocking suggestion. 👍

PR #1710, opened 2024-11-14
chore(#9578): add desire structure for the e2e tests files

tatilepizs:
# Description

After the clean-up work, we define the structure that the e2e test should follow, which needs to be updated in the documentation

[cht-core #9578](https://github.com/medic/cht-core/issues/9578)

# License


lorerod (approved):
Thanks!

Issue #509, opened 2021-06-01
Improve use of footnotes


abbyad:
Page: /apps/reference/app-settings/transitions/
Version: 2021-03-24 06:41:04  0200  0200
---

The footnotes are a relic from a previous document format, and are difficult to find/follow in the doc site:

![image](https://user-images.githubusercontent.com/789512/120338943-adc8e400-c2c2-11eb-9b61-e7a4f051432e.png)

When reading the documentation it isn't clear where to go to see the footnotes since they aren't cross-linked, and [7] is also separated out to the next header.

We don't use footnotes elsewhere in the doc site, so we should re-evaluate if footnotes are the best way to include this information, and if so (1) restructure the footnotes to make the content clearer, as well as (2) add best practices about them in the style guide.

esthermmoturi:
Closed by https://github.com/medic/cht-docs/pull/1714

PR #1714, opened 2024-11-18
Remove footnotes

esthermmoturi:
# Description

This PR seeks to close issue https://github.com/medic/cht-docs/issues/509.

The [transitions documentation](https://docs.communityhealthtoolkit.org/building/reference/app-settings/transitions/)
 contained footnotes that not only made it confusing to read, but were not standard with the rest of the documentation format. To achieve uniformity and improve readability. This PR includes changes that have:

- Removed superscripts that were used as a reference to the footnote information provided within the page
- Converted the footnote reference to a collapsible paragraph as the information is detailed and optional to the readers)(serves as additional information)
- Fixed some grammatical errors
- Changed some heading/subheading levels to make right-hand navigation easier and neater.

Future recommendation for this page:
- The page has information overload and should be broken down and subdivided into various pages.

### DIFF
--- a/content/en/building/reference/app-settings/transitions.md
+++ b/content/en/building/reference/app-settings/transitions.md
@@ -29,7 +29,7 @@ As of version 3.12.0 some transitions will partially run on the client for offli
 }
 ```
 
-In this example the `d` transition will not be applied, `a`, `b` `c` will be applied on the server and on the client, while `e` will only be applied on the server. 
+In the example above, the `d` transition will not be applied, `a`, `b` `c` will be applied on the server and on the client, while `e` will only be applied on the server. 
 
 ## Available transitions
 
@@ -57,11 +57,10 @@ The following transitions are available and executed in order.
 | [mark_for_outbound]({{% ref "building/reference/app-settings/outbound" %}}) | Enables outbound pushes. Available since 3.5.x |
 | [self_report](#self_report) | Maps patient to sender. Available since 3.9.x |
 | [create_user_for_contacts](#create_user_for_contacts) | Allows for automatically creating or replacing users based on data from their associated contact. Available since 4.1.x  | |
-## Transition Configuration Guide
 
-Guides for how to setup specific transitions.
+Below are guides to setup specific transitions.
 
-### multi_report_alerts
+## multi_report_alerts
 
 Send alert messages by SMS when specific conditions are received through reports. More flexible than simple Alerts.
 
@@ -87,7 +86,7 @@ Understanding the different types of reports used in the configuration:
 
 `0`, `1` : `new_reports` : `counted_reports` that came in since the previous alert was sent. They haven't been messaged about yet.
 
-#### Configuration
+##### Configuration
 
 ```
 "multi_report_alerts" : [
@@ -112,11 +111,11 @@ Note that we are using Mustache templates for our message templates (anything wi
 
 For performance reasons the `num_reports_threshold` cannot exceed 100.
 
-### death_reporting
+## death_reporting
 
 Updates patient documents with a `date_of_death` field which updates how the patient is displayed in the UI.
 
-#### Configuration
+##### Configuration
 
 Configuration is stored in the `death_reporting` field of the settings.
 
@@ -136,7 +135,7 @@ Configuration is stored in the `death_reporting` field of the settings.
 }
 ```
 
-### Registration
+## registration
 
 Configuration is held at `app_settings.registrations`, as a list of objects connecting forms to validations, events and messages.
 
@@ -307,22 +306,22 @@ To provide an alternative location for the place name, provide a `place_name_fie
 
 Sets the `case_id` on the root of the registration document.
 
-### Generate Shortcode on Contacts
+## generate_shortcode_on_contacts
 
-No custom configuration for `generate_shortcode_on_contacts`.
+There is no custom configuration for `generate_shortcode_on_contacts`.
 
-### Generate Patient ID On People
+## generate_patient_id_on_people
 
 **Deprecated since 3.8.x** in favor of `generate_shortcode_on_contacts`
 
-No custom configuration for `generate_patient_id_on_people`.
+There is no custom configuration for `generate_patient_id_on_people`.
 
 
-### update_notifications
+## update_notifications
 
 **Deprecated in favor of [Muting](#muting)**
 
-#### Configuration
+##### Configuration
 
 ```
 "notifications": {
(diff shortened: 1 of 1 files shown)


content/en/building/reference/app-settings/transitions.md
```diff
@@ -352,53 +351,57 @@ No custom configuration for `generate_patient_id_on_people`.
   }
 ```
 
-### Muting
+## muting
 
 Implements muting/unmuting of persons and places. Supports multiple forms for each action, for webapp and sms workflows.
 
-##### Muting action 
+### Muting action
```
jkuester, line 358: These docs are pretty tricky to begin with.  I am afraid that just nesting the former footnotes into a `details` block is going to make things worse since it further fragments the relationships between all the information here.  I have taken a stab at re-working these docs a bit to just de-normalize the info from the footnotes inline with the rest of the documentation.  I have included some cross-referencing via deep-links.  What do you think about something like this?

```md
### Muting action 

Updates the target contact and all its descendants, setting the `muted` property equal to the current date in ISO format.

#### Client-side

*Added in 3.12.0*

Client-side muting runs offline on a user's device. Only the contacts and reports available on the device will be updated.

- Sets the `muted` property on the target contact and all its descendants to the device's current `date` (the moment the report is processed). 
- Adds/updates the [`muting_history` property]({{< ref "#client-side-muting-history" >}}) on every updated contact, to keep track of all the updates that have been processed client-side, as well as the last known server-side state of the contact and sets the `last_update` property to `client_side`
- Updates the report doc to add a `client_side_transitions` property to track which transitions have run client-side

#### Server-side

Server-side muting runs as a typical Sentinel transition. Contacts that are already in the correct state are skipped. This applies to updates to the contact itself, updates to the Sentinel `muting_history` and to the connected registrations (registrations of a contact that is already in the correct state will not be updated).

- Sets the `muted` property on the target contact and all its descendants to the moment Sentinel processed the muting action.
  - If the contact was already muted by a client, the `muted` date will be overwritten. The [client-side `muting_history`]({{< ref "#client-side-muting-history" >}}) will have a copy of the client-side muting date.
- Adds a [`muting_history` entry]({{< ref "#server-side-muting-history" >}}) to Sentinel `info` docs for every updated contact
- Updates all connected registrations (for the target contact and descendants), changing the state of all unsent `scheduled_tasks` to `muted`
  - Unsent `scheduled_tasks` have either a `scheduled` or `pending` state
- Updates the contact's client-side `muting_history` to set the `last_update` property to `server_side` and update the `server_side` section with the current date and muted state. 
- If the report was processed client-side, all "following" muting/unmuting events that have affected the same contacts will be replayed. This means the transition _could_ end up running multiple times over the same report. 
  - Replaying is required due to how PouchDB <-> CouchDB synchronization does not respect the order in which the documents have been created, to ensure that contacts end up in the correct muted state.

### Unmuting action

Updates the target contact's topmost muted ancestor and all its descendants, removing the `muted` property. Because the muted state is inherited, unmuting cascades upwards to the highest level muted ancestor. If none of the ancestors is muted, unmuting cascades downwards only.

#### Client-side

*Added in 3.12.0*

Client-side unmuting runs offline on a user's device. Only the contacts and reports available on the device will be updated.

- Adds/updates the [`muting_history` property]({{< ref "#client-side-muting-history" >}}) on every updated contact, sets the last known server-side state of the contact and sets the `last_update` property to `client_side`
- Updates the report doc to add a `client_side_transitions` property to track which transitions have run client-side

#### Server-side

Server-side unmuting runs as a typical Sentinel transition. Contacts that are already in the correct state are skipped. This applies to updates to the contact itself, updates to the Sentinel `muting_history` and to the connected registrations (registrations of a contact that is already in the correct state will not be updated).

- Adds a [`muting_history` entry]({{< ref "#server-side-muting-history" >}}) to Sentinel `info` docs for every updated contact
- Updates all connected registrations (for the target contact and descendants), changing the state of all present/future `scheduled_tasks` (ones due today or in the future) with the `muted` state to have the `scheduled` state.
  - All `scheduled_tasks` with a due date in the past will remain unchanged.
- Updates the contact's [client-side `muting_history`]({{< ref "#client-side-muting-history" >}}) to set the `last_update` property to `server_side` and update the `server_side` section with the current date and muted state.
- If the report was processed client-side, all "following" muting/unmuting events that have affected the same contacts will be replayed. This means the transition _could_ end up running multiple times over the same report.
  - Replaying is required due to how PouchDB <-> CouchDB synchronization does not respect the order in which the documents have been created, to ensure that contacts end up in the correct muted state.

### Muting history

#### Server-side muting history

Each time the `muted` state of a contact changes, an entry is added to a `muting_history` list saved in Sentinel `info` docs (stored as an array property with the same name).
Entries in `muting_history` contain the following information:

| Property | Description |
| --- | --- |
| muted | Boolean representing the muted state |
| date | Date in ISO Format |
| report_id | An `_id` reference to the report that triggered the action |


#### Client-side muting history

*Added in 3.12.0*

Each time the client changes the `muted` state of a contact, an entry is added to a `muting_history` property on the contact's doc. The `last_update` entry is also changed to `client_side`.
The `muting_history` property contains the following information:

| Property | Values | Description | 
| --- | ---- | ---- |
| last_update | `server_side` or `client_side` | Updated every time a service updates the contact, with the corresponding value | 
| server_side | Object |  |
| server_side.muted | `true` or `false` | Last known server-side muting state |
| server_side.date | Date in ISO format | Last known server-side muting/unmuting date | 
| client_side | Array | Client-side muting/unmuting events list. <br> New events are pushed at the end of this list and it should never be re-ordered. <br>The list represents the "chronological" order in which the reports that triggered muting were created. | 
| client_side[].muted | `true` or `false` | Client-side muting state | 
| client_side[].date | Date in ISO format | Client-side muting/unmuting date |
| client_side[].report_id | uuid | The uuid  of the muting/unmuting report that triggered the update |
```

jkuester (changes requested):

content/en/building/reference/app-settings/transitions.md
```diff
@@ -87,7 +86,7 @@ Understanding the different types of reports used in the configuration:
 
 `0`, `1` : `new_reports` : `counted_reports` that came in since the previous alert was sent. They haven't been messaged about yet.
 
-#### Configuration
+##### Configuration
```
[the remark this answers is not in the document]
esthermmoturi: Thanks for pointing this out @jkuester . I was trying to avoid the headings appearing on the right-hand nav since they were too many and making it too long.
image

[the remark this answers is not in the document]
esthermmoturi: What do you think?

content/en/building/reference/app-settings/transitions.md
esthermmoturi: I like the rework you have done. So in that case we will get rid of the footnotes/collapsible block completely, right?

content/en/building/reference/app-settings/transitions.md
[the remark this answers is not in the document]
jkuester: :+1: I could not find any other way that Docsy lets you configure which headings show up in the sidebar.  I am totally fine leaving these as `h5` since we have an actual visual reason for doing it.

content/en/building/reference/app-settings/transitions.md
jkuester: Yes!  Basically all the info that was in the footnotes/collapsible block is now just inlined with the rest of the docs.

content/en/building/reference/app-settings/transitions.md
```diff
@@ -423,9 +435,7 @@ The `muting_history` property contains the following information:
 | client_side[].muted | `true` or `false` | Client-side muting state | 
 | client_side[].date | Date in ISO format | Client-side muting/unmuting date |
 | client_side[].report_id | uuid | The uuid  of the muting/unmuting report that triggered the update |
-
-
-#### Configuration
+##### Configuration
```
jkuester, line 438:
```suggestion

##### Configuration
```

jkuester (approved):
Excellent!  LTGM!

PR #1718, opened 2024-11-19
fix: correct docs for contact-summary field `appliesIf`

jkuester:
# Description

Currently the [contact-summary `fields` docs](https://docs.communityhealthtoolkit.org/building/contact-summary/contact-summary-templated/#contact-summarytemplatedjs-fields) indicate that the `appliesIf` property for a field could be:

> `function()` or `boolean`

However, checking the [actual code in cht-conf](https://github.com/medic/cht-conf/blame/da017f990c954950afb1235e90ca7ea6b2179d9a/src/contact-summary/contact-summary-emitter.js#L17), there is no way that a `boolean` value could be set here with a semantically logical meaning.  

```js
        if (!f.appliesIf || f.appliesIf()) {
```

If you set `appliesIf = false`, this if-statement would resolve `true` because `!false === true`.  If you set `appliesIf = true`, then we get the error `TypeError: f.appliesIf is not a function` because it will try to call `appliesIf` as a function when the property exists.

This PR simply updates the docs to clarify that `appliesIf` can _only_ be a function.

# License

### DIFF
--- a/content/en/building/contact-summary/contact-summary-templated.md
+++ b/content/en/building/contact-summary/contact-summary-templated.md
@@ -55,7 +55,7 @@ Each field that can be shown on a contact's profile is defined as an object in t
 | `width` | `integer` | The horizontal space for the field. Common values are 12 for full width, 6 for half width, or 3 for quarter width. Default 12. | no |
 | `translate` | `boolean` | Whether or not to translate the `value`. Defaults to false. | no |
 | `context` | `object` | When `translate: true` and `value` uses [translation variables](https://angular-translate.github.io/docs/#/guide/06_variable-replacement), this value should provide the translation variables. | no |
-| `appliesIf` | `function()` or `boolean` | Return true if the field should be shown. | no |
+| `appliesIf` | `function()` | Return `true` if the field should be shown, and `false` if it should be hidden. Default is `true`. | no |
 | `appliesToType` | `string[]` | Filters the contacts for which `appliesIf` will be evaluated. For example, `['person']` or `['clinic', 'health_center']`. It defaults to all types if it is not defined. | no |
 
 <!-- TODO: See [How to configure profile pages]() for an example.  -->


jkuester:
@kennsippell when you have a second, can you just confirm this passes your smell test? :pray: 

cc @tatilepizs

content/en/building/contact-summary/contact-summary-templated.md
```diff
@@ -55,7 +55,7 @@ Each field that can be shown on a contact's profile is defined as an object in t
 | `width` | `integer` | The horizontal space for the field. Common values are 12 for full width, 6 for half width, or 3 for quarter width. Default 12. | no |
 | `translate` | `boolean` | Whether or not to translate the `value`. Defaults to false. | no |
 | `context` | `object` | When `translate: true` and `value` uses [translation variables](https://angular-translate.github.io/docs/#/guide/06_variable-replacement), this value should provide the translation variables. | no |
-| `appliesIf` | `function()` or `boolean` | Return true if the field should be shown. | no |
+| `appliesIf` | `function()` | Return `true` if the field should be shown, and `false` if it should be hidden. Default is `true`. | no |
```
kennsippell, line 58: > or `boolean`

Can't you do `appliesIf: true`?

kennsippell: https://github.com/medic/cht-conf/blob/main/src/nools/task-emitter.js#L60

kennsippell, line 58: I dunno how much detail to go into.  Smells fine but slightly incomplete... it's a bit more complicated than saying "false if hidden" because visibility is determined by `appliesIf && !resolvedIf`. I think this is explained somewhat well [here](https://github.com/medic/cht-docs/blob/ab1bd5276f05ab7ee98cd09399a1513ecebb375b/content/en/building/tasks/complex-tasks.md?plain=1#L100).

jkuester: Thanks for having a look!

> visibility is determined by appliesIf && !resolvedIf

These docs are specifically for the contact-summary `fields`.  Unless I am totally missing something, I think that `resolvedIf` is just for _tasks_.  In the tasks world, yes, visibility (and everything else :sweat_smile:) is more complicated, but here for the contact summary fields, I think the logic is more straight forward.  

That being said, if I totally misunderstood what you meant, let me know and I can open a new PR to better clarify things!

PR #1722, opened 2024-11-22
Data migration and Migrating from 3.x to 4.x in Kubernetes

henokgetachew:
# Description
This guide answers:
- How do you use couchdb-migration in Kubernetes environments?
- How do you migrate a 3.x instance to 4.x without using docker-compose as an intermediary?

# License


content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
+---
+title: "Migration from CHT 3.x to CHT 4.x in Kubernetes"
+linkTitle: "K8s Data Migration to 4.x"
+weight: 1
+aliases:
+  - /apps/guides/hosting/4.x/k8s-data-migration
```
mrjones-plip, line 6: Since this is an entirely new page, we don't need any aliases

```suggestion
```

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+description: >
+  Guide to migrate existing data from CHT 3.x to CHT 4.x in Kubernetes environments
+relatedContent: >
+---
+
+The hosting architecture differs entirely between CHT-Core 3.x and CHT-Core 4.x. When both versions are running in Kubernetes, migrating data requires specific steps using the `couchdb-migration` tool. This tool interfaces with CouchDB to update shard maps and database metadata.
```
mrjones-plip, line 12: Can you add a link to the CouchDB Migration tool? That way folks can read up on it before using it.

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+
+{{% alert title="Note" %}}
+If after upgrading you get an error, `Cannot convert undefined or null to object` - please see [issue #8040](https://github.com/medic/cht-core/issues/8040) for a work around. This only affects CHT 4.0.0, 4.0.1, 4.1.0 and 4.1.1. It was fixed in CHT 4.2.0.
+{{% /alert %}}
+
+## Migration Steps
+
+### 1. Initial Setup
```
mrjones-plip, line 20: Please update all the numbers to use native markdown enumeration  like this:

```suggestion
1. Initial Setup
```

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+
+# Install npm (if not installed with Node.js)
+apt install -y npm
+
+# Install git and clone repository
+apt update
+apt install -y git
+git clone https://github.com/medic/couchdb-migration.git
+cd couchdb-migration
+npm ci --omit=dev
```
mrjones-plip, line 51: We can tidy this up a bit:

```suggestion
# Add node apt repository and update apt
curl -fsSL https://deb.nodesource.com/setup_16.x | bash -
apt update

# Ensure nodejs, npm and git are installed
apt install -y nodejs npm git

#  clone repository
git clone https://github.com/medic/couchdb-migration.git
cd couchdb-migration
npm ci --omit=dev
```

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+### 4. Run Pre-Migration Commands on 3.x
+
+Get credentials from 1Password and set them inside the pod:
```
mrjones-plip, line 56: Since we're in the container, we should be sure to remind them

```suggestion
While still `exec`ed in the `medic-os` container,  get credentials from 1Password and set them inside the pod:
```

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+```
+
+Save the output containing:
+- CouchDB secret (used for encrypting passwords and session tokens)
+- CouchDB server UUID (used for replication checkpointing)
+- CouchDB admin credentials
+
+### 5. Clone the 3.x Data Volume
+
+First, identify the volume ID from your 3.x data:
```
mrjones-plip, line 89: We should remind them that they need to exist the `medic-os` container before preceeding

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+```shell
+# Get the PVC name
+kubectl get pvc -n $NAMESPACE
+
+# Get the volume ID from the PVC
+VOLUME_ID=$(kubectl get pvc <your-3x-pvc-name> -n $NAMESPACE -o jsonpath='{.spec.volumeName}')
+EBS_VOLUME_ID=$(kubectl get pv $VOLUME_ID -o jsonpath='{.spec.awsElasticBlockStore.volumeID}' | cut -d'/' -f4)
```
mrjones-plip, line 96: I think because this section is so long, it's likely having them manually set  `<your-3x-pvc-name>` to be an env var.  let's  break this out into it's own step to find the the PVC name and set the three env vars. the next larger section can then just by copy and pasted blindly

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+    --resources $NEW_VOLUME_ID \
+    --tags Key=kubernetes.io/cluster/$CLUSTER_NAME,Value=owned Key=KubernetesCluster,Value=$CLUSTER_NAME
+
+# Verify your tags took effect
+aws ec2 describe-volumes --region eu-west-2 --volume-id $NEW_VOLUME_ID | jq '.Volumes[0].Tags'
+```
+
+### 6. Deploy CHT 4.x with Existing Data
+
+Create a values.yaml file using the volume ID from the previous step:
```
mrjones-plip, line 144:
```suggestion
Create a `values.yaml` file using the volume ID from the previous step:
```

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+
+```yaml
```
mrjones-plip, line 146: Two suggestions: let's use the updated template [here](https://github.com/medic/helm-charts/blob/main/charts/cht-chart-4x/values.yaml) and as well let's simplify it for single node.  We can then link to the larger yaml file and say, something like "use this larger one for multi-node"

combining these two suggestions will make this yaml file a lot less complex to edit and have have fewer changes need made to it

mrjones-plip (commented):
Nice work! We're in strong need of more k8s docs - yay!

I've made some suggestions - since this is endemic to Medic, we should move this to the [Medic section](https://docs.communityhealthtoolkit.org/contribute/medic/product-development-process/) so it's a peer of the existing [EKS docs](https://docs.communityhealthtoolkit.org/contribute/medic/product-development-process/deploy-on-eks/)

mrjones-plip (changes requested):
oops! prior feedback I submitted as "comment" as opposed to "request changes"

content/en/hosting/4.x/data-migration-k8s.md
dianabarsan: So we're running this code in the pod that hosts medic-os and medic-os is contantainerized within this pod or how? 
We would have node 8 installed for medic-os, does installing node 16 interfere in some way?

dianabarsan: Also, if you `npm link`, then you can use the `bin` commands directly.

content/en/hosting/4.x/data-migration-k8s.md
```diff
@@ -0,0 +1,282 @@
...
+# Set COUCH_URL
+export COUCH_URL="http://${ADMIN_USER}:${ADMIN_PASSWORD}@localhost:5984"
+
+# Verify connection
+curl -s $COUCH_URL/_up
+```
+
+Pre-index views to minimize downtime:
+```shell
+./bin/pre-index-views <desired CHT version>
```
dianabarsan: With `npm link` this would just be `pre-index-views 4.2.2`

dianabarsan (changes requested):
Such a shame that we can't spin up a container with the migration tool. Is there absolutely no way we can do that in k8s?

mrjones-plip (approved):
Nice work!  Thanks for all the changes.

While I told you to to use native numbers, we loose the linkable subheads and clickable TOC on the right - so I did the work of reverting the changes I requested - only fair!

As well, the YAML file has a LOT of changes that are needed, and unless we call them all out, literally having a list of what you need to change, it's too easy to miss one and then the deployment fails.  I listed out every item that needs to be changed - but feel free to update nomenclature if you feel it doesn't make sense.

approving, but please update the 4 sections that literally say `Henok what are dev and prod values` ;) 

thanks again!

dianabarsan (changes requested):
I think my only comment here is to add a partial for the sections that are common between these two PRs. More details here: https://github.com/medic/cht-docs/pull/1723#pullrequestreview-2458189515

dianabarsan (approved):
Cool then! Let's get this going!

PR #1725, opened 2024-12-03
fix(na): link to list of language codes instead of list of country codes

mrjones-plip:
# Description

link to list of language codes instead of list of country codes

# License


andrablaj (approved):
👍

PR #1729, opened 2024-12-10
feat(#1726): add 404 check script for new branches 

### LINKED ISSUE #1726: Add scripts/CI to check for `404` URLs after a big move
Hugo does a great job ensuring all pages that link to each other internally don't `404`.  However, for large moves [like we did recently with forms](https://github.com/medic/cht-docs/pull/1696), we may `404` a number of inbound links from other sources, or bookmarks folks have.  To ensure these don't break, it's nice to generate a list of all known URLs on `main`, do a big move, and then check that all the known URLs safely redirect.  

Two scripts [were written already](https://github.com/medic/cht-docs/issues/1351#issuecomment-2062693922) which we may choose to repurpose - but likely this should be:
- rewritten in node
- run in CI and block a merge if it fails
- allow users to run locally so they don't have to wait for CI

mrjones-plip:
# Description

Add script to check incoming branch for 404s 

closes #1726

# License

### DIFF
--- a/.github/scripts/check-for-404s.js
+++ b/.github/scripts/check-for-404s.js
@@ -1,0 +1,34 @@
+
+import Sitemapper from 'sitemapper';
+import http from 'http';
+
+
+// https://github.com/seantomburke/sitemapper
+const sitemapper = new Sitemapper();
+const testUrl = 'http://localhost:1313'
+sitemapper.timeout = 1000;
+
+sitemapper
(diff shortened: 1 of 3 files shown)


andrablaj:
@mrjones-plip I see the related ticket was closed. Is this PR still relevant?

mrjones-plip:
ah yes - we can close - thank you!

PR #1736, opened 2025-01-06
Update CHT Deploy docs to include node

mrjones-plip:
# Description

This PR:
* ensures both `node` >22 and `npm` are installed
* instructs users to run `npm install` so that dependencies are installed

# License


lorerod (approved):
I can't reproduce the steps in this guide because I [don't have proper access](https://github.com/medic/medic-infrastructure/issues/963#issuecomment-2575708028), but the change looks good to me. I trust the steps are working on your end.

PR #1744, opened 2025-01-16
fix(na): add link to bad shard forum post

mrjones-plip:
# Description

add link to bad shard [forum post](https://forum.communityhealthtoolkit.org/t/restoring-corrupted-couchdb-database/4551/1) on [migration page](https://docs.communityhealthtoolkit.org/hosting/4.x/migration/)

# License


andrablaj (approved):
As @1yuv is off until end of April, I suggest we merge this so it can benefit the community.

PR #1791, opened 2025-03-10
feat: add section about rtl support

dianabarsan:
# Description

Adds section about RTL support: 
image

Removes outdated section about setting user language via admin app. 

# License


andrablaj (approved):
🚢

PR #1795, opened 2025-03-20
feat: add docs for new contact and report REST endpoints

sugat009:
# Description

Add documentation for new REST endpoints for Contacts and Reports.

Solves: https://github.com/medic/cht-core/issues/9798

# License

### DIFF
--- a/content/en/building/reference/api.md
+++ b/content/en/building/reference/api.md
@@ -505,6 +505,254 @@ Endpoint used by cht-gateway to send sms messages. More documentation in the [ch
 
 Endpoint for integration with SMS aggregators. More details on the [RapidPro]({{< relref "building/messaging/gateways/rapidpro" >}}) and [Africa's Talking]({{< relref "building/messaging/gateways/africas-talking" >}}) pages.
 
+## Contact
+
+### GET/api/v1/contact/{{uuid}}
+
+*Added in 4.18.x*
+
+Returns a contact's data in JSON format.
+
+#### Permissions
+
+`can_view_contacts`
+
+#### Query parameters
+
+| Name         | Required | Description                                                                                        |
+|--------------|----------|----------------------------------------------------------------------------------------------------|
+| with_lineage | false    | If "true", the contact's parent lineage will be included in the returned data. Default is "false". |
+
+#### Examples
+
+Get a contact by uuid:
+
+```
+GET /api/v1/contact/d1c560e2-a07b-4562-a775-a2c7d0a9766f
+```
+
+```
+HTTP/1.1 200 OK
+Content-Type: application/json; charset=utf-8
+
+{
+    "_id": "d1c560e2-a07b-4562-a775-a2c7d0a9766f",
+    "_rev": "1-d13970de1501783cc52de0fd6db8beaf",
+    "type": "person",
+    "name": "Person Person",
+    "short_name": "August",
+    "date_of_birth": "1990-3-0",
+    "date_of_birth_method": "",
+    "ephemeral_dob": {
+        "dob_calendar": "1990-3-0",
+        "dob_method": "",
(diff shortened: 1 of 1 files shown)


content/en/building/reference/api.md
```diff
@@ -505,6 +505,254 @@ Endpoint used by cht-gateway to send sms messages. More documentation in the [ch
...
+
+#### Permissions
+
+`can_view_contacts`
+
+#### Query Parameters
+
+| Name     | Required | Description                                                                                                                                                                                                                                                                                                                       |
+|----------|----------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| type     | true     | ID of the contact_type for the contacts to fetch. _This field or freetext is required for this endpoint._                                                                                                                                                                                                                         |
```
jkuester, line 669: suggestion: I think instead of `This field or freetext is required for this endpoint.`, we could just add `true if freetext not provided` in the `Required` column. If I was a dev just trying to quickly scan the docs, I think that would be the easiest way for me to parse this.

Same goes for other similar places below.

content/en/building/reference/api.md
```diff
@@ -505,6 +505,254 @@ Endpoint used by cht-gateway to send sms messages. More documentation in the [ch
...
+| freetext | true     | A string to search the contacts by. The string value should be atleast 3 in length. The additional condition is that the value should not contain a whitespace(' ') unless the value is in the `key:value` pattern. The `key:value` pattern is for exact match searching. _This field or freetext is required for this endpoint._ |
```
jkuester, line 670: suggestion: I think we can follow the pattern from the cht-datasource docs and not mention the `key:value` stuff. (We support that for backwards compatibility, but I would rather not advertise that functionality.)  

> A string search term for filtering the contacts to fetch. The string value should be at least 3 characters in length and not contain whitespace.

Same goes for similar docs below.

content/en/building/reference/api.md
```diff
@@ -505,6 +505,254 @@ Endpoint used by cht-gateway to send sms messages. More documentation in the [ch
...
+| cursor   | false    | The token identifying which page to retrieve. A `null` value indicates the first page should be returned. Subsequent pages can be retrieved by providing the cursor returned with the previous page. Defaults to `null`.                                                                                                          |
+| limit    | false    | The total number of contacts to fetch. Defaults to `100`.                                                                                                                                                                                                                                                                         |
```
jkuester, line 672: issue:

> The total number of identifiers to fetch. Defaults to `10000`.

content/en/building/reference/api.md
```diff
@@ -1545,6 +1793,151 @@ Content-Type: application/json
...
+|----------|----------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| freetext | true     | A string to search the reports by. The string value should be atleast 3 in length. The additional condition is that the value should not contain a whitespace(' ') unless the value is in the `key:value` pattern. The `key:value` pattern is for exact match searching. |
+| cursor   | false    | The token identifying which page to retrieve. A `null` value indicates the first page should be returned. Subsequent pages can be retrieved by providing the cursor returned with the previous page. Defaults to `null`.                                                 |
+| limit    | false    | The total number of contacts to fetch. Defaults to `100`.                                                                                                                                                                                                                |
```
jkuester, line 1883: issue:

> The total number of identifiers to fetch. Defaults to `10000`.

jkuester (changes requested):
Great work Sugat!  Thanks go getting this turned around quickly!  Got a couple minor comments here and this will be good to go!

jkuester (approved):

Issue #1799, opened 2025-03-26
Some internal links are not working in 'CHT User Management tool'


binokaryg:
Page: /building/contact-management/user-management-tool/
Version: 2025-02-20 10:06:40  0000 UTC
---

Section: [Implementation Steps](https://docs.communityhealthtoolkit.org/building/contact-management/user-management-tool/#implementation-steps)

![Image](https://github.com/user-attachments/assets/1d724a26-7a5b-415e-83bc-771c434d584f)

These links to the section [ConfigProperty](https://docs.communityhealthtoolkit.org/building/contact-management/user-management-tool/#configproperty) are not working because of a wrong fragment identifier.

Expected: `#configproperty`
Used: `#ConfigProperty`

mrjones-plip:
@binokaryg - I believe #1800 fixed this - closing!

please reopen if I'm wrong

PR #1800, opened 2025-03-26
fix(#1799): internal links not working

binokaryg:
Fix the internal fragment/hash links

# Description

Fixing some internal links that are not working by:
- changing all uppercase letters in the identifiers to lowercase

# License


mrjones-plip (approved):
Just tested these anchors - fix is good - thanks @binokaryg !

Issue #1792, opened 2025-03-12
Update docs homepage: remove most of top half, focus on terse information to direct users


mrjones-plip:
Docs group of @antonykhaemba @andrablaj @lorerod @dianabarsan @mrjones-plip @paulpascal suggest simplifying the docs home page to focus on key personas arriving and looking to read documentation about a specific task. As such, this ticket will remove most of the top half of the docs page, greatly reduce most of the text for each persona (app builder etc), something like below.  Opening this ticket for cited folks above to comment on how best to proceed. 

![Image](https://github.com/user-attachments/assets/65812410-94ca-4437-9ee9-f91f603b1f08)

PR #1802, opened 2025-03-26
fix(#1792): Update homepage 

mrjones-plip:
# Description

Remove top half of homepage to make it more lean per docs squad and per #1792

# License

### DIFF
--- a/content/en/_index.md
+++ b/content/en/_index.md
@@ -14,7 +14,7 @@ cascade:
   type: "docs"
 ---
 
-## CHT Core Framework
+## Core Development
 An overview and reference for development of the Core Framework of the Community Health Toolkit (CHT). Most CHT App developers are able to build great experiences for their users without extending CHT Core, but you might still find this section useful for its [overview of CHT components]({{< ref "core/overview" >}}).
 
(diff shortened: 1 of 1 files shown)


lorerod (changes requested):
Thank you, @mrjones-plip, for this! I went ahead and suggested updating the titles. I’m not sure if this change fully aligns with the scope of this PR, and since wording can be a matter of preference, feel free to discuss any suggestions with me. Also, I’m experimenting with the table at the beginning—not sure if it’s the best approach, but I wanted to try it out. Let me know what you think!

mrjones-plip:
@lorerod - thanks for the thoughtful review!  While I think your changes are improvements, I've resolved all of them as I don't think we should do them - yet!!!  For now, lets keep party of the left hand nave with the body of the text.  As we finalize the left hand nave side, we can update the home page topics to match (eg we don't plan on having an "Operations" and may actually keep "Designing" (vs suggested "Design")


Here's current left hand nav for reference: 

```
CHT Core Framework
Exploring
Designing
Building
Hosting
Contributing
```

lorerod (approved):
Fair enough! Lgtm thanks!

PR #1825, opened 2025-04-10
chore: add to glossary and change format

andrablaj:
# Description

Add items to the glossary. Change the style of entries to increase readability.

# License

### DIFF
--- a/content/en/glossary.md
+++ b/content/en/glossary.md
@@ -6,338 +6,345 @@ weight: 10
 
 ## A
 
-###### Accredited Social Health Activist (ASHA)
+#### Accredited Social Health Activist (ASHA)
 
-###### Antenatal Care (ANC)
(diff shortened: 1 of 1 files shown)


dianabarsan (approved):
I am honestly surprised we have this and choose to keep this.

andrablaj:
@dianabarsan, would you suggest to remove this page completely?

dianabarsan:
> @dianabarsan, would you suggest to remove this page completely?

Uh ... I think so. I think it's seriously all over the place. You might think it's an acronym glossary, but then it has a random definition for `tool`, of all things. 

Checking analytics, this page had 8 views from 5 active users in the whole of latest year, with the last three of them yesterday :D 

Yes, I suggest we remove it.

andrablaj:
[Done](https://github.com/medic/cht-docs/pull/1827)

PR #1826, opened 2025-04-10
chore: remove unnecessary images

andrablaj:
# Description

Opinionated PR:

The two images on the "Why the CHT?" page are of little value and look awkward. The localization one is not up to date. I think we should remove them. 

Screenshot 2025-04-10 at 18 23 24
Screenshot 2025-04-10 at 18 23 19

# License


mrjones-plip (approved):
less is more \o/

Issue #1838, opened 2025-04-15
Update CHT local development instructions with technical prerequisites


andrablaj:
Per [this forum conversation](https://forum.communityhealthtoolkit.org/t/trouble-with-build-dev-during-development-setup/4845/6), one needs a machine with at least 8GB of RAM to be able to run the CHT locally.

Add these details to the [CHT local development setup documentation](https://docs.communityhealthtoolkit.org/contribute/code/core/dev-environment/).

swet40:
@andrablaj Can you assign this issue to me, I have started to work on this

andrablaj:
Thanks for volunteering to contribute @swet40! Please check out the [contributing documentation instructions and style](https://docs.communityhealthtoolkit.org/contribute/docs/), and reach out on the [CHT Forum](https://forum.communityhealthtoolkit.org/) if you have any questions. Good luck!

swet40:
@andrablaj I have made the PR. Can you check

Sub-codes:
Is this issue still open?

mrjones-plip:
oh no! I'm sorry @Sub-codes - I missed your comment and whipped a one line fix for this issue.  I'll do better next time!

Sub-codes:
No issues @mrjones-plip 👍

PR #1839, opened 2025-04-15
Add WSL2 recommended configuration to local setup documentation

swet40:
fixes: #1838 

## Description

This PR adds a note under the "Setup environment" section of the local-setup.md guide to recommend allocating at least 8GB of RAM for CHT development using WSL2. It includes an example .wslconfig block based on internal recommendations.

## Result Produced Locally -
Screenshot 2025-04-15 202952


content/en/building/local-setup.md
```diff
@@ -41,6 +41,18 @@ The *CHT Core Framework* makes it faster to build full-featured, scalable digita
...
+> Add the following to your `.wslconfig` file (typically located at `C:\Users\<YourUser>\.wslconfig`):
+>
+> ```
+> [wsl2]
+> memory=8GB
+> processors=10
+> swap=512MB
+> localhostForwarding=true
+> ```
+
```
mrjones-plip, line 55: We can clean this up a bit to be more concise. As well, `processors`, `localhostForwarding` and `swap` already have sane [defaults](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#main-wsl-settings)  so I think leaving them undefined is best and least invasive.

Finally, we can normalize the format to be in line with the rest of our docs:

``````suggestion
> **WSL2 configuration**:  
For Windows users,  configure WSL2 to use more memory so the CHT has enough RAM by adding the below values to `.wslconfig` file. This is typically located at `C:\Users\<YourUser>\.wslconfig`:

```
[wsl2]
memory=8GB
```

See the [WSL Settings docs](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#main-wsl-settings) for more information.
``````

mrjones-plip (changes requested):
Thanks so much for the helpful change @swet40 ! This is a helpful change that make the lives of Windows developers easier!

While I did make a suggestion - when researching this ticket I think maybe that this PR might not be the way to go.  Let's follow up [with the forums first](https://forum.communityhealthtoolkit.org/t/trouble-with-build-dev-during-development-setup/4845/9?u=mrjones) before we make any changes.

Stay tuned!

content/en/building/local-setup.md
swet40: Thanks for the feedback! I’ll clean up the config to remove the redundant fields and match the formatting style used across the docs.

mrjones-plip:
As a reviewer, I try to make it as easy as possible!  If you agree with all my suggestions, you can just click the "Commit suggestion" button which fixes all the concerns I raised: 

image

mrjones-plip (changes requested):
hey @swet40 !

After not getting any updates from the [forum thread](https://forum.communityhealthtoolkit.org/t/trouble-with-build-dev-during-development-setup/4845/9?u=mrjones) and thinking on this some more, let's just make this a PR to increase the [App Dev requirements](https://docs.communityhealthtoolkit.org/hosting/requirements/#app-developer-hosting) to 8gb from 4gb.  If you'd like, you can add a note in the [local setup](https://docs.communityhealthtoolkit.org/building/local-setup/) guide to link users to the requirements - just a few words with the link will do!

separately, you've committed changes from another PR into this branch.  Please revert those. 

thanks!

andrablaj:
@swet40, please review and apply @mrjones-plip suggestions, and reach out if you encounter any blocker. Thank you!

mrjones-plip:
Thanks for your efforts on this @swet40 !  We're going to go with [another PR](https://github.com/medic/cht-docs/pull/1876) that fixes this problem.

PR #1862, opened 2025-05-01
fix(na): remove legacy deps

mrjones-plip:
# Description

 remove legacy deps per @jkuester 

# License


jkuester (approved):
:grapes:

PR #1894, opened 2025-05-20
feat: add page for cht calendar

andrablaj:
# Description

Add a page for the CHT Event Calendar. 

# License


andrablaj:
@Phil-Mwago per our chat today, I added a page on the docs side with the CHT Event Calendar so that we can reference it in our comms. 

@mrjones-plip, I would also love your feedback on this PR regarding the Hextra/code sanity perspective. 

Thank you both!

mrjones-plip (approved):
Overall, having the calendar in docs makes sense! Approving.

Given we only have one instance of the `iframe` for the calendar the responsive iframe template seems a little overkill (just keep it as inline HTML) - but don't feel strongly abou this at a all!

Issue #1870, opened 2025-05-06
Update Contact Management section images


lorerod:
Update the images in [Contact Management](https://docs.communityhealthtoolkit.org/building/contact-management/) section to use the new images generated with the visual e2e tests.

PR #1898, opened 2025-05-21
chore(#1870): update Contact Management section images to latest UI

lorerod:
# Description

This PR updates the images in the Contact Management section to reflect the latest UI, using new screenshots generated with the visual e2e tests.

closes https://github.com/medic/cht-docs/issues/1870

# License


content/en/building/contact-management/contact-and-users-1.md
```diff
@@ -102,8 +105,7 @@ Once you submit, a new CHW Area will be created. On the right-hand you should se
 
 {{< figure src="new-chw-supervisor/new-person.png" link="new-chw-supervisor/new-person.png" class="right col-6 col-lg-8" >}}
 
-To create a primary contact for an existing Place (in this case, for the Health Facility that we created without a primary contact); select the Place and then select the **New Person** action.
-
+To create a primary contact for an existing Place (in this case, for the Health Facility that we created without a primary contact); select the Place and then select the Floating Action Button (FAB) that opens a menu with all actions, in the menu select the **New Person** action.
 
 {{< figure src="new-chw-supervisor/belongs-to.png" link="new-chw-supervisor/belongs-to.png" class="right col-6 col-lg-8" >}}
```
mrjones-plip, line 110: The new screenshot (left) is clicking the "Full name" vs the old screenshot (right) is clicking "Belongs to".   I believe the new screenshot is wrong and the old is correct:
image

content/en/building/contact-management/contact-and-users-1/new-chw-user/fill-user-details.png
mrjones-plip, line 1: the old screenshot has the entire new user form which I think is helpful to see all the fields.  

I'm not sure we need to show clicking on the Place as you'll need to click on all the fields to edit them

content/en/building/contact-management/contacts.md
```diff
@@ -23,19 +23,16 @@ aliases:
 
 Users can access their “people” and “places” from the **People** tab. The permissions set for your role and your placement in the hierarchy will determine which contacts you’re able to see. Advanced configuration options are available for a specific offline user role to manage what [level of contact data]({{< ref "building/guides/performance/replication#contact-depth" >}}) is downloaded and stored on their device.
 
-{{< cards rows="2" >}}
-{{< card link="people-mobile.png" image="people-mobile.png"  method="resize">}}
-{{< card link="people-desktop.png" image="people-desktop.png"  method="resize">}}
-{{< /cards >}}
+
+{{< figure src="people-desktop.png" link="people-desktop.png" class="right col-6 col-lg-8" >}}
```
mrjones-plip, line 27: I think it would be nice to have feature parity where we show the mobile and desktop UIs.  now we just have the desktop.

content/en/building/contact-management/contacts.md
```diff
@@ -81,9 +78,9 @@ From profiles, users can edit contact information, take actions, and, if viewing
 
 The top card on all profiles contains general information for the contact. All the fields shown in this summary card are configurable.
```
mrjones-plip, line 79: these larger contact summary screenshots are an improvement over the old ones - nice!

mrjones-plip (changes requested):
This is AMAZING to see this all being automated - long term this is huge and easy for us to keep up to date.

Is there a way we can add borders?  For example the left is the new and on the right is the old.  note on the right you can tell where the screenshot ends because there's  small border.  I think when there's white on white screens you may even get confused what the on page content vs in screenshot content is:

image

content/en/building/contact-management/contact-and-users-1.md
lorerod: Good catch! I moved the taping hand to the `Belongs To` field.

content/en/building/contact-management/contact-and-users-1/new-chw-user/fill-user-details.png
lorerod: Thanks for the feedback! You're right that clicking on “Place” specifically might not be necessary if we're not highlighting something unique about it — I updated the screenshot to keep the focus on the overall editing process instead.

That said, I’m not sure we’ll be able to show the entire new user form in the screenshot. We updated the screen size used for desktop screenshots ([reference](https://github.com/medic/cht-core/blob/9c37401e9bd949654a7b3b7b7c0b5e28594b0d37/tests/e2e/visual/wdio.conf.js#L24C8-L24C30)), and this is what’s currently visible at that resolution.

Let me know what you think.

content/en/building/contact-management/contacts.md
lorerod: I imagine admin users primarily use desktop for configuring users, so the current setup reflects that. However, I can update the [wdio.conf.js setup](https://github.com/medic/cht-core/blob/9c37401e9bd949654a7b3b7b7c0b5e28594b0d37/tests/e2e/visual/wdio.conf.js#L33C1-L34C1) to run the `contacts/contact-user-hierarchy-creation.wdio-spec.js` test at a mobile resolution, which would generate mobile screenshots. Let me know if you feel strongly about including those.

One thing to clarify: the yellow tap finger in the screenshots isn’t automatically generated—I have to add it manually. That made me wonder: Should we skip the tap finger so the documentation images are easier to maintain? On the other hand, the icon helps illustrate the flow. What do you think?

mrjones-plip:
sorry for the delay @lorerod ! I'll get to this in the next day or so.

mrjones-plip (approved):
thanks for updates - looks good!

PR #1920, opened 2025-06-05
fix(na): oops - we need the medic-infro repo cloned ;)

mrjones-plip:
# Description

oops - we need the medic-infro repo cloned ;) 

reverts #1914

# License


mrjones-plip:
@Hareet - ping!

PR #1924, opened 2025-06-12
feat(#9974): add docs for opening contact form from task

jkuester:
# Description

https://github.com/medic/cht-core/issues/9974
https://github.com/medic/cht-core/pull/9975

# License

### DIFF
--- a/content/en/building/tasks/managing-tasks/launch-contact-form.md
+++ b/content/en/building/tasks/managing-tasks/launch-contact-form.md
@@ -1,0 +1,97 @@
+---
+title: "Launching a contact form from a task"
+linkTitle: Launching a contact form from a task
+description: >
+  Demonstrates how to create or edit a contact from a task 
+relatedContent: >
+  building/tasks/complex-tasks
+  building/tasks/tasks-js
+  building/tasks/tasks-js/#tasksjs
+---
+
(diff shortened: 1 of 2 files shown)


content/en/building/tasks/managing-tasks/launch-contact-form.md
```diff
@@ -0,0 +1,97 @@
...
+      content.edit_id = contact._id;
+    },
+  }],
+}
+```
+
+## Navigate to contact profile
+
+{{< callout >}}
+Added in CHT `4.21.0`.
```
jkuester, line 94: So, interestingly, the docs over in `tasks-js.md` have always (since the page was written in 2020) said this about the action `type`:

> When `'contact'`, the action redirects to a contact's profile page

_However,_ as far as I can tell from the code in webapp, this have never been the case :confused:.  I did some archeology and I think this feature originated with https://github.com/medic/cht-core/issues/1505 and https://github.com/medic/cht-core/tree/f196d53ea9669b2f73329ead048af29430667853.  But, this just supports popping a contact create form, not navigating to the contact's profile page.  

Regardless, @ChinHairSaintClair and I felt that navigating to the profile page was a reasonable default behavior (previously, you would just end up at `contacts/add` when a `type` was not set and the user just sees:

image

dianabarsan (approved):
Crystal clear I understood eeeeeeverything!

PR #1928, opened 2025-06-26
chore: mark 3.x guides as archive

andrablaj:
# Description

Mark 3.x hosting guides as archive and EOL. 

# License

### DIFF
--- a/content/en/hosting/3.x/_index.md
+++ b/content/en/hosting/3.x/_index.md
@@ -7,6 +7,10 @@ aliases:
     - /apps/guides/hosting/3.x
 ---
 
+{{< callout type="warning" >}}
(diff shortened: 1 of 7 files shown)


mrjones-plip (approved):
Hey @andrablaj - this PR is a great addition to clearly call out 3.x is EOL. Let's ship it!

That said, I would still like to suggest deleting the 3.x wholesale.  We'll have [an archive](https://old-docs.dev.medicmobile.org/hosting/3.x/) if folks are still using it, but I think there's potentially a lot of clarity, simplicity and tech debt reduction by deleting ALL the the 3.x content (there's a lot of nooks and crannies of 3.x!!)

PR #1931, opened 2025-07-03
chore: reduce the size of supported versions table

andrablaj:
# Description

Reduce the size of the unsupported CHT versions table to improve readability. The release dates for older versions can be found in the respective release notes, if needed. 

# License


andrablaj:
Agreed. I removed the `earliers`. Thanks for the suggestion!

PR #1957, opened 2025-08-12
chore(#1936): shuffle build quick guides part 1

dianabarsan:
# Description

[description]

# License


mrjones-plip (approved):
We should be sure to fill out the description for our future selves or for other collaborators so folks know what we're doing when.

however, not at all a blocker for this specific effort :ship:

mrjones-plip:
Also - confirmed redirects are clean:

```

=== SUMMARY ===
Total URLs checked: 465
Redirects (200): 14
Other errors: 0
404 Not Found: 0
```

PR #1959, opened 2025-08-12
chore(na): removing line breaks so tickets look good

mrjones-plip:
# Description

Tickets created from the new template on good first issue page don't like the line breaks.  removing them!

# License


lorerod (approved):
Yes! Thanks!

PR #1963, opened 2025-08-18
Define & add contribution section  

Phil-Mwago:


andrablaj (commented):
@Phil-Mwago I suggest you move the text you created in a separate "What is a contribution page?" for easier readability and navigation. That page should be the first in the hierarchy. Thank you!

mrjones-plip (approved):
sorry for delaying my feedback!

I think this is good to go as is or would be OK as well with Andra's feedback to move it.

Phil-Mwago:
Closing this... opened #1975

Issue #2006, opened 2025-09-26
Search functionality has changed after upgrading Hextra


mrjones-plip:
Something changed on search where it mostly doesn't work.  For example it returns `0` results for `test` which it should return dozens.  As well, `medic` returns only 3 results when it too should return dozens.

here's a video of production site vs local branch [running a branch](https://github.com/medic/cht-docs/pull/1986) that was behind `main`:

https://github.com/user-attachments/assets/573651e6-3b70-474e-95a7-a5bcb698009b

andrablaj:
I can confirm the change was introduced with the [Hextra version 0.10](https://imfing.github.io/hextra/blog/v0.10/#enhanced-search-experience).

I downgraded Hextra to `0.9.7` and the search behaves as it used to (you will notice the rest of the site is broken due to the changes performed during the uplift): 

https://github.com/user-attachments/assets/e15758e3-3120-4d2c-bc70-8a6879b4f158

I would not say the search is broken, but it gives answers to more detailed prompts: for example, if you search for `Testing Forms`, `Testing the CHT`, or `Testing Contact Summary`, or other similar prompts, you would get results. I personally prefer this more focused result experience over the one before, which returned every page that mentioned the word `Testing`. 

There are [search configurations](https://imfing.github.io/hextra/docs/guide/configuration/#flexsearch-index) that we could explore to customize the experience further.

mrjones-plip:
I've [reached out on the forum](https://forum.communityhealthtoolkit.org/t/search-on-docs-site-do-you-use-it-does-it-work-for-you/5386/1) to see what the larger community thinks.

If we get no feedback (or ambiguous feedback) all things being, I think we should just keep it as is and close this ticket.

1yuv:
Searching on the doc site has still been a problem for me. I can't find what I am searching for since this new feature is in the production.

binokaryg:
I tried searching for "SMS recipient resolution" today, and didn't get the correct suggestion until I typed: `SMS recipient resol`.

andrablaj:
We recently changed the search functionality to use the Kapa.ai search https://github.com/medic/cht-docs/issues/2102, so we can close this issue.

PR #2012, opened 2025-10-02
chore(#2006): change flexsearch tokenize to full

andrablaj:
# Description

Change `flexsearch` `tokenize` parameter to `full` to increase the number of results returned by the search.

Per [documentation](https://github.com/nextapps-de/flexsearch#tokenizer-partial-match) + [additional documentation](https://lotusdocs.dev/docs/features/flexsearch/): 

- `strict` - Only exact matches for the search term will be shown in the results.
- `forward` (default) - Words are searched for starting from the first letter in a forward order, allowing partial matches to be displayed; unfinished words or words with mispellings at the end will still display results e.g. if searching for ‘Testing’, entering only ‘Tes’ will provide some results.
- `reverse` - A result will be shown when the letters of the search term occur in both forward or reverse order, provide even more support for mispellings or incomplete searches e.g. if searching for ‘Testing’, entering either ‘Tes’ or ‘ing’ will provide some results.
- `full` - Combines all of the above, plus partial matches for the middle section of the search term e.g. if searching for ‘Testing’, entering either ‘Tes’, ‘ing’ or ‘sti’, or other combinations thereof will provide some results.

#2006 

# License

### DIFF
--- a/hugo.toml
+++ b/hugo.toml
@@ -184,7 +184,7 @@ osEnv = ['(?i)^((HTTPS?|NO)_PROXY|PATH(EXT)?|APPDATA|TE?MP|TERM|HOME|ASDF_DIR)$'
 
     [params.search.flexsearch]
     index = "content"
-    tokenize = "forward"
+    tokenize = "full"
 
   [params.editURL]
   enable = true


jkuester (approved):
🍇

PR #2015, opened 2025-10-02
fix: correct docs to say targets only enabled for offline users

jkuester:
# Description

The current documentation for aggregate targets says they "can be configured for both online and offline users".

However, I don't think this is true.  In the targets component we [check if the rules engine service is enabled](https://github.com/medic/cht-core/blob/1098e344a8d71ae5dd5570b17bb34312689e0f87/webapp/src/ts/modules/analytics/analytics-targets.component.ts#L95) (and otherwise show a message saying 

image

If you have an online user, the [rules engine service always says it is disabled](https://github.com/medic/cht-core/blob/da5fe7f731303c6f69315b3992455125f539f91a/webapp/src/ts/services/rules-engine.service.ts#L101).

Am I missing something here? :thinking: This same should be true for aggregate targets, right?  Cannot see aggregate targets if you cannot see any targets....

# License


dianabarsan (commented):
I believe aggregate targets _are_ available for online users still. What makes you believe they are not?

dianabarsan:
Ok, so something is wrong with target aggregates for online users. It seems that the page tries to load targets from 3 months ago, and when it fails, it goes to a stop. I think this is related to recent changes. 
Anyway, target aggregates are available for online users as well.

PR #1932, opened 2025-07-05
fix(#1790): replace relref links with standard markdown linksFix reference iss1790

ITSUREN:
# Description

This PR addresses issue #1790 by replacing all Hugo relref shortcodes in markdown files with standard markdown links.  
I added a helper script (`tools/fix-relref.py`) to automate this process for future use, and used it to update relevant files with the '.md' extension.

Let me know if you'd like me to split the script into a separate PR or keep it as-is.

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #2017, opened 2025-10-03
chore(#1790): replace relref with link render hooks

### LINKED ISSUE #1790: Migrate from `relref` shortcode to use the new Hugo link render hooks
Currently we use the [relref shortcode](https://gohugo.io/shortcodes/relref/) to make internal links. Among other things, this will  warn us when these links break (e.g. because we moved a page or something).  

Andra [noticed](https://github.com/medic/cht-docs/pull/1789) that this shortcode is not superseded by the new [link render hooks](https://gohugo.io/render-hooks/links/) functionality.  Seems to perform a similar function, but with less syntax.  

## Good first issue

Welcome first time contributor!  Please read through all of these notes before beginning 🙏 

This ticket is going to involve a bunch of searching and replacing - [regular expressions](https://en.wikipedia.org/wiki/Regular_expression) might come in handy!  The goal will to replace all instances of `rel` and `relref` with the newer style of links: 

* `[Example and Reference Apps]({{< ref "reference-apps" >}})` will be converted into `[Example and Reference Apps](/reference-apps)`
* `[replication]({{< relref "building/guides/performance/replication" >}})` will be converted into `[replication](/building/guides/performance/replication)`

Some important notes:
* this PR will effect a very large percent of the documents in the repository - move carefully!  
* consider writing a script in a language you're comfortable in.  This will allow you to run the script, test the results, revert all the changes if you failed and re-run the script.  
* expect something to break! if you submit the PR and you have a bunch of pages that are broken, that's OK!  This is part of the process.  again, having a script run that will allow you to easily re-run the process will be helpful
* be sure you set up hugo locally. this will allow hugo to compile the entire site and check for errors
* likely [docker](https://github.com/medic/cht-docs?tab=readme-ov-file#docker) is the easiest way to run hugo 
* any link that currently ends in `.md` will need to have that removed.  For example the release pages all link targets are `releases/4_18_0.md` and they need to be `releases/4_18_0`
* all links will need to have  leading `/` in them. For example, again on the release pages, have no leading slash like this `releases/4_18_0.md` and need to have it added like this `/releases/4_18_0` 
* The new [link render hooks](https://gohugo.io/render-hooks/links/) support a title.  You don't need to add these.  For example,  "My first post" is the title  in this example: `[Post 1](/posts/post-1 "My first post")`  and none of our existing links have titles so you don't need to add them. 

While questions about the ticket are here on the ticket and  more general questions are best asked on the [forum](https://forum.communityhealthtoolkit.org/), don't hesitate to ask wherever you like!  That said, when you do ask a question, be sure to list explicit steps on how to reproduce your issue.

andrablaj:
# Description

Closes #1790 

Builds on the excellent work started in https://github.com/medic/cht-docs/pull/1932, with minor link fixes. I didn't commit the script that was created in that PR, as it will be unnecessary moving forward.  

# License


mrjones-plip (commented):
🪓  Epic! nice work

PR #2101, opened 2025-12-19
feat: add interactive cost calculator to the hosting costs page

jkuester:
# Description

This is a little present for @mrjones-plip when he returns from the holidays! :christmas_tree: 

image

It is a first draft of an interactive Hosting Cost estimation tool ([forum thread](https://forum.communityhealthtoolkit.org/t/hosting-cost-estimation-improvement-suggestions/4694)).

It turns out it is pretty easy to add interactive HTML to your Hugo site via a custom shortcode.  And, the modern CSS functionality provide very nice layout and visualization so I did not have to resort to any dependencies.  

I am pretty happy with the features/layout/functionality of the whole thing, but the _tuning_ of it could use more eyes.  

One simplification that I have started with (but which could be easily refactored) is that I have only factored in the costs [of 5 example EC2 instances](https://github.com/medic/cht-docs/pull/2101/changes#diff-3c24e5eaba03526c09365be6c7c4c9e967f2fbd8e08ff59da4a1f69302fc597fR3) and I try to pick one of those based on load (users * workflows).  This approach has obvious limitations, but it was simple enough to let me get started with real world numbers.  Definitely open to suggestions on the best way to improve this.

The other things that need more tuning are the various constant value that are used in down-stream computations:

```js
  DISK_COST_PER_GB: 1,
  CONTACTS_PER_PLACE: 5,
  WORKFLOW_YEARLY_DOCS_PER_CONTACT: 12,
  DOCS_PER_GB: 12000,
```

`DISK_COST_PER_GB` is a rounded-up estimate from EBS. `CONTACTS_PER_PLACE` is roughly the "household size", but it gets used to estimate how many ancestor contacts are in the hierarchy tree (given the user input of the population size).  Currently the logic assumes a completely even distribution and density of contacts throughout the tree.  `WORKFLOW_YEARLY_DOCS_PER_CONTACT` is by far the most hand-wavy. Basically I need some way to connect how many workflows are being supported by the instance with an estimate of how many reports will be generated per year.  So, in this case `12` means that I think we will have an average of 1 report created per month per workflow per contact. This may be way off.  `DOCS_PER_GB` is a rough estimate that I made based on Watchdog data from a large production instance. 

Most of this code was written (or heavily influenced) by Claude, but I have carefully reviewed and edited it for maximum maintainability.  

# License

### DIFF
--- a/assets/css/cost-calculator.css
+++ b/assets/css/cost-calculator.css
@@ -93,6 +93,7 @@ html.dark {
     font-size: 1.5rem;
     font-weight: bold;
     color: var(--calc-link);
+    white-space: nowrap;
   }
   .inline-cost-monthly {
     font-size: 0.75rem;


mrjones-plip (changes requested):
OMG This is amazing - I love it!  Like, I loooooove it!  What a great gift to return to after the holidays - you were right!

Overall this is great and we should most definitely ship it! Before we ship it, I have some comments I made as I was playing with it.  As with all GUIs, they bring about a opinions, bugs and corner cases. Feel free to push back where ever!  Let's work through at least few iterations focusing on one set of feedback at a time.  Possibly maybe starting with feedback from Loukman, Simon, Nitin and Derick (Jeff too?) who face the cost question so often in the field.  Seems a shame to spend a bunch of time gussying up the GUI and logic only to have a new item added or a key component removed.  I'm open about this approach though! As well, we'll need to ground truth the outputs.  Happy to chat on this tomorrow/Fri as needed! No code review just yet - that'll come last I think!

First, what's to love about this?  So much!
- It's easy to use - 4 sliders covers 90% of user cases 
- It's intuitive - population, workflows, age and users are accessible concepts.  They also prompt healthy discussion about how to plan for a deployment in a mostly non-technical way ("what's a workflow? how do I know how many I need?")
- Because it can answer such a wide range of questions with all the sliders, it really addresses the ongoing questions of "I still don't know how much the CHT costs to run!!!" - this is a lovely answer!
- It's just really pretty to look at it 😻 
- Being instant is a real dopamine hit.  You instantly see the changes and it's a real joy to use.
- I'm inspired for v2, v3! Maybe adding some pre-filled examples to get people thinking? Like, "this shows the cost of an East African country in it's 6th year of operation.  There were 100k people and 5 workflows. Note how the disk becomes the predominant cost at over 80%" vs something like "This small deployment serving a population of 1k with 100 CHWs has an EC2 cost of almost 99% and is well under $1k/year"
- From a technical perspective, this is readable code, easy enough to maintain and not megs and megs of some superfluous JS framework - Raw JS FTW! \o/


GUI suggestions:
- add a reset to advanced parameters. I was playing with it and got way outside of normal parameters and had to hard reload to page to reset them all
- don't show cents on cost estimates & instance (maybe OK on cost per user?). Round up to the dollar.
- don't show decimals in GB (Storage Breakdown) or on people per user
- make sure "Cost Per User" dollar amount is on one line (no break between "$0.46/" and "mo"
- add darker border color around the 4 boxes when in light mode. maybe `#a4a4a4` ?
- add slightly darker background color behind the the 4 boxes when in light mode. maybe `#eee` ?
- mobile layout is a bit wonky.  On my phone, the page loads with it slightly zoomed in so the parameters and cost estimates are cropped on the right. Recommended hosts goes way off the screen. when in responsive mode I would make all 4 boxes 100% width and ensure they don't crop or require horizontal scrolling to see
- if you're trying to set up a small amount of users, say 100, it's pretty much impossible to get there: it jumps from 99 to 134 with no inbetween when using mouse input (I do know you can click the slider and use arrow keys, but lets assume most folks won't do this.) Interestingly my zoom level affected the numbers it jumped between!  Suggestion: go in increments of 100 if possible or something more smooth.  Age of deployment works really well with a range of 1-10 years in 1 year increments.
- "Cost Per User" should be shown in $/yr like the rest of the system.  We could do a smaller $/mo like we do in the larger "Cost estimate"
- Given the calculator is likely both 99% of what people want and has disparate data than the current page, make the calculator the default under "Hosting Cost" and move the existing page to something like, "Hosting Cost Case Study".  
- for "Storage Breakdown", spell out "Over-provisioning" in full. Likely if we get rid of decimals on $s, make RAM and CPU on their own line (instead of one line with a `+`), we'll have room to split the "Recommended Host" box 50/50 instead of the 66/33 (or what ever it is ;)
- remove "Block Volume Storage" under "Disk" - it's effectively a redundant label even if it's accurate. 
- I'm torn on the inclusion of "Age" as a slider. Is it helpful to know your 100k population, 5k users will cost $3.6k in year one, but $13k in year 10? Likely year 2 and 3 are key to know there's an increase.   Maybe break this out into a separate "cost over time" table?  Maybe leave as is?!

Calculator logic:
- When I do `400k` population, 4 workflows, 1 year, `200` users, my "People per user" is `2400` out of a scale of `250`. Similar issues for "Docs per user".  Pretty as they are, I'm not sure a slider visualization works for these two.  Maybe it does for the happy path and I'm over thinking it? Clearly 200 users can't service 400k population ;)
- We should include info about "resource utilization". I was confused how I could be adding 100s of users but my cost was static - it was because resource utilization started at 12% and was working it's way toward 90%
- I don't think we can sanely allow say...>80% utilization?  I was able to get it to >99% which will guarantee a high replication latence - sad times. Suggestion: add an "advanced" setting to flip the recommended host over to the next level up at 75% utilization.
- Add a total for "Storage Breakdown". I was so confused why Used and Overprov were always the same.
- Sad as it is, sigh...we need to set "Database Over-provisioning Factor" to a default of 5. Given our upgrades require it, I don't see how we could reasonably suggest any other amount :(
- not sure what the solution is, but 100k population with 10k users, 5 workflows, at year 10, won't have `0.0%` utilization. Feels odd to bake in a fixed 20% minimum.  Maybe an advanced setting?!  Maybe I shouldn't over complicate it.

jkuester:
Okay, I think this is ready for another round of review!  

I have tried to further simplify things (while allowing for more precision and flexibility) by moving away from targeting fixed sample ec2 instances in the logic and instead just using a set per-CPU cost and then determining the needed CPU count based on the number of users.  I am just targeting `~50%` CPU utilization, so the "Resource Utilization" visualization was no longer needed.

I also updated the default over-provisioning to `3` (based on the latest 5.x numbers).  I am willing to go to `5`, though, if you feel it is essential. :+1: 

Otherwise, I think the "Calculation Details" section hopefully contains sufficient justification/explanation for the rest of the calculation logic/details. One of my goals with that section was to provide enough detail that it would be easy to reproduce the results (or calculate updated values in the future).

mrjones-plip (approved):
This is great!  Thank so much for tending all my feedback - it just works!  I think the optimizations you've made make the tool much easier to read and more accessible for the high level planners that might use it. While I have some suggestions, I don't think they're blocking and will `approve` now assuming someone will bust out the code review while I'm out next week and we can ship.

I'll be happy to do a deeper dive either if this PR is still open when I get back or on the live code in a separate issue/PR as needed.

------------------

The biggie that I didn't have time to dive into the calculations.  They feel right, but I didn't do any sanity checks. Before going live we should still have Elijah/Loukman/Philip do this sanity check.

I think we should move the `Calculation details` section right below the calculator and shift all the other sections as is down so it's followed by `Accuracy`, `prod vs dev` etc..  Really, the calculator is what people will be here for and we want to answer questions about how we came to our numbers right away. 

It might be nice to have thousandth separators? I suspect that's US centric, so no biggie,   but `$11,020/yr` is so much easier  for me to parse than `$11020/yr`. No biggie if others disagree and we don't add it.

I still think it's weird that when you do something silly like 1.5M people, 1 workflow, 10 users, you have 156k people per user, but with a max of 250.  Same for docs per people (370k max of 20k).  Maybe remove the small numbers (`250` & `20k`)?  I don't feel strongly, but feel like it might open us up to unneeded scrutiny.

Maybe make this change?  Seems simpler, but I don't feel strongly:

Image

I'd also consider removing the base 50GB somewhat like it was before - but - meh - not that big of a deal.

sugat009 (commented):
Review from code quality, security, and architecture analysis.

assets/js/cost-calculator.js
```diff
@@ -0,0 +1,239 @@
...
+    yearlyCostPerUser,
+    { minimumFractionDigits: 2, maximumFractionDigits: 2 }
+  );
+  els.costPerUser.textContent = formatCurrency(
+    monthlyCostPerUser,
+    { minimumFractionDigits: 2, maximumFractionDigits: 2 }
+  );
+
+  els.popPerUser.textContent = m.popPerUser.toFixed();
+  const popPct = updateRangeMarker(els.popPerUserMarker, m.popPerUser, 1, 250);
```
sugat009: **Magic numbers not in `DEFAULTS`**

The bounds `250` (max people per user) and `20000` (line 121, max docs per user) are hardcoded here but have no corresponding entry in the `DEFAULTS` object at the top of the file. If these values need tuning (the same way `USERS_PER_CPU` and `MEDIC_DOCS_PER_GB` were derived from production data), they are invisible to future maintainers.

Suggestion: add them to `DEFAULTS`:
```js
MAX_POP_PER_USER: 250,
MAX_DOCS_PER_USER: 20000,
```

eljhkrr:
Thanks for the great work! 

1. For sampling instances, I'd recommend:
    - Small - Elgeyo Marakwet, Mandera (cluster 1 - shared hosting on echistanariver node), Kitui, Kirinyaga (cluster 1 - shared hosting on echistharakanithi node) & Mombasa (cluster 1 - shared hosting on echistanariver node)
    - Large - Homa Bay, Bungoma, Trans Nzoia, Makueni & Meru
    
     Most of the instances are sitting on individual VMs apart from the ones on shared hosting

2. The calculator looks good and I have opinions on possible useful additions:
    - Use more direct language by switching 'users' for 'health workers' or CHWs
    - A time series of cost vs deployment age would be useful for visualizing expected cost changes over time

jkuester:
Okay, thank you everyone again for your helpful input!  I have updated things based on Elijah's comments (and with new data from the recommended instances):

- MEDIC_DOCS_PER_GB: `76111` > `186089`
- USERS_PER_CPU: `211` > `288`

The improvement here was largely expected because my previous numbers were assuming the tharakanithi and tanariver nodes were only serving a single CHT instance each.  I balanced out these gains by increasing the default DB over-provision factor from `3` > `5` (based on the latest 5.x upgrade numbers we have seen).  So in the end for our default specs, the price-per-user-per-month went from `$0.11` > `$0.09` (which is better than our _current_ "estimated cost" of `$0.10`).  :+1:   


image


@mrjones-plip could you give this one final smoke test early next week just to be sure (got me all cautious here after the discussion in the TCO squad).  Otherwise, I think this is ready to rock and roll!  :man_dancing: '

PR #2111, opened 2026-01-15
CHT-Docs MCP Server Documentation

Hareet:
### Description

Me and agent Claude created documentation around the CHT-Docs MCP Server. Genspark helped create the image.

Notes:
- I think I remember we wanted an "AI + CHT" at the top-level section. Feel free to recommend another re-organization of the titles.
- I tried to add an intro paragraph about AI and it's use with CHT, but wasn't sure if that should live here. I was trying to have something similar to content in the "Why CHT?" section, that answered like why are we trying to implement AI or for what.
- I wasn't sure if we wanted to nest CHT-Docs under MCP Servers, or have that content just be listed inside the MCP Server category.
  - We'll be getting other MCP servers, so we can always modify that later


Hareet:
Using a custom /docs-reviewer command we built with Claude:

https://github.com/user-attachments/assets/8cb93f8b-c552-4dd8-ae12-b3bdaf2e5ee3

andrablaj (commented):
I have some minor wording suggestions, but nothing is blocking. I am excited about this new category in our Docs Site!

PR #2156, opened 2026-03-21
Add dynamic Create Docs Issue link with prefilled page context

### LINKED ISSUE #2025: Add Create documentation issue link on the documentation pages that passes the page name and link to page
Bring back `Create documentation issue` on CHT documentation site. This used to be on old documentaiton site. This feature is useful for anyone to report issue in a documentation page with Required information already passed through the github parameters.

For anyone working on this, In a github issue link, a `title` and `body` parameter can be passed to new issue endpoint. Something like this: 
https://www.github.com/medic/cht-docs/issues/new?title=Improve{current_page}&body=Page:/link/to/current/page/or/section

This should be available on all document pages. 

**Additional context:**

Currently, there's `Create doc issue` on the left navigation on documentation site that takes to new issue template without any information.

gmarav05:
# Description

This PR enhances the "Create Docs Issue" link by pre-filling the GitHub issue form with the current page context.

Fixes #2025 

## Changes

1) Dynamically updates the "Create Docs Issue" link to include.
  - Page title in the issue title
  - Page URL in the issue body

2) Implemented using JavaScript to ensure compatibility with the Hugo theme module setup.

## Why

Previously, the "Create Docs Issue" link opened a blank issue form, requiring users to manually provide context.

This change improves user experience by automatically including the page name and link. It also makes it easier to report documentation issues.

## Screenshots

https://github.com/user-attachments/assets/65e81b32-08a0-4d52-93c7-9d005d756da9

# License

### DIFF
--- a/layouts/docs/single.html
+++ b/layouts/docs/single.html
@@ -18,6 +18,7 @@
       {{ partial "components/feedback.html" . }}
       {{ partial "components/pager.html" . }}
       {{ partial "components/comments.html" . }}
+      {{ partial "custom/docs-issue-link.html" . }}
     </main>
   </article>
 </div>
--- a/layouts/partials/custom/docs-issue-link.html
+++ b/layouts/partials/custom/docs-issue-link.html
@@ -1,0 +1,4 @@
+{{- $issueTitle := urlquery (printf "Improve: %s" .Title) -}}
+{{- $issueBody := urlquery (printf "Page: %s" .Permalink) -}}
+{{- $link := printf "https://github.com/medic/cht-docs/issues/new?title=%s&body=%s" $issueTitle $issueBody -}}
+<a href="{{ $link }}" target="_blank" rel="noreferrer">Create Docs Issue ↗</a>
--- a/layouts/partials/custom/head-end.html
+++ b/layouts/partials/custom/head-end.html
@@ -402,5 +402,6 @@
   } else {
     init();
   }
+
 })();
 </script>


mrjones-plip (changes requested):
Thanks for the PR!  I see your JS based solution works well in that dynamically picks up the URL and page name and puts it into the link to create a docs issue in GitHub.

While this is a small amount of JS to run client side, ideally we could statically generate this when we generate the site and have zero client side operations. Could you see about using a static shortcode solution instead?  I think maybe the examples of the [existing logic](https://github.com/medic/cht-docs/blob/main/layouts/shortcodes/github-url.html) to generate the `$githubURL`  along with this [template functions](https://discourse.gohugo.io/t/hugo-v0-112-0-new-template-functions/44512) could unlock how to achieve this.

mrjones-plip (changes requested):
Nice work!  

We need to simplify the logic - this PR now has almost 200 lines of template that include a ton of logic for unrelated changes.  Please see the [blog post](https://discourse.gohugo.io/t/hugo-v0-112-0-new-template-functions/44512) I cited already in hopes a much more simple solution of a single partial that we can include in either the `hugo.toml` file where we define the sidebar links or in the `single.html` where we define a page.

If that's not tenable, then let's revert to the prior approach of javascript as this logic, while using client side CPU, is much less invasive and easier to maintain.

thanks!

mrjones-plip (changes requested):
Thanks for the follow up @gmarav05 !  

While the templates are perfectly clean, there's some issues (see below).  Please test your code before requesting feedback.  

Issues:
* for the template `layouts/docs/single.html` the link is showing below the body instead in the left hand nav (`#1) (testing on [SSO page](http://localhost:1313/hosting/sso/overview/))
* the old link is still in place (`#2`) (testing on [SSO page](http://localhost:1313/hosting/sso/overview/))
* the `docs-issue-link.html` wasn't added to the `layouts/docs/list.html` document so it's not showing on all pages (see [sso landing](http://localhost:1313/hosting/sso/) and [home page](http://localhost:1313/))

Image

mrjones-plip:
Also - as a follow up - if it's too complicated to use this (wonderfully simple and clean!) `layouts/partials/custom/docs-issue-link.html` [solution](https://github.com/medic/cht-docs/pull/2156#pullrequestreview-4047520342) above, feel free to go back to your [earlier JS solution](https://github.com/medic/cht-docs/pull/2156#pullrequestreview-4009440221).

PR #2184, opened 2026-04-10
fix(#8974): remove note about form `end` issue

jkuester:
# Description

cht-core issue: https://github.com/medic/cht-core/issues/8974
cht-core PR: https://github.com/medic/cht-core/pull/10784

There used to be a bug in the cht-core handling of the `end` field in forms.  Now the bug is fixed!

# License


mrjones-plip (approved):
w00t!  great to see this old item be cleaned up with the bug fix - thanks for remembering! 

🐈

PR #2190, opened 2026-04-17
feat(#10508): document Android app support for primitive arrays

jkuester:
# Description

@mrjones-plip can I get a quick review when you have a second. 

This is for functionality I just merged in over in cht-core: https://github.com/medic/cht-core/pull/10842

Issue: https://github.com/medic/cht-core/issues/10508

# License

### DIFF
--- a/content/en/building/forms/app.md
+++ b/content/en/building/forms/app.md
@@ -142,6 +142,8 @@ To define the widget, create a `group` with the appearance `android-app-launcher
 | ... | ... | ... | ... | ... | ... | ... |
 | end group | camera-app |  |  |  |  | ... |
 
+#### Basic input/output
+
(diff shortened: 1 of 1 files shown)


mrjones-plip (approved):
would love to see an example of how "Arrays of primitive values"  works, but not at all blocking to go live.  thanks for adding the missing "Serializing/Deserializing structured data " header - the old doc is so awkward without it!

jkuester:
Yeah, so I thought about adding an example, but there is already a lot going on in that section and really the only thing to show is a plane `text` row.... :shrug:   

Going to leave it like this for now, but in the future we can consider breaking these docs into their own page.  (Especially as this functionality gets more and more attention.)

PR #2187, opened 2026-04-13
fix(#2186) fix CI so push to prod works

mrjones-plip:
# Description

This PR:
* fixes CI so push to prod works by bumping actions to `v4` vs `v2`
* converts all pinned versions to hashes as a [best practice](https://mattsch.com/blog/2026/03/28/harden-your-github-actions-workflows-with-zizmor-dependency-pinning-and-dependency-cooldowns/#dependency-pinning) - I used `pinact` [CLI util](https://dev.to/suzukishunsuke/pin-github-actions-to-a-full-length-commit-sha-for-security-2n7p)

Closes #2186

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

Issue #2189, opened 2026-04-16
Add dependabot version bumps to github CI, include cooldown


mrjones-plip:
We [recently](https://github.com/medic/cht-docs/pull/2187) changed from versions (`@v4`) to hashes (`b7bcd026f18772e44fe1026d729e1611cc435d47`).  This offers a strong defense against supply chain attacks.  however, we should continue to update these actions to be current in an automated fashion.

like we did with CHT Core [recently,](https://github.com/medic/cht-docs/pull/2187) we should add a dependabot action that updates the action hashes and includes a cool down(see [CHT Core's PR's dependabot file](https://github.com/medic/cht-docs/pull/2187) for an example) .  This way we can both keep the actions up to date, continue to use hashes, but not rush to implement something that may have a vulnerability by using the 7 day cooldown

rovertrack:
can i work on this?

PR #2192, opened 2026-04-20
chore(na): bump actions - link checker asdf & prod push rtCamp

mrjones-plip:
# Description

this PR 
* bump actions  in link checker `asdf` as it's [broken now](https://github.com/medic/cht-docs/actions/runs/24657793121). same as [we did earlier in main ci](https://github.com/medic/cht-docs/pull/2187)
* as well it bumps prod push rtCamp action to be the same, newer version as in link checker

# License

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Main Branch Only - Report errors to Slack, if any
         if: ${{ github.ref == 'refs/heads/main' && failure() }}
-        uses: rtCamp/action-slack-notify@96d5e2a64fc78a6b7ac13265f55bee296869967a # v2.0.2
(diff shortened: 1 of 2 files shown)


sugat009 (approved):
**question (non-blocking):** The step name in `cron-check-links.yml` was changed to "All Branches - Install asdf & tools in .tool-versions". In ci.yml this prefix makes sense because it triggers on `push` and `pull_request` across branches, but `cron-check-links.yml` triggers via `on: schedule` which only runs on the default branch. Was the rename intentional here, or should the original name be kept?

**question (non-blocking):** Both files have the header comment "If you bump one version up to latest, please bump them all so we stay current!" `actions/checkout` is still at v2.7.0 in both files (latest is v4.x). Worth bumping here too, or better as a separate follow up?

mrjones-plip:
Thanks @sugat009 !  Appreciate your eye for detail. Action name was reverted - good call - see 468fef4 .  For bumping to a newer version, we'll fix this now and ongoing in #2189

PR #2195, opened 2026-04-21
fix(#2188): fix mobile nav visibility and desktop navbar behavior

### LINKED ISSUE #2188: navigation doesn't work on mobile site
In Chrome or Firefox on Android, the mobile interface doesn't let you use the hamburger menu to navigate the site.  Instead of seeing the navigation, you only see the last 4 links:

Image

Briefly checking, I built the site on commit [before](https://github.com/medic/cht-docs/commit/ded5b853c1151ba858284c647a2fc7e12efa50d3) we [added](https://github.com/medic/cht-docs/commit/c13d723e7cd01078147340f234156661ba34bb7c) Kapa AI and it works as expected.  This very likely means the HTML/DOM got changed in a way to break responsive mode:

Image

gmarav05:
# Description

This PR fixes the mobile menu and keeps desktop navigation working as expected.

Closes #2188 

## Changes
- Updated `hugo.toml` menu settings.
- Added `hideInNavbar = true` for section links that should not appear in the desktop top bar.
- Kept top-right utility links (`Search`, `GitHub`, `Website`).
- Kept sidebar “More” links (`Create Docs Issue`, `CHT Forum`, `CHT Academy`, `Privacy Policy`).
- Added `layouts/_partials/navbar.html` to control navbar behavior.
- Updated `i18n/en.yaml` labels for menu and UI text.

The default setup did not handle mobile and desktop nav the way we needed and it fixes mobile nav while keeping desktop clean.

# License

https://github.com/user-attachments/assets/4731eae2-8bda-4278-a125-35cb686b2e4f


mrjones-plip (approved):
I was hoping to avoid adding the new partial layout for navigation, but it looks like it was unavoidable. 

thanks for the fix!

PR #2197, opened 2026-04-22
feat(#10650): document SSO max_age setting

jkuester:
# Description

Docs updates for the new SSO feature coming in https://github.com/medic/cht-core/pull/10801.

See https://github.com/medic/cht-core/issues/10650 for more details.

# License

### DIFF
--- a/content/en/hosting/SSO/overview.md
+++ b/content/en/hosting/SSO/overview.md
@@ -42,9 +42,15 @@ SSO authentication is implemented with the industry standard [OpenID Connect](ht
 
 ### Require re-authentication
 
-By default, SSO Authentication will use the user's current session with the OIDC Provider (if one exists). If the user is already logged in and has an active session with the OIDC Provider, the user may be automatically logged in to the CHT when clicking the "Login with SSO" button (without actually needing to re-authenticate).
+{{< callout >}}
+Added in TODO
+{{< /callout >}}
 
-In some cases, this behavior is not desired and the user should be required to re-authenticate with the OIDC provider. You can define the maximum allowed age of the user's current session with the OIDC Provider, after which the user will be required to re-authenticate when logging into the CHT even if their session is still active. Set the `oidc_provider.max_age` setting to the allowable elapsed time in seconds since the last time the user was actively authenticated by the OIDC Provider. If the elapsed time is greater than this value, the user will be required to re-authenticate.
+By default, SSO Authentication will use the user's current session with the OIDC provider (if one exists). If the user has already authenticated with the OIDC provider and has an active session on the device, the user may be automatically logged in to the CHT when clicking the "Login with SSO" button (without actually needing to re-authenticate with the OIDC provider).
+
+In some cases, this behavior is not desired and the user should be required to re-authenticate with the OIDC provider. For example, if you have multiple users sharing the same device, it may be challenging for an SSO user to log out of the CHT and have a different user log in to the CHT. The previous user's active session with the OIDC provider might get used during the SSO login flow (even though the previous user's _CHT session_ was ended when they logged out). This results in the new user being logged into the CHT account for the previous user.
+
+To avoid this behavior, you can define the maximum allowed age of the user's current session with the OIDC Provider, after which the user will be required to re-authenticate when logging into the CHT even if their session is still active. Set the `oidc_provider.max_age` setting to the allowable elapsed time in seconds since the last time the user was actively authenticated by the OIDC Provider. If the elapsed time is greater than this value, the user will be required to re-authenticate.
 
 Setting `max_age: 0` will always require re-authenticating with the OIDC Provider when logging into the CHT.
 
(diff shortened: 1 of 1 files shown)


content/en/hosting/SSO/overview.md
```diff
@@ -40,10 +40,17 @@ SSO authentication is implemented with the industry standard [OpenID Connect](ht
...
 5. Use the "Login with SSO" button on the CHT login page.
 
+### Require re-authentication
+
+By default, SSO Authentication will use the user's current session with the OIDC Provider (if one exists). If the user is already logged in and has an active session with the OIDC Provider, the user may be automatically logged in to the CHT when clicking the "Login with SSO" button (without actually needing to re-authenticate).
+
+In some cases, this behavior is not desired and the user should be required to re-authenticate with the OIDC provider. You can define the maximum allowed age of the user's current session with the OIDC Provider, after which the user will be required to re-authenticate when logging into the CHT even if their session is still active. Set the `oidc_provider.max_age` setting to the allowable elapsed time in seconds since the last time the user was actively authenticated by the OIDC Provider. If the elapsed time is greater than this value, the user will be required to re-authenticate.
+
+Setting `max_age: 0` will always require re-authenticating with the OIDC Provider when logging into the CHT.
```
mrjones-plip, line 55: "when logging into the CHT" is key though, right?  The SSO flow has the CHT [set the session cookie](https://docs.communityhealthtoolkit.org/hosting/sso/technical/#oidc-login-sequence) - so what does this mean in practice to have `max_age` be `0`? How long will the session cookie be which is the same way of asking: how often will they need to re-authenticate with the SSO provider? It can't be "every time" because they couldn't effectively use the CHT.  But if it uses the default CHT value of a year, I'm not sure what a value of `0` gains you here?

But I feel like I'm maybe I'm missing something!

mrjones-plip (changes requested):
I thought I was just gonna approve, but then the more I thought about it, the more confused I got.  I ended requesting a change in the form of a question.  We may answer it with no changes, or a clarification in this new section about the two different session expiration times

content/en/hosting/SSO/overview.md
jkuester: You are correct!  None of this affects the way _CHT login sessions_ work. A CHT user (SSO or not) that is logged into the CHT will not be required to re-authenticate (as long as they have accessed the CHT server within the last year).  This setting pertains to the user's session _with the OIDC Provider._  

The scenario is that you have configured your CHT instance to use Google as your OIDC Provider. On the device, User A uses "Login with SSO" to login to the CHT and in the process authenticates via their Google account.  At that point, on the device, there is both a CHT session and a Google session, both for User A.  

Next, User B needs to use the device.  So, User A logs out of their CHT account (and so their CHT session is removed from the device) and gives the device to User B.  User B taps "Login with SSO", but instead of being presented with an opportunity to login to their Google account, instead the OIDC flow automatically redirects to Google which sees there is already and active/approved session on the device (from User A). So without presenting a login page or requiring any re-authentication Google redirects back to the CHT with valid authentication codes for User A. And the CHT server happily completes the login flow and User B is actually logged in with a CHT session for User A.  

The only way to escape this flow is to actually end User A's Google session on the device, but this opportunity might never actually presented during the CHT login flow because the auto-redirects just do their thing.  

---

It is also worth nothing that these auto-redirects and the reusing of the existing SSO session are _by design_ and there are definitely other more common use cases that justify this behavior being the default.  Take the case where you are provisioning a bunch of single-user devices. Each one is pre-configured with the Google account for that particular user.  By using SSO login for the CHT in this case, you do not have to have an additional provisioning step to worry about logging in the users to the CHT app.  If the user is already logged into their Google account on the device, then all they have to do to access the CHT is click "Login with SSO" and they will be taken right to their CHT user.

It is only in some special cases like multiple users on the same device where you are logging out of the CHT and back in as a different user where things can get rough and you would possibly want `max_age`.

mrjones-plip: awesome - thank you!! Makes total sense now and I remember this discussion about multiple users of the same device.

mrjones-plip (approved):
Per our discussion, it might be helpful to have as distilled version as possible of the user A and user B as an example of how this might get used. 

otherwise: 🥍

jkuester:
:thinking: @mrjones-plip did you merge this on purpose?  :sweat_smile: The [core PR](https://github.com/medic/cht-core/pull/10801) is still not merged yet.... :grimacing:

mrjones-plip:
oh no!  That's my bad - I'm sorry!  Do you want me to do  quick PR to revert and another to restage this PR so we can ship when it's ready?

It so clearly says `Added in TODO` in there - sorry again :facepalm:

mrjones-plip:
Ok!   Please approve and merge #2235 which will undo the merge in this PR. 

Then, when the SSO feature has shipped, we can update #2236 with the correct release version and merge that.  Alternately, you can just resubmit your  `jkuester:10650_max-age` branch

Sorry again about all this mess!

PR #2200, opened 2026-04-24
feat(#2112): enhance 404 page with branding, search, and helpful links

### LINKED ISSUE #2112: improve 404 page: add CHT branding, search and tips
Right now you get this when you end up on a [404 page](https://docs.communityhealthtoolkit.org/are-we-in-kansas-toto):

```
404
This page could not be found.
```

Image

As an end user, you:
*  do not know if this is a CHT page
* have nothing to click on to get you back to content
* can't search for what you're looking for.

If possible, we should address this by  making the 404 be like any other page, which would mean:

* there's the whole navigation to choose from on the left
* search is at the top
* there's CHT docs branding so you know which site you're on
* we should be sure the `404` header is still sent

Here's a quick and dirty mockup I did:

Image

## Good first issue

This ticket has the [Good first issue label](https://github.com/medic/cht-core/issues?q=state%3Aopen%20label%3A%22Good%20first%20issue%22)! This means it's been especially curated by other CHT contributors to be easy to work on for first time contributors [per the docs](/community/contributing/creating-good-first-issues/).

To succeed on this ticket, please:
1. get the docs system installed [locally](https://github.com/medic/cht-docs/blob/main/README.md) - likely the easiest is to use docker to run `hugo`
2. add an the `404.html` file in the `layouts` directory
3.  put hugo template content in the new file based on [docs](https://gohugo.io/templates/404/) and on the  CleverCloud [example](https://github.com/CleverCloud/documentation/blob/main/layouts/404.html)

imtushar01:
## Problem
The current 404 page only displays “404 - This page could not be found” without navigation, search, or branding. As a result, users cannot:
- Identify that they are on a CHT documentation page
- Navigate back to relevant content
- Search for the information they need

## Solution
Added a custom `layouts/404.html` page aligned with the existing documentation layout:
- Integrated sidebar navigation and top search bar
- Included CHT branding for consistency
- Added helpful guidance and links for users

Screenshot 2026-04-25 at 00 10 13

Fixes #2112

### DIFF
--- a/layouts/404.html
+++ b/layouts/404.html
@@ -1,0 +1,31 @@
+{{ define "main" }}
+<div class='hx:mx-auto hx:flex hextra-max-page-width'>
+  {{ partial "sidebar.html" (dict "context" .) }}
+  <article
+    class="hx:w-full hx:break-words hx:flex hx:min-h-[calc(100vh-var(--navbar-height))] hx:min-w-0 hx:justify-center hx:pb-8 hx:pr-[calc(env(safe-area-inset-right)-1.5rem)]">
+    <main class="hx:w-full hx:min-w-0 hx:max-w-6xl hx:px-6 hx:pt-4 hx:md:px-12">
+
+      <div class="content">
+        <h1 class="hx:text-8xl md:hx:text-9xl hx:font-extrabold hx:tracking-tight">
+          404
+        </h1>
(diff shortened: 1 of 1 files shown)


mrjones-plip (approved):
Thanks so much @imtushar01 !  I simplified the language a bit, but this is good to go.

PR #2201, opened 2026-04-24
fix(#2198): prevent sidebar-only takeover at tablet widths (768px width) and zoom

### LINKED ISSUE #2198: navigation takes over entire screen in some scenarios
In some scenarios, only the left navigation shows, with no content showing, which makes the docs site feel broken.  

I was able to set my browser to exactly 768pixels wide and it regularly happened as I increased the zoom level.   We should ensure this doesn't happen at any zoom level so that it's either behind the hamburger menu or fully visible.

here's video showing it in action and below is a screenshot

<details><summary>video</summary>
<p>

https://github.com/user-attachments/assets/fb79c2f9-5f50-422a-8471-13fba2afb95b

</p>
</details> 

Image

gmarav05:
Prevent sidebar-only view at 768px

closes #2198 

# Description

This PR fixes a responsive layout issue where the left sidebar could take over the page around 768px width (especially with browser zoom), making main content hard to access.

### What I changed
- Updated `assets/css/custom.css`.
- Added a tablet/mobile CSS guard for `.hextra-sidebar-container` (`max-width: 1023.98px`).
- Set fixed position and height so it behaves like a proper drawer.
- Ensured main content (`article`) keeps `width: 100%` and `min-width: 0`.

### Result
At 768px, the layout no longer gets stuck in a sidebar-only state.

# License

Screenshot 2026-04-25 at 1 07 37 AM

### DIFF
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -282,9 +282,3 @@ Fix Resolve conflict between Hextra's 48rem and Tailwind's 768px breakpoints
     transform: none !important; 
   }
 }
-
-@media (max-width: 767.98px) {
-  aside.hextra-sidebar-container:not(.hx\:max-md\:\[transform\:translate3d\(0\,0\,0\)\]) {
-    transform: translate3d(0, -100%, 0) !important;
-  }
-}
--- a/assets/css/custom.css
+++ b/assets/css/custom.css
@@ -273,10 +273,18 @@ video {
 }
 
 /* -----------------------------------------------
-Fix Close subpixel gap between max-md (767px) and md (768px) breakpoints
+Fix Resolve conflict between Hextra's 48rem and Tailwind's 768px breakpoints
 --------------------------------------------------*/
-@media (min-width: 767.01px) and (max-width: 767.99px) {
-  .hextra-sidebar-container:not(.hx\:max-md\:\[transform\:translate3d\(0\,0\,0\)\]) {
+@media (min-width: 768px) {
+  aside.hextra-sidebar-container {
+    position: sticky !important;
+    width: 16rem !important; 
+    transform: none !important; 
+  }
+}
+
+@media (max-width: 767.98px) {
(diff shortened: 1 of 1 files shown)


mrjones-plip (changes requested):
thanks for the PR @gmarav05 !

While the PR was quick, please be more thorough in your testing:

* the nav bar  is  hidden, you can't access it
* there's a "system" theme selector in the upper left and it persists through out all responsive modes, introducing regressions. this shouldn't be shown until you click the hamburger menu
* speaking of which, the hamburger menu, search and other top icons are not in the mobile responsive mode
* I don't think we need any [color change needed](https://github.com/medic/cht-docs/pull/2201/files#diff-54a66f9a42172ece239b80abd6538c28c80c25375a2ef411908be80fac1ef58fR286) - can you explain why you added this?
* the [media detection width](https://github.com/medic/cht-docs/pull/2201/files#diff-54a66f9a42172ece239b80abd6538c28c80c25375a2ef411908be80fac1ef58fR278)  of `1023.98px` seems suspicious - why did you choose this value? 

while a relatively small PR, I suspect the final PR will be even smaller that fixes where two `@media` and non-`@media` CSS declarations don't quite overlap.

please try again

gmarav05:
Thanks for the detailed review @mrjones-plip .  You’re right, my CSS fix was too broad and caused regressions.  

Now i have made some updates:  
- Removed sidebar override and theme/color changes .
- Applied minimal breakpoint fix at md boundary.
- Re-tested at 767/768/769 with zoom and menu scenarios.

gmarav05:
Screenshot 2026-04-25 at 12 31 16 PM

mrjones-plip (approved):
Thanks @gmarav05 !  

Since the hide/show feature of navigation should never be visible at the `768p` width, I've removed it from this PR since it's not needed

gmarav05:
yeah sure @mrjones-plip, Thank you for making that change and for your feedback.

PR #2206, opened 2026-04-29
chore(na): remove g++ & make, use python3 isntead of 2 for windows

mrjones-plip:
# Description

When I recently [deprecated the windows page](https://github.com/medic/cht-docs/commit/2b746016349b04efc40ad6d7c9a4484eef4b96d9), I noted we still installed python2 for windows - which isn't right.  Then  I saw we still install `make` and `g++` which I was used to when still installed stuff with `python` and `pip` which we no longer do.  So it should be safe to remove? Counting on my code reviewer to keep me honest

# License

### DIFF
--- a/content/en/community/contributing/code/core/dev-environment.md
+++ b/content/en/community/contributing/code/core/dev-environment.md
@@ -35,7 +35,7 @@ _(Node {{< param nodeVersion >}} is the environment used to run the CHT server i
   {{< tab >}}
 ```shell
   sudo apt update && sudo apt -y dist-upgrade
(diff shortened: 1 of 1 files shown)


jkuester (approved):
All good for removing `gcc`, `g++`, and `python2`.  I do not have any of those available in my default shell environment and have not needed them.  👍 

I do think we should keep `make` as it is used for the haproxy and couch tests.

mrjones-plip:
awesome - thanks @jkuester ! Added back in `make` in 43f26f1a - merging!

PR #2209, opened 2026-05-01
feat(#10341): remove `chtPermissionsSettings` param

jkuester:
# Description

As a part of resolving https://github.com/medic/cht-core/issues/10341, https://github.com/medic/cht-core/pull/10795 has removed the need for every passing the `chtPermissionsSettings` parameter when calling `hasPermissions` or `hasAnyPermissions` on the cht-datasource api (from tasks/targets/contact-summary/purging config code).  Previously this parameter was optional for tasks/targets/contact-summary since the webapp would automatically supply it if it was missing. However, it was required for purging config.  Now it is not required at all and there is no situation where the  `chtPermissionsSettings` parameter should ever be provided.  (Though we still accept it for backwards compatibility.)  

I am removing documentation for this parameter since it should no longer be relevant (even for purging config).

# License


sugat009 (approved):
Approved. One question I wanted to ask, no problem either way.

Since `chtPermissionsSettings` is technically still accepted (you've framed it as deprecated in the `index.ts` docs) and cht-docs isn't version pinned, would a small marker on these two rows help users on CHT < 5.2.0? Something like:

`_Updated in v5.2.0: chtPermissionsSettings is no longer required._`

in line with the existing `_Introduced in v4.11.0_` pattern on `analytics.getTargetDocs()`. Curious what you think.

Also a tiny nit: stray `<br>` left at the end of the `userRoles` description on the `hasAnyPermission` row.

jkuester:
Thanks @sugat009!  I ended up going for a bit extra completion here.  The `chtPermissionsSettings` has been optional for the webapp config (contact-summary, tasks-targets) since it got added to the webapp so I don't think we need to worry about documenting the parameter for _that API_.  However, the purge config actually has access to a slightly different CHT API (without the webapp extensions).  To try and mitigate some of the config confusion here, I have updated the purge documentation to have its own API reference docs that are more accurate. In those docs, I noted how the permissions settings parameter was optional after `5.2.0`. :+1:

PR #2211, opened 2026-05-06
fix(#1866): restore bootstrap grid classes to fix huge images

### LINKED ISSUE #1866: Review image size with new theme
Some pages have huge images, and it would more readable if they were smaller. Example: 

- [Android](https://docs.communityhealthtoolkit.org/building/branding/android/#manually-testing-on-the-device)
- [Access](https://docs.communityhealthtoolkit.org/building/concepts/access/#downloading-and-launching) vs [new](https://hextra-docs.dev.medicmobile.org/building/concepts/access/#downloading-and-launching)
- [Contacts](https://docs.communityhealthtoolkit.org/building/contact-management/contacts/)

ayushman1210:
# Description

Fixes #1866.

Some images on the documentation site were rendering at their original, huge sizes because the new theme does not support the legacy Bootstrap grid classes (`col-6`, `col-lg-8`, `right`, etc.) that are extensively used in the `figure` shortcodes across the markdown files.

This PR adds these legacy utility classes to `assets/css/custom.css`. By doing so, it restores the original layout constraints and correctly scales down the images across the entire documentation site, without requiring manual updates to hundreds of markdown files.


mrjones-plip (changes requested):
Thanks for the PR @ayushman1210 !

There should be no CSS removed in this PR, correct?  If yes, please explain why, otherwise please restore the CSS that was removed, including comments.

Please add a comment as to why the old CSS is being added back in.

Finally, please add three examples of before and after screen shots showing the images being too big and then being fixed with the new CSS.  Thanks!

PR #2219, opened 2026-06-05
feat: add more advanced params to cost calculator

jkuester:
# Description

Based on feedback [on slack](https://medic.slack.com/archives/C024KTGRW/p1774889431388059), I have made several changes to the Hosting Cost Calculator:

- I have added some copy to the page to provide a bit of context regarding the ratios on the Health Workers panel (guiding folks to more realistic numbers).
- I added some new logic that ties the CPU/RAM requirements to the number of docs (and not just the number of users). So, as you increase the Population/Workflows sliders, you will also increase the requirements for CPU/RAM.
  - The logic here is based on the docs-per-user (with the idea that if your average number of docs per user is too high, it will also impact CPU/RAM when syncing).
  - Basically if your number of docs per user is above the baseline (`8717` - calculated as an average-ish docs per user [for production instances](https://docs.google.com/spreadsheets/d/1OgJ-szJjHCd8fFjdn_nzGKWN9H0rp6xczR7gpP2mLk8/edit?usp=sharing)) then we start increasing the "number" of users that you have to provide CPU for.
  - It is not super scientific, but given our recent findings about the performance implications of high-doc users, it is probably still erring on the low side...
- I have also added new "Advanced Parameters" for docs per workflow and users per CPU. These were both customizations that were specifically requested (on the slack thread).

I did not make any visual changes to the layout (besides the new elements on "Advanced Parameters").

## AI Disclosure

Claude did most of the heavy lifting here, but I wrote all the copy and carefully reviewed the code changes.

# License

### DIFF
--- a/assets/css/cost-calculator.css
+++ b/assets/css/cost-calculator.css
@@ -116,16 +116,13 @@ html.dark {
     flex-direction: column;
 
     .advanced-inputs-scroll {
-      overflow-y: auto;
+      overflow-y: scroll;
       min-height: 0;
       max-height: 19rem;
       margin-right: -0.75rem;
       padding-right: 0.25rem;
       scrollbar-width: thin;
-      scrollbar-color: transparent transparent;
-      &:hover {
-        scrollbar-color: var(--calc-border) transparent;
-      }
+      scrollbar-color: var(--calc-border) transparent;
     }
 
     > .calc-divider {
--- a/assets/js/cost-calculator.js
+++ b/assets/js/cost-calculator.js
@@ -6,7 +6,6 @@ const DEFAULTS = {
   USERS_PER_CPU: 288,
   RAM_PER_CPU: 2,
   COST_PER_CPU_MONTH: 20.85,
-  DOCS_PER_POP_WORKFLOW_YEAR: 1,
   ROOT_VOLUME_GB: 50
 };
 
@@ -18,6 +17,7 @@ const UI_CONSTANTS = {
   USERS: { value: 1000, min: 10, max: 15000, step: 10 },
   DEPLOYMENT_AGE: { value: 1, min: 1, max: 10 },
   DB_OVERPROVISION: { value: 5, min: 1, max: 10 },
+  DOCS_PER_WORKFLOW: { value: 1, min: 1, max: 100 },
 };
 
 const formatCurrency = (amount, opts = {}) => amount.toLocaleString(navigator.language, {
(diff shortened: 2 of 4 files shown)


mrjones-plip (changes requested):
Yay!  Thanks for iterating on this.  I want to confirm my findings and then have one minor UI clean up that we can do.

> as you increase the Population/Workflows sliders, you will also increase the requirements for CPU/RAM.

I input 2000000/10/1000 on the old one I get `$1,757/yr` and on the new one I get `$3,258/yr`. Obviously not sane at 2k patients/CHW...

 **I want to confirm** the 1.8x increase is what your above statement translates to, right? When I go to 10k CHWs new and old both even out at `$9513/yr`, which sounds right-ish

I see I can "revert" this so  new old old match up  by setting users/cpu to `688` in advanced - neat!

---------------------

In advanced, we need to avoid scrolling as I didn't notice I'd scroll wheeled while over the advanced pane and then FF hid the scroll bar and I thought I broke it: 

I believe setting `max-height` from `19rem` -> `21rem` fixes it maybe?

```css
.cht-cost-calculator {
  & .calc-card.params {
    & .advanced-inputs-scroll {
      overflow-y: auto;
      min-height: 0;
      max-height: 21rem;
      margin-right: -0.75rem;
      padding-right: 0.25rem;
      scrollbar-width: thin;
      scrollbar-color: transparent transparent;
    }
  }
}
```

Image

jkuester:
> I want to confirm the 1.8x increase is what your above statement translates to, right? When I go to 10k CHWs new and old both even out at $9513/yr, which sounds right-ish

Right this is exactly what we expect to see.  The main difference in the new logic is that when your doc count per user gets too high, it eventually starts driving up your CPU/RAM requirements.

> In advanced, we need to avoid scrolling as I didn't notice I'd scroll wheeled while over the advanced pane and then FF hid the scroll bar and I thought I broke it:

I want to push back on this a bit. I think the experience is better if we keep the card the same size, but always show the scroll-bar on the advanced params.  This way the page layout does not dramatically change when switching to the advanced card, and the user is visually prompted that they can scroll for more settings.  I think it is also more scalable if we want to add more params in the future.

Happy to discuss more if you disagree or have alternative ideas for better UX. :+1:

mrjones-plip (approved):
Gotcha! Agreed on all fronts, thanks for the quick fix .

🕤

PR #2226, opened 2026-07-08
feat(cht-core#10903): document per-doc attachment routing 

benkags:
# Description

Documents the attachment changes shipping in medic/cht-core#10923 (issues medic/cht-core#10903, medic/cht-core#10904), targeting 5.2.0:

- `building/forms/app.md` — binary attachments are stored on the doc that owns the field (nearest `db-doc="true"` section, or the report); new `user-file-<owner-relative-path>` naming for inline binaries, with the reference stored as the field value. Existing docs keep their legacy names.
- `building/forms/configuring/additional-docs.md` — new section on binary attachments on additional docs, including repeats and the fallback for fields outside any `db-doc="true"` section.
- `building/forms/contact.md` — uploads in the `contact` group / `child` repeat attach to that contact's doc; attachment size validation now also applies to contact forms.

Do not merge before medic/cht-core#10923.

# License


mrjones-plip:
Reviving this older docs PR as the feature is shipping in 5.3 which we're hoping to ship soon!  I made some light edits so it says `5.3` instead of `5.2` and changed to be at the end of the feature description matching earlier docs.

cc @benkags and @jkuester to see this PR to get merged!

jkuester (approved):
LGTM!  Thank you @benkags !

PR #2232, opened 2026-08-04
feat(#2231): add multi-tenant info

### LINKED ISSUE #2231: Answer the question  "does CHT support multi-tenancy"
The community has asked at [least once](https://forum.communityhealthtoolkit.org/t/does-cht-support-muti-tenant-architecture-and-how-much-load-a-single-tenant-can-take/5652) about CHT's support for multi-tenancy.  

We should do a write up:
* what defines what multi-tenant is in this context (ie no GUI, technical prowess required)
* that points to existing best practices of hosting, monitoring, backup and reporting (eg CHT Sync) 
* is not that technical in terms there's no numbered steps with a literal process to follow

mrjones-plip:
# Description

Adds work-arounds on how to do multi-tenant

closes #2231 

# License

### DIFF
--- a/content/en/hosting/cht/multi-tenant.md
+++ b/content/en/hosting/cht/multi-tenant.md
@@ -3,20 +3,35 @@ title: Multi-tenant
 linkTitle: Multi-tenant
 weight: 30
 description: >
-  Multi-tenant
+  How to run multiple instances of the CHT on one server
+relatedContent: >
+  hosting/cht/kubernetes
+  hosting/cht/docker
+  hosting/cht/docker/backups
+  hosting/monitoring
+  hosting/analytics
 
 ---
 
-Replace [this ticket](https://github.com/medic/cht-docs/issues/2231) content with real content ;)
+While the CHT has no built-in features to support multi-tenant, there are viable work arounds using existing hosting, backup, monitoring and reporting solutions.
 
---------
+## What is multi-tenant
 
-The community has [asked at least once](https://forum.communityhealthtoolkit.org/t/does-cht-support-muti-tenant-architecture-and-how-much-load-a-single-tenant-can-take/5652) about CHT's support for multi-tenancy.
+To clarify what this document covers, "multi-tenant" is defined as the ability to host multiple CHT instances on a single server, physical or virtualized.  There is no graphical interface to deploy these instances, so technical staff will be needed to configure each instance per the linked documentation below.
 
+It is the responsibility of each multi-tenant host to ensure all [requirements](/hosting/cht/requirements/) and [considerations](hosting/cht/considerations/) are met.  
 
+## Hosting
 
-We should do a write up:
+Both production [Docker](/hosting/cht/docker/) and production [Kubernetes](/hosting/cht/kubernetes/) CHT deployments can safely be configured to be multi-tenant. Each instance will silo all the it's own data and not allow any unauthenticated access from instances hosted on the same server.  
 
-* what defines what multi-tenant is in this context (ie no GUI, technical prowess required)
-* that points to existing best practices of hosting, monitoring, backup and reporting (eg CHT Sync)
-* is not that technical in terms there's no numbered steps with a literal process to follow
+See the [Kubernetes vs Docker](/hosting/cht/kubernetes-vs-docker/) for more information on which technology to use.
+
+## Backup, Monitoring and Reporting
+
+Each multi-tenant deployment can choose to have unique [backup](/hosting/cht/docker/backups/), [monitoring](/hosting/monitoring/) and [reporting](/hosting/analytics/).  This would allow a 1:1 mapping of an instances data, with no cross pollination of the data.
+
+Alternately, these same systems can be configured to have a single shared solution. These will be allowed to aggregate backup, monitoring and reporting into a single, global solution. This can be done in addition to or in place of a 1:1 setup as mentioned above.
+
+The benefit of a global solution is that statistics from all deployments can be aggregated together for reporting purposes, which could be desired in a national deployment, for example. 
+
--- a/content/en/hosting/cht/multi-tenant.md
+++ b/content/en/hosting/cht/multi-tenant.md
@@ -1,37 +1,70 @@
 ---
-title: Multi-tenant
-linkTitle: Multi-tenant
+title: Multi-tenancy in the CHT
+linkTitle: Multi-tenancy
 weight: 30
 description: >
-  How to run multiple instances of the CHT on one server
+  What "multi-tenancy" can mean for a CHT deployment, what the CHT supports today, and why
 relatedContent: >
+  hosting/cht/requirements
+  hosting/cht/considerations
   hosting/cht/kubernetes
   hosting/cht/docker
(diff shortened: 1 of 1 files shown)


andrablaj (changes requested):
Thanks for turning this around so fast, @mrjones-plip.

I'd suggest a few changes to bring it closer to what we chatted about in the team call. I'll share an updated draft building on your structure + @dianabarsan's [forum reply,](https://forum.communityhealthtoolkit.org/t/does-cht-support-muti-tenant-architecture-and-how-much-load-a-single-tenant-can-take/5652) but here's the reasoning first:

- Cover the different meanings of "tenant" instead of _defining_ one. On @jkuester's point: the most useful thing we can do is break down what people _might_ mean (separate orgs, shared-instance access, white-label provisioning) and give the reality for each. Defining multi-tenant as "multiple instances on one server" answers only one of the three.

- Rework "can safely be configured to be multi-tenant." This is the language slide @Hareet flagged as it reads as production-grade multi-tenant support, and it's what AI/search will index. Suggest framing instance-per-tenant as standard containerization, not a CHT feature.

- Add the "why." @dianabarsan's forum reply already has the one-stack/one-config/one-admin-scope breakdown and we can use that language.

- Include the soft multi-tenancy limits. Multiple implementers, one instance seems to be a common ask. Key facts from @dianabarsan's post: offline users can be partitioned by hierarchy, but online users/API see all data and config is shared. 

- I am unsure about the shared/global backup & monitoring guidance. Per Diana's concern on the call, recommending aggregation architecture creates a support surface we are unsure we want to own. 

Updated draft coming in a follow-up comment; happy to talk through any of it. 

Thanks again for kicking this off 🚀

andrablaj:
---
title: Multi-tenancy in the CHT
linkTitle: Multi-tenancy
weight: 30
description: >
  What "multi-tenancy" can mean for a CHT deployment, what the CHT supports today, and why
relatedContent: >
  hosting/cht/requirements
  hosting/cht/considerations
  hosting/cht/kubernetes
  hosting/cht/docker
  hosting/cht/kubernetes-vs-docker
  hosting/cht/docker/backups
  hosting/monitoring
  hosting/analytics
---

The CHT does not have built-in multi-tenancy features. Because "multi-tenant" is an umbrella term covering several very different setups, this page describes the most common scenarios the community asks about, what is and isn't possible in each, and why.

If your scenario isn't covered here, please describe it on the [CHT Forum](https://forum.communityhealthtoolkit.org/). The more specific you can be about what a "tenant" means in your context, the better the guidance you'll get.

## Why the CHT is not multi-tenant

A CHT instance is designed as **one stack, one configuration, one admin scope**. Each instance is a fixed set of services pointed at a single CouchDB server:

- **api** – REST API and replication endpoint
- **sentinel** – background processor (transitions, scheduled messages, purging)
- **couchdb** – the data store (single-node or a 3+ node cluster)
- **haproxy** – database load balancer
- **nginx** – TLS termination and static assets

All organizational data lives in a handful of databases on that one CouchDB server, and the entire deployment shares a single app configuration (forms, hierarchy, tasks, targets) and a single admin scope. There is no mechanism to divide an instance into independently configured or independently administered partitions.

See related [forum discussion](https://forum.communityhealthtoolkit.org/t/does-cht-support-muti-tenant-architecture-and-how-much-load-a-single-tenant-can-take/5652).

## Scenario 1: Separate organizations, each with their own configuration and data

*Example: several implementing partners each want their own forms, hierarchy, and data, fully isolated from one another.*

This is not possible within a single CHT instance. The supported pattern is **instance-per-tenant**: deploy a separate CHT stack for each organization. Each instance fully silos its own data; there is no shared access between instances.

Running multiple independent CHT instances on shared infrastructure is standard containerization, not a CHT feature. Each instance is deployed and operated exactly as documented for a single instance:

- On Kubernetes, each tenant is its own Helm release with its own project name, namespace, and CouchDB. Multiple releases can run on a shared cluster. See [Kubernetes hosting](/hosting/cht/kubernetes/).
- With Docker, multiple instances can run on one host using separate Compose projects. See [Docker hosting](/hosting/cht/docker/).
- See [Kubernetes vs Docker](/hosting/cht/kubernetes-vs-docker/) to choose between them.

Note that the operator of the shared infrastructure is responsible for ensuring every instance independently meets all [requirements](/hosting/cht/requirements/) and [considerations](/hosting/cht/considerations/), including [backups](/hosting/cht/docker/backups/) and [monitoring](/hosting/monitoring/) for each instance. Sizing, port allocation, TLS, and resource isolation across instances are general infrastructure concerns every CHT deployment should own. 

## Scenario 2: Multiple groups sharing one instance, with separated data access

*Example: several implementers work in different regions of the same district and each should only see their own contacts and reports.*

This is only **partially** possible, and the limits are the following:

- **Offline users** are scoped by the contact hierarchy and replication depth, so they can be partitioned: each offline user only replicates and sees the part of the hierarchy they belong to.
- **Online users and the API have read access to all data in the instance.** There is no mechanism to restrict this.
- **Configuration is shared.** All groups in the instance use the same forms, tasks, targets, and hierarchy definitions. One group cannot have its own variant of the app.

If the groups need isolated online/admin access or their own configuration, this scenario becomes Scenario 1: run separate instances.

## Scenario 3: Self-service or white-label provisioning

*Example: a host organization wants a graphical interface where third parties can spin up their own customized CHT deployment under a shared umbrella.*

There is no graphical interface for deploying or provisioning CHT instances; every instance is deployed by technical staff following the hosting documentation above.

## Scale of a single instance

Multi-tenancy questions are sometimes really scale questions: "do we need to split into tenants to handle our load?" Usually not. Single CHT instances run in production at national and sub-national scale, comfortably serving tens of thousands of users. The real constraint is replication load — how many offline users sync and how many documents each replicates — rather than raw user count. See [scalability considerations](/hosting/cht/considerations/) before assuming you need multiple instances for capacity reasons.

mrjones-plip:
Thanks for the feedback @andrablaj !  

I've committed your changes (0598f88) and then made some edits in a separate commit (7f3f4ed) so you can see what I edited. The main two changes were updating the list of services to use more of the language from our [architecture page](https://docs.communityhealthtoolkit.org/technical-overview/architecture/cht-core/) and them removing all the info on how Kubernetes and Docker are set up.  I don't think the fact that an instance may or may not have a helm chart (none of eCHIS KE have their own charts for example) or that Docker can use compose is relevant to multi-tenant.  These details are downstream in the links provided on the page.

Feel free to push back on these edits!

I think the removal of my [language](https://github.com/medic/cht-docs/pull/2232/changes/e4ff98fa36f8cfc5b11a5cca44f3ff2e386bad05#diff-a609fc99d479eed5f9a7135e7f228a9776a62d19995cb754b5f19593817581c6R34) of shared monitoring, backup and reporting is a loss though.  I think it makes the offering more palatable if it's explicitly stated we support both 1:1 of these services:CHT Instances and 1:many — they'll otherwise be left guessing!  Let's leave this out and ship it as is for now though.

dianabarsan (approved):
This is great! Nice work @andrablaj and @mrjones-plip !

andrablaj (approved):
🚀

PR #2243, opened 2026-08-26
feat(#10855): document outbound headers and proxy

magp18:
# Description

Documents the new outbound `destination.headers` and `destination.proxy` options added in medic/cht-core#10861.

`headers` is an object of extra HTTP headers. Each header uses exactly one of `value` (plain string) or `value_key` (CHT credentials). `auth.type: "header"` still only supports the `Authorization` header; other secrets go in `headers`.

`proxy` is optional and is passed through to undici `ProxyAgent`.

Related: medic/cht-core#10855

# License


jkuester (approved):
Thank you for the docs @magp18 !  🚀  LGTM!

PR #2248, opened 2026-08-28
some slim linux distros don't have bzip2

mrjones-plip:
# Description

Testing our core dev setup I found my [incus](https://linuxcontainers.org/incus/)  Ubuntu VM didn't have `bzip2` installed and I was getting an error when I called `npm run build-dev` per [the docs](https://docs.communityhealthtoolkit.org/community/contributing/code/core/dev-environment/#cht-core-cloning-and-setup):

```
npm error npm error Phantom installation failed Error: Command failed: tar jxf /tmp/phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2                                                                 
npm error npm error tar (child): bzip2: Cannot exec: No such file or directory                   
npm error npm error tar (child): Error is not recoverable: exiting now                           
npm error npm error tar: Child returned status 2                                                                                                                                                   
npm error npm error tar: Error is not recoverable: exiting now                                   
npm error npm error                                                                                                                                                                                
npm error npm error     at genericNodeError (node:internal/errors:983:15)                                                                                                                          
npm error npm error     at wrappedFn (node:internal/errors:537:14)                               
npm error npm error     at ChildProcess.exithandler (node:child_process:417:12)                                                                                                                    
npm error npm error     at ChildProcess.emit (node:events:519:28)                                                                                                                                  
npm error npm error     at maybeClose (node:internal/child_process:1101:16)                      
npm error npm error     at Socket.<anonymous> (node:internal/child_process:456:11)               
npm error npm error     at Socket.emit (node:events:519:28)                                      
npm error npm error     at Pipe.<anonymous> (node:net:347:12) {
npm error npm error   code: 2,                                                                   
npm error npm error   killed: false,                                                             
npm error npm error   signal: null,                                                              
npm error npm error   cmd: 'tar jxf /tmp/phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2'                                                                                                           
npm error npm error } Error: Command failed: tar jxf /tmp/phantomjs/phantomjs-2.1.1-linux-x86_64.tar.bz2 
```

# License

The software is provided under AGPL-3.0. Contributions to this project are accepted under the same license.

PR #2250, opened 2026-08-29
fix(#2249): correct macOS Node, Python and CouchDB steps in dev guide

### LINKED ISSUE #2249: Five defects in the CHT Core dev environment setup guide (macOS)
Following [the CHT Core dev environment guide](https://docs.communityhealthtoolkit.org/community/contributing/code/core/dev-environment/) end to end on macOS 26.3.1 (Apple Silicon), five steps are wrong. Fixes are in #2250, all in `content/en/community/contributing/code/core/dev-environment.md`.

**Defects**

- **The Node install doesn't take effect.** `brew install node@22` is [keg-only](https://docs.brew.sh/FAQ#what-does-keg-only-mean), so `node` stayed on the 26.4.0 I already had. The verify step (`node -v && npm -v`) then prints the wrong version while the prose says it should print 22.x.x — so everything downstream builds against the wrong Node with nothing flagging it.
- **`pyenv global 2.7.18` pins an EOL Python nothing in the repo uses.** The only Python tooling is `haproxy-healthcheck/`, whose own `.tool-versions` says `python 3.10.13`. I have no `python2` installed and the full documented setup completed clean.
- **`[python 2.7]` again in the Prerequisites list** — same fix.
- **`docker stop medic-couchdb` names a container that doesn't exist.** The compose file the guide downloads creates `cht-docker-couchdb-1` and `cht-docker-nouveau-1`. A bare `docker compose stop` doesn't work either — the compose file marks `COUCHDB_PASSWORD` mandatory with `:?`, so it aborts without the env prefix.
- **Broken markdown link:** `* [bzip2])(https://sourceware.org/bzip2/downloads.html)` — `])(` should be `](`.

**Improvements** — proposals, not in the PR, happy to drop any of them:

- Put the CouchDB credentials in `~/cht-docker/.env` so the compose commands work bare (tested; would also simplify the fix above).
- Link to [automated-tests.md](https://docs.communityhealthtoolkit.org/community/contributing/code/core/automated-tests/) at the end — the guide stops once the app is running and never says how to run the suites.
- Say once that the paths assume a `~/cht-core` clone (hardcoded in 8 places). Possibly related to #1116.
- Note that CouchDB data lands in `~/cht-docker/srv`, so it's clear what `docker compose down -v` would destroy.

ken-talltree-io:
# Description

Fixes #2249

Five steps in the CHT Core dev environment guide are wrong on macOS. I hit all five following the guide end to end today on macOS 26.3.1 (Apple Silicon); each fix is one I ran rather than reasoned about.

**1. `node@{{< param nodeVersion >}}` is keg-only, so the install step doesn't take effect.** Homebrew does not symlink keg-only formulae onto the PATH, so after the macOS tab's `brew install`, `node` is still whatever it was before. Adds the `export PATH` line, and gives the `node -v && npm -v` verification an explicit failure branch — as written it prints the wrong version and reports nothing wrong, so the reader goes on to build and run everything against the wrong Node.

**2 & 3. `pyenv global 2.7.18` pins an EOL Python the repo cannot use**, in the macOS tab and again in the prerequisites list. The only Python tooling in `cht-core` is `haproxy-healthcheck/`, whose `Makefile` runs `python3 -m venv .venv` and whose `.tool-versions` declares `python 3.10.13`. The Linux and WSL2 tabs of this same guide install `python3`.

**4. `docker stop medic-couchdb` names a container that is never created.** The `docker-compose.yml` this guide downloads creates `cht-docker-couchdb-1` and `cht-docker-nouveau-1`, so the documented command fails with `No such container`, and correcting only the name would still leave `nouveau` running. Replaces stop and start with the compose equivalents. They need the `COUCHDB_*` prefix because the compose file marks `COUCHDB_PASSWORD` mandatory with `:?` — without it `docker compose` aborts during interpolation.

**5. Malformed markdown link** on the `bzip2` prerequisite — `])(` should be `](`.

### How this was verified

| Claim | How it was checked |
|---|---|
| `node@22` is keg-only | `brew info node@22` → `[keg-only]`, "was not symlinked into /opt/homebrew" |
| The verify step passes on the wrong Node | Ran the macOS tab verbatim; `node -v` reported 26.4.0, not 22.x.x |
| Nothing needs Python 2 | `python2` absent from the machine; `npm ci` (3439 pkgs), `npm run build-dev` and `npm run dev-api` all completed clean |
| `.tool-versions` asks for 3.10.13 | Read `haproxy-healthcheck/.tool-versions` |
| `docker stop medic-couchdb` fails | Ran it: `Error response from daemon: No such container: medic-couchdb` |
| Bare `docker compose stop` fails | Ran it: `required variable COUCHDB_PASSWORD is missing a value` |
| The replacement commands work | Full round trip — `stop` → both containers `Exited (143)` → `start` → both `Up` → CouchDB answering on 5984 |
| `callout type="warning"` renders | Theme is Hextra; `callout type` already used in three other `content/en` pages |

### Not in this PR

#2249 also lists six **improvements** — a `.env` for the CouchDB credentials, a link to the automated-tests page, a note about the hardcoded `~/cht-core` paths, where the CouchDB data lives, a question about `| $SHELL` in the Linux nvm install, and the missing `.nvmrc` in cht-core. I've deliberately kept them out of this PR so the fixes can land on their own. Happy to add any of them here or in a follow-up — your call.

### What I did not verify

Only the macOS path. I have no Linux or WSL2 machine, so I have not confirmed those tabs are unaffected — I believe they are, since neither uses Homebrew or `pyenv`. I also did not run the e2e or integration suites, so the Python 2 claim is scoped to the setup this guide walks through rather than to the whole repo. I have not built the site locally to view the rendered page.

### AI disclosure

Claude Code (Anthropic) was used to run the setup, reproduce each failure, test the replacement commands, and draft this PR. Every command in the table above was executed and its output read. I have reviewed the change and am accountable for it. — Ken Britton

# License

### DIFF
--- a/content/en/community/contributing/code/core/dev-environment.md
+++ b/content/en/community/contributing/code/core/dev-environment.md
@@ -48,9 +48,11 @@ _(Node {{< param nodeVersion >}} is the environment used to run the CHT server i
   # Uses Homebrew: https://brew.sh/
   brew update
   brew install curl jq pyenv git make node@{{< param nodeVersion >}}
-  # Python no longer included by default in macOS >12.3 
-  pyenv install 2.7.18
-  pyenv global 2.7.18
+  # node@{{< param nodeVersion >}} is keg-only, so Homebrew does not symlink it onto your PATH.
+  echo "export PATH=\"\$(brew --prefix node@{{< param nodeVersion >}})/bin:\$PATH\"" >> ~/.$(basename $SHELL)rc
+  # Python is no longer included by default in macOS >12.3
+  pyenv install 3.10.13
+  pyenv global 3.10.13
   echo "eval \"\$(pyenv init --path)\"" >> ~/.$(basename $SHELL)rc
   . ~/.$(basename $SHELL)rc
 ```
@@ -74,6 +76,10 @@ Now let's ensure NodeJS {{< param nodeVersion >}} and npm {{< param npmVersion >
 ```shell
 node -v && npm -v
 ```
+
+{{< callout type="warning" >}}
+On macOS, `node@{{< param nodeVersion >}}` is [keg-only](https://docs.brew.sh/FAQ#what-does-keg-only-mean) — installing it does not put it on your PATH. If the command above reports any version other than {{< param nodeVersion >}}.x.x, the `export PATH` line from the macOS tab has not taken effect; open a new shell, or re-run it, before continuing. Everything below will otherwise silently build and run against the wrong version of Node.
+{{< /callout >}}
 
 Install Docker:
 
@@ -166,7 +172,14 @@ cd ~/cht-core && npm run dev-sentinel
 
 That's it!  Now when you edit code in your IDE, it will automatically reload.  You can see the CHT running locally here: [http://localhost:5988/](http://localhost:5988/)
 
-When you're done with development you can `ctrl + c` in the three terminals and stop the CouchDB container with `docker stop medic-couchdb`.  When you want to resume development later, run `docker start medic-couchdb` and re-run the three terminal commands.
+When you're done with development you can `ctrl + c` in the three terminals and stop the CouchDB containers with:
+
+```shell
+cd ~/cht-docker
+COUCHDB_USER=medic COUCHDB_PASSWORD=password docker compose -f docker-compose.yml -f couchdb-override.yml stop
+```
+
+When you want to resume development later, run the same command with `start` in place of `stop`, then re-run the three terminal commands. The `COUCHDB_*` variables are required because `docker-compose.yml` declares `COUCHDB_PASSWORD` as mandatory; without them `docker compose` aborts before it reaches the container.
 
 ### Adding and accessing data
 
@@ -190,10 +203,10 @@ If you had issues with following the above steps, check out these links for how 
 
 * [Node.js {{< param nodeVersion >}}.x](https://nodejs.org/) & [npm {{< param npmVersion >}}.x.x](https://npmjs.com/) - Both of which we recommend installing [via `nvm`](https://github.com/nvm-sh/nvm#installing-and-updating)
 * [xsltproc](https://github.com/ilyar/xsltproc) 
-* [python 2.7](https://www.python.org/downloads/)
+* [python 3](https://www.python.org/downloads/)
 * [Docker](https://docs.docker.com/engine/install/)
 * [CouchDB](https://docs.couchdb.org/en/stable/install/index.html) - OS package instead of in Docker - you **MUST** use CouchDB 2.x for CHT < 4.4! We still strongly recommend using Docker.
-* [bzip2])(https://sourceware.org/bzip2/downloads.html) - if you're on Ubuntu call: `sudo apt install bzip2`
+* [bzip2](https://sourceware.org/bzip2/downloads.html) - if you're on Ubuntu call: `sudo apt install bzip2`
 
 ### Windows WSL2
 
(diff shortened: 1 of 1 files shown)


content/en/community/contributing/code/core/dev-environment.md
mrjones-plip: Let's get a fix [from another PR](https://github.com/medic/cht-docs/pull/2248) in your PR - thanks!

```suggestion
  sudo apt -y install xsltproc curl uidmap jq python3 git make bzip2
```

mrjones-plip: Let's get a fix [from another PR](https://github.com/medic/cht-docs/pull/2248) in your PR - thanks!

```suggestion
  sudo apt -y install xsltproc curl uidmap jq python3 git make bzip2
```

content/en/community/contributing/code/core/dev-environment.md
```diff
@@ -190,10 +203,10 @@ If you had issues with following the above steps, check out these links for how
 
 * [Node.js {{< param nodeVersion >}}.x](https://nodejs.org/) & [npm {{< param npmVersion >}}.x.x](https://npmjs.com/) - Both of which we recommend installing [via `nvm`](https://github.com/nvm-sh/nvm#installing-and-updating)
 * [xsltproc](https://github.com/ilyar/xsltproc) 
-* [python 2.7](https://www.python.org/downloads/)
+* [python 3](https://www.python.org/downloads/)
 * [Docker](https://docs.docker.com/engine/install/)
 * [CouchDB](https://docs.couchdb.org/en/stable/install/index.html) - OS package instead of in Docker - you **MUST** use CouchDB 2.x for CHT < 4.4! We still strongly recommend using Docker.
-* [bzip2])(https://sourceware.org/bzip2/downloads.html) - if you're on Ubuntu call: `sudo apt install bzip2`
+* [bzip2](https://sourceware.org/bzip2/downloads.html) - if you're on Ubuntu call: `sudo apt install bzip2`
```
mrjones-plip: let's just explicitly add it per my other suggestion. then we can just remove this line and broken link:

```suggestion
```

content/en/community/contributing/code/core/dev-environment.md
```diff
@@ -75,6 +77,10 @@ Now let's ensure NodeJS {{< param nodeVersion >}} and npm {{< param npmVersion >
 node -v && npm -v
 ```
 
+{{< callout type="warning" >}}
+On macOS, `node@{{< param nodeVersion >}}` is [keg-only](https://docs.brew.sh/FAQ#what-does-keg-only-mean) — installing it does not put it on your PATH. If the command above reports any version other than {{< param nodeVersion >}}.x.x, the `export PATH` line from the macOS tab has not taken effect; open a new shell, or re-run it, before continuing. Everything below will otherwise silently build and run against the wrong version of Node.
```
mrjones-plip: Fair point about keeping this in-line as it doesn't silently fail (er, "fail"?  I suspect it works against latest LTS node 🤷 ).  Let's keep it here but snug up the wording:

```suggestion
On macOS, `node@{{< param nodeVersion >}}` is [keg-only](https://docs.brew.sh/FAQ#what-does-keg-only-mean) so it doesn't end up in your `PATH`. If `node -v` shows a different version than {{< param nodeVersion >}}.x.x, the `export PATH` line from the macOS tab has not taken effect. To fix this,  open a new shell, or re-run `export PATH`, before continuing.
```

mrjones-plip (changes requested):
Thanks for the improvements!  As I mentioned, new eyes [make all bugs shallow](https://en.wikipedia.org/wiki/Linus%27s_Law) ;) 

requesting some small tweaks, but overall looks good.

mrjones-plip (approved):
Thanks @ken-talltree-io !  Is a much improved document for macOS users thanks to you .

Looking at it in the fully rendered site, the MacOS warning was a bit too out of place, looking like it was possibly in the Linux or WSL2 instructions, so I moved inside the MacOS tab group and a "in the command below"

PR #2256, opened 2026-09-04
chore(#2255): bump android OS recommended OS

### LINKED ISSUE #2255: Increase recommended android OS in docs
Like we did in https://github.com/medic/cht-docs/issues/2244, Stewardship agreed we should bump the suggested OS from `9` to `16`  but keep the minimum OS `5`.  

As well the wording on why/when the change was made is confusing - some users thought they needed to click through to see specs instead of just referencing the table right there:

> \* Recommended specs increased in CHT [5.3](https://docs.communityhealthtoolkit.org/releases/5_3_0/). This is not due to 5.3 changes, but reflect [suggestions from Google](https://www.androidauthority.com/android-15-minimum-storage-requirement-3543169/) and [the CHT Community](https://forum.communityhealthtoolkit.org/t/android-device-specifications-to-run-cht/5336) at the time of the 5.3 release.

mrjones-plip:
# Description

Bumps Recommended android OS from `9` -> `16` - but leaves minimum OS at `5`. 

closes #2255

# License

### DIFF
--- a/content/en/releases/_index.md
+++ b/content/en/releases/_index.md
@@ -77,12 +77,12 @@ The following is the minimum specification recommendation for smartphones to han
 
 | Specification   | Minimum          | Recommended*     |
 |-----------------|------------------|------------------|
-| Android version | 5.0              | 9.0+             |
+| Android version | 5.0              | 16.0+            |
 | Processor       | 1.0GHz dual-core | 2.0GHz quad-core |
 | RAM             | 1GB              | 4GB              |
 | Storage         | 8GB              | 32GB             |
 
-_\* Recommended specs increased in CHT [5.3](/releases/5_3_0).  This is not due to 5.3 changes, but reflect [suggestions from Google](https://www.androidauthority.com/android-15-minimum-storage-requirement-3543169/) and [the CHT Community](https://forum.communityhealthtoolkit.org/t/android-device-specifications-to-run-cht/5336) at the time of the 5.3 release._
+_\* Recommended specs increased in CHT [5.3](/releases/5_3_0).  This table captures all the recommended specs, but to see background on why these were changed made, see [suggestions from Google](https://www.androidauthority.com/android-15-minimum-storage-requirement-3543169/) and [the CHT Community](https://forum.communityhealthtoolkit.org/t/android-device-specifications-to-run-cht/5336) at the time of the 5.3 release._
 
 If CHWs will be collecting GPS data, autonomous GPS sensors in addition to assisted GPS (A-GPS) for areas with poor GSM network connectivity will enhance the quality of GPS data collected. Autonomous GPS is usually labeled in terms of the supported navigation satellite system:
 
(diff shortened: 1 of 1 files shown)


content/en/releases/_index.md
```diff
@@ -77,12 +77,12 @@ The following is the minimum specification recommendation for smartphones to han
...
 |-----------------|------------------|------------------|
-| Android version | 5.0              | 9.0+             |
+| Android version | 5.0              | 16.0+            |
 | Processor       | 1.0GHz dual-core | 2.0GHz quad-core |
 | RAM             | 1GB              | 4GB              |
 | Storage         | 8GB              | 32GB             |
 
-_\* Recommended specs increased in CHT [5.3](/releases/5_3_0).  This is not due to 5.3 changes, but reflect [suggestions from Google](https://www.androidauthority.com/android-15-minimum-storage-requirement-3543169/) and [the CHT Community](https://forum.communityhealthtoolkit.org/t/android-device-specifications-to-run-cht/5336) at the time of the 5.3 release._
+_\* Recommended specs increased in CHT [5.3](/releases/5_3_0).  This table captures all the recommended specs, but to see background on why these were changed made, see [suggestions from Google](https://www.androidauthority.com/android-15-minimum-storage-requirement-3543169/) and [the CHT Community](https://forum.communityhealthtoolkit.org/t/android-device-specifications-to-run-cht/5336) at the time of the 5.3 release._
```
binokaryg:
```suggestion
_\* Recommended specs increased in CHT [5.3](/releases/5_3_0).  This table captures all the recommended specs, but to see background on why these changed were made, see [suggestions from Google](https://www.androidauthority.com/android-15-minimum-storage-requirement-3543169/) and [the CHT Community](https://forum.communityhealthtoolkit.org/t/android-device-specifications-to-run-cht/5336) at the time of the 5.3 release._
```

binokaryg (changes requested):
Change requested to fix grammar.

Now that we have long-jumped the recommended Android version, the rest of the recommended specs look low. Should we mention that if the device is not a dedicated device for running CHT Android, and has to share resources with other apps, the specs might not be enough?

Not an official one, but one [source](https://www.gsmarena.com/here_are_googles_new_minimum_ram_and_storage_requirements_for_android_phones-news-67387.php) says:

> your experience using a phone with 32GB of storage will be dreadful, Android 15 or not.

> starting on Android 16 they will also have to go the Android Go route, meaning there will in effect be a minimum RAM requirement of 6GB for running full Android.

content/en/releases/_index.md
mrjones-plip: good catch!  tidied both up in a16e98d3

mrjones-plip:
> Now that we have long-jumped the recommended Android version, the rest of the recommended specs look low. 


 Maybe @eljhkrr or @inromualdo have ideas so we can [voltron](https://en.wikipedia.org/wiki/Voltron) up to form a WA, EA & Nepal super robot to figure what recommended specs we want?


Otherwise - I don't feel that strongly about this and defer to your real world experience @binokaryg - go for it with the suggested text and let's get it done!

binokaryg (approved):
Thanks for updating this @mrjones-plip.

We can work on updating the hardware specs further when needed.

PR #2258, opened 2026-09-10
feat(#2257): order docs/forum results first in Kapa search

### LINKED ISSUE #2257: Docs search results shouldn't include results from non-docs sources
Right now if you search the docs for "household geolocation" you get results from CHT Core repo's code.  We shouldn't show this by default, possibly if at all.  It should be very clearly showing results form docs site and maybe let users toggle between other sources if they want - but as search on our docs site, docs results should be primary:

Image

jkuester:
# Description

Closes https://github.com/medic/cht-docs/issues/2257

According to the Kapa [search config docs](https://docs.kapa.ai/integrations/website-widget/configuration/behavior#search-configuration) we can set the `data-search-source-ids-order` property to configure the _order_ we want the search results to show up in.  The id values we provide are our [Kappa AI Sources](https://app.kapa.ai/1cc25d03-dcd0-4c75-9a3e-8fe3443da795/sources) (so docs, forum, cht-core files, cht-core issues, etc.  

In this PR I have just set the ordering so the Docs results show first, followed by the forum results, followed by everything else.  (Happy to adjust this as seems appropriate).

image

## Alternatives

As you can see in the docs, there are several other properties we _could_ set to further tune the experience. The most relevant of which is `data-search-source-ids-include` (or `data-source-group-ids-include`).  With these properties we could filter exactly which sources we show results for on the search page.  (E.g. only show Docs/Forum and never show code.  Or even, show code, but not issues, or vice-versa.)

Because all of this is just an api call to Kappa, we can test different combos locally and see which results we like before pushing to prod...  :+1: 

# License

### DIFF
--- a/layouts/partials/custom/head-end.html
+++ b/layouts/partials/custom/head-end.html
@@ -47,6 +47,7 @@
   data-modal-open-on-command-k="false"
   data-modal-border-radius="6px"
   data-search-mode-enabled="true"
+  data-search-source-ids-order="1af28c7b-d8b4-485c-a8bf-98b4ba246ad9,d7dd5893-1101-432a-b240-259014d66800"
   data-button-hide="true"
 ></script>
 


mrjones-plip (approved):
Nice work on this!  And...sorry - as it looks like I could have figured that out instead of punting it over to you or andra 🤦 

Approving! main goal of "docs first in search" is achieved.

However, when i search for `How to use the new Archive feature coming CHT 5.3` , the title of [a forum post](https://forum.communityhealthtoolkit.org/t/how-to-use-the-new-archive-feature-coming-cht-5-3/5725?u=mrjones) I made >20 days ago, it is not found in the search results.  I think this has to do with search overall functionality though, and not the ordering we're facing here.

jkuester:
I can also confirm that forum threads were not pulling through as I expected in the search results.  Not sure what's up with the indexing... :shrug:   To clarify, I do see forum threads in the search results, but just not all the ones I expect based on my search criteria....

I agree with your  assessment, though, that this is unrelated to our configuration here on the docs site.

PR #2259, opened 2026-09-14
add info about archiving and fragmentation, normalize hextra callout markdown

mrjones-plip:
# Description

Add notice about large amounts of archiving cause high disk use from fragmentation 

# License


dianabarsan (approved):
Cool, thanks for the addition!

